Issue #4, opened 2018-01-05
Improve documentation redaction & highlighting


francisbrito:
For presentational reasons, it would be nice to have code snippets be highlighted in JavaScript language. This is possible by making some simple changes to the markdown text in the `README.md`.

In the same vein, there are some extra spaces around some method calls that can be removed and `@param` strings that could be highlighted.

I'd like to contribute with this changes if you're OK with it.

vrolland:
Feel free to make PR

vrolland:
PR merged. Thank you !

PR #6, opened 2018-01-06
Improve documentation.

francisbrito:
* Add syntax highlighting to code snippets.
* Remove trailing whitespace in some method definitions.
* Fix double quote characters in code snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ Custom configuration
 
 Instantiates a new RequestNetwork instance that provides the public interface to the requestNetwork.js library.
 
-* @param   provider  The Web3.js Provider instance you would like the requestNetwork.js library to use for interacting with the Ethereum network.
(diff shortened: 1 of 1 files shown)


francisbrito:
Here are the changes proposed on #4 :)

Also, since this project uses TypeScript, do you think it would be a good idea to switch into using a dedicated documentation tool such as `typedoc`? It can statically analyze the code an generate developer friendly documentation. I think it would work great for an "API reference" page in the repository's wiki.

Here's an example of how `typedoc`'s output looks like:
screen shot 2018-01-06 at 2 40 23 pm

If you like this proposal, I can add it to this PR :)

vrolland:
We approved your changes, Thank you francisbrito!

typedoc sounds like a good idea feel free to suggest some new commits with it.

PR #7, opened 2018-01-08
Add scripts to generate and server API reference.

francisbrito:
This changes will allow developers to generate API reference documentation in HTML format from the project's source using `typedoc`.

I've opted to write output to a directory called `.tmp/doc` instead of npm's `doc` [convention](https://docs.npmjs.com/files/package.json#directoriesdoc) so that `doc` can be used in the future for other kind of documentation (examples, diagrams, etc). If you prefer following another approach, this PR can be adjusted to so.

In order to view the documentation, one may either open the `.tmp/docs/index.html` HTML file directly in a browser (by double-clicking it or using a command such as `open` or `xdg-open`) or by running `npm run serve-doc`, which will generate documentation and then server it using (minimalist) HTTP server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 # compiled output
+.tmp
 
 # dependencies
 /node_modules
@@ -29,6 +30,7 @@
 npm-debug.log
 testem.log
 /typings
+*.log
 
 # perso
 *.maybe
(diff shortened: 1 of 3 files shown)


francisbrito:
While making this change, I realized that `testrpc` npm script is broken due to a missing dependency (details in #8). I created a separate issue in order not to mix unrelated changes in a single PR, but both could be added here if deemed appropriate.

romaric-juniet:
We opted for a doc generated by compodoc and deployed on aws:
https://github.com/RequestNetwork/requestNetwork/commit/62f4a193bd604725e66ef7ca903b8c1552d680d5

Thanks for the PR!

PR #10, opened 2018-01-11
chore: use tslib helpers

QuentinFchx:
This should reduce the size of the `/dist` folder by importing helpers instead of adding them on the top of the files as seen on the [project page](https://github.com/Microsoft/tslib)

I did not commit the `/dist` folder but can do it. It just feels odd :D. Why is that folder in the repo?


vrolland:
Thank you Quentin, 

Good point, we will delete dist on this repo. (we put it here before because or app was using github to retrieve the code (and not npm)

QuentinFchx:
@etiennetatur I think the issue comes from the fact I added tslib to dev-deps instead of deps 😅.
Thus any library consuming it needed to add `tslib` to its own deps

vrolland:
yep, done ;)

PR #11, opened 2018-01-11
refactor: add types + improve logic in servicesContracts

QuentinFchx:
- Added `InterfaceArtifact` type reference
- Use `.some` instead of `.forEach` since it [returns immediately when truthy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)


vrolland:
Thank you Quentin, PR merged.

PR #23, opened 2018-01-27
Make npm run build work on windows

Hardsix:
Make `npm run build` platform agnostic by using platform independent build and clean scripts

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "ethereum"
   ],
   "scripts": {
-    "clean": "rm -rf ./dist",
-    "build": "npm run clean && tsc && cp -r ./src/artifacts ./dist/src/ && cp ./src/*.json ./dist/src/ && cp -r ./src/lib ./dist/src/",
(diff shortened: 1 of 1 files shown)


vrolland:
Hello Hardix, Thank for the PR.
Why did you add the "mkdir dist" and "mkdir dist/src" while it is already done by the "tsc"? 

Please, correct your PR and I would be more than happy to merge it.

Hardsix:
Thanks for the response @vrolland 
Maybe it's just on my system(Windows 10), but during `npm run build` I'd get an error message saying the `dist` directory doesn't exist, etc.

I can either remove the `mkdir` commands, or replace them with https://www.npmjs.com/package/mkdirp which would create the directories if they don't exist after `tsc`. What do you think?

vrolland:
I would rather avoid to add external library

Hardsix:
@vrolland Ok, I've done as you've requested

vrolland:
Thanks Hardsix. note that this repo will be soon deprecated.

PR #35, opened 2018-03-22
Add burn contract

romaric-juniet:
- Add burner contract
- Add tests for burner contract
- Rename truffle.js into truffle-config.js to avoid conflict on windows
- Add vscode config to run smart contract tests


packages/requestNetworkSmartContracts/contracts/test/vendor/KyberMock.sol
```diff
@@ -0,0 +1,51 @@
+pragma solidity 0.4.18;
+
+
+/// @dev From https://github.com/KyberNetwork/smart-contracts/blob/master/contracts/ERC20Interface.sol
+interface ERC20 {
```
vrolland: Could be imported (used in two contracts: KyberMock & Burner)

romaric-juniet: Discussed offline. The two are slightly different. The ERC20 in Burner has an additional burn() function

PR #37, opened 2018-03-23
modification after verichain audit

vrolland:
- delete unnecessary check
- optimize bytes computation with assembly

### DIFF
--- a/packages/requestNetworkSmartContracts/contracts/core/RequestCore.sol
+++ b/packages/requestNetworkSmartContracts/contracts/core/RequestCore.sol
@@ -355,7 +355,7 @@ contract RequestCore is Administrable {
         constant
         returns(uint8)
     {
-        for (uint8 i = 0; i < 256 && subPayees[_requestId][i].addr != address(0); i = i.add(1)) {
+        for (uint8 i = 0; subPayees[_requestId][i].addr != address(0); i = i.add(1)) {
             // nothing to do
         }
         return i;
@@ -404,7 +404,7 @@ contract RequestCore is Administrable {
     {
         int256 balance = requests[_requestId].payee.balance;
 
-        for (uint8 i = 0; i < 256 && subPayees[_requestId][i].addr != address(0); i = i.add(1))
+        for (uint8 i = 0; subPayees[_requestId][i].addr != address(0); i = i.add(1))
         {
             balance = balance.add(subPayees[_requestId][i].balance);
         }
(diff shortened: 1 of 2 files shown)


packages/requestNetworkSmartContracts/contracts/core/RequestCore.sol
```diff
@@ -355,7 +362,7 @@ contract RequestCore is Administrable {
         constant
         returns(uint8)
     {
-        for (uint8 i = 0; i < 256 && subPayees[_requestId][i].addr != address(0); i = i.add(1)) {
+        for (uint8 i = 0; subPayees[_requestId][i].addr != address(0); i = i.add(1)) {
```
romaric-juniet: Do we have guarantees this won't go into an infinite loop?
We could keep the `i` counter but having it as uint16

packages/requestNetworkSmartContracts/contracts/synchrone/RequestEthereum.sol
```diff
@@ -528,11 +531,13 @@ contract RequestEthereum is RequestEthereumCollect {
 		internal
 	{
 		// Check if the _fromAddress is a payeesId
-		// in16 to allow -1 value
+		// int16 to allow -1 value
 		int16 payeeIndex = requestCore.getPayeeIndex(_requestId, _fromAddress);
 		if(payeeIndex < 0) {
+			uint8 payeesCount = requestCore.getSubPayeesCount(_requestId).add(1);
```
romaric-juniet: indentation issue

packages/requestNetworkSmartContracts/contracts/synchrone/RequestEthereum.sol
```diff
@@ -709,68 +714,64 @@ contract RequestEthereum is RequestEthereumCollect {
...
-        for(uint8 j = 0; j <20; j++) {
-            data[offset+j] = b[j];
-        }
-    	return data;
-    }
+	function updateBytes20inBytes(bytes data, uint offset, bytes20 b)
+		internal
+		pure
+	{
+		require(offset >=0 && offset + 20 <= data.length);
```
romaric-juniet: Change indentation to spaces

romaric-juniet:
It seems the commit "add ERC20 emergency withdraw for requestCore and requestEthereum" doesn't belong in this PR

packages/requestNetworkSmartContracts/contracts/core/RequestCore.sol
vrolland: the safeMath i.add(1) prevent from a infinite loop. the add will assert if i > 255.

Issue #55, opened 2018-07-29
Angular CLI warning when using requestNetwork.js 


giltotherescue:
When I run `ng serve` on a project that includes `requestNetwork.js`, I receive the following warnings:

```
WARNING in ./node_modules/requestnetworkartifacts/README.md
Module parse failed: Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
| # Request_Artifacts
| Artifacts for Request Network Ethereum contracts
| 

WARNING in ./node_modules/requestnetworkartifacts/LICENSE
Module parse failed: Unexpected token (1:4)
You may need an appropriate loader to handle this file type.
| MIT License
| 
| Copyright (c) 2018 
```

It doesn't seem to hurt anything, but is annoying to see in console output every time ng reloads.

Simply deleting these files from node_modules removes the warnings.

romaric-juniet:
Hey Gil, you already reported this issue (#49) and I'm sorry we still haven't gotten around to fixing it. I had a quick look one day, didn't manage to find the solution and now it's somewhere near the bottom of the  todo list, given that it's not that important. If you have an idea of how to fix it, I'll be happy to try it, or merge a PR :)

giltotherescue:
So sorry for posting this twice!! I forgot. I just closed the other issue since this one has more detail.

I am not sure how to fix it, but a quick Google search shows that it is a webpack issue. Seems like there is something that can be done to exclude certain files from webpack. If we do get around to fixing it, I will be sure to send a PR :-)

vicnicius:
I faced the same issue while working with requestNetwork yesterday. I gave it a try and tried to fix it here https://github.com/RequestNetwork/requestNetwork/pull/63.

PR #63, opened 2018-09-25
Fix webpack warning by making require more explicit

vicnicius:
While bundling a project with requestnetwork.js as a dependency with webpack, I saw some warnings of invalid file parsing, like mentioned [here](https://github.com/RequestNetwork/requestNetwork/issues/55).

This PR tries to fix it by being more explicit about what is required. It would be nice to double-check if this would cause any issues on different environments, though I don't think that should be the case.


romaric-juniet:
Hi, thanks for the PR. There's an issue, though, the build is failing: Error: Cannot find module './RequestCore/RequestCore-0.0.5-test.json.json'
https://travis-ci.org/RequestNetwork/requestNetwork/builds/432833743

vicnicius:
Yeah, just saw it. That's because you already have the `.json` extension on the `artifacts.json` file. A solution would be to remove it from there. So you would just have:
```
"private": {                                                                                                                              
  .travis.yml                  |  3         "last-requestcore": "RequestCore/RequestCore-0.0.5-test",
  ...
```
for example. But I don't know if it is used somewhere else. Would you be able to tell? If it is fine I'll update the PR changing also `artifacts.json`.

romaric-juniet:
That can work! But I'd prefer to have `index.js` handling the removal of the potential trailing `.json`. It would to keep the explicit file names in artifact.json, and would keep the webpack specific behavior in the same place

romaric-juniet:
Thanks, looking great now! 
I'm sorry to ask another change, but I noticed you branched off `master`, but we develop from `development`. Would you mind updating the PR?

vicnicius:
Don't worry! I should've checked before! :D Updated it.

edyrkaj:
Hi @romaric-juniet, is this fix inlcuded in release of `0.8.0`?
Thnx

romaric-juniet:
Hi! It's not, 0.8.0 only includes the new ERC20s. We're currently preparing the next very that will include it, it will ship in the coming days

edyrkaj:
thank you @romaric-juniet ;)

PR #75, opened 2019-12-04
docs(encryption): typo in field name

benjlevesque:
unsure about the other `data` fields, should they also be renamed `encryptedData`
Also, should this bump the version?

### DIFF
--- a/packages/transaction-manager/specs/encryption.md
+++ b/packages/transaction-manager/specs/encryption.md
@@ -78,14 +78,14 @@ Example:
 
 | Property | Type   | Description                                                |
 | -------- | ------ | ---------------------------------------------------------- |
-| **data** | String | Encrypted data in base64                                   |
+| **encryptedData** | String | Encrypted data in base64                                   |
 | **hash** | String | Normalized Keccak256 hash of the message before encryption |
 
 Example:
 
 ```JSON
 {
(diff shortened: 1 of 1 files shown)


packages/transaction-manager/specs/encryption.md
```diff
@@ -50,20 +50,20 @@ The encrypted data, the encrypted keys and a hash of the data are pushed on chai
...
-| **data**             | String | First encrypted data of the channel in base64                                                                                    |
+| **encryptedData**             | String | First encrypted data of the channel in base64                                                                                    |
 | **hash**             | String | Normalized Keccak256 hash of the message before encryption                                                                       |
 | **keys**             | Object | AES-256 key encrypted with ECIES from the parties public keys, encoded in base64 and indexed by their multi-formatted identities |
 | **encryptionMethod** | String | Encryption method use for the channel _('ECIES-AES256' here)_                                                                    |
 
-The data are encrypted with the algorithm AES-256 from the channel key generated for this channel.
+The data is encrypted with the algorithm AES-256 from the channel key generated for this channel.
```
ltacker: Both seems to be grammatically correct

rittme: data is theoretically plural... In practice both work though, so I'm ok with leaving it as is or changing it.

PR #76, opened 2019-12-05
build: automate next version package releases

rittme:
## Description of the changes

Automates the release of `next` package versions.
Fix some dependencies to test deploy.


packages/request-client.js/package.json
```diff
@@ -62,6 +59,8 @@
   },
   "devDependencies": {
     "@compodoc/compodoc": "1.1.10",
+    "@requestnetwork/data-access": "0.5.3",
```
romaric-juniet: we shouldn't need this, even for the tests

PR #87, opened 2019-12-12
test: use chai-as-promised instead of try/catch in tests 

vrolland:
## Link to Jira

https://requestnetwork.atlassian.net/browse/PROT-318


packages/integration-test/test/node-client.test.ts
```diff
@@ -350,7 +354,7 @@ it('create an encrypted and unencrypted request with the same content', async ()
   assert.notExists(plainRequestData.meta!.transactionManagerMeta.encryptionMethod);
 });
 
-it('cannot decrypt a request with the wrong decryption provider', async () => {
+it.only('cannot decrypt a request with the wrong decryption provider', async () => {
```
romaric-juniet: nope nope

rittme: ok

PR #99, opened 2019-12-17
fix: update yarn lock

rittme:


benjlevesque:
this might mean we should have a check in CI to verify the yarn.lock isn't updated after `yarn` runs ? 

https://yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile

PR #106, opened 2020-01-15
feat: persist transaction with custom ethereum provider

vrolland:
## Description of the changes

Add the possibility to persist a transaction without using the request node wallet (for example using Metamask)


packages/integration-test/test/layers.test.ts
```diff
@@ -33,7 +33,7 @@ let encryptionDataPayer: any;
 
 let dataAccess: DataAccessTypes.IDataAccess;
 
-describe('Request system', () => {
+describe.skip('Request system', () => {
```
benjlevesque: forgot to remove a `skip`? if not, should comment why

packages/request-client.js/package.json
```diff
@@ -49,6 +49,7 @@
     "@requestnetwork/epk-signature": "0.5.6",
     "@requestnetwork/multi-format": "0.2.3",
     "@requestnetwork/request-logic": "0.8.2",
+    "@requestnetwork/smart-contracts": "0.2.0",
```
benjlevesque: have you checked the size of the final bundle? Just curious if it increases, since we add new dependencies...

packages/request-node/test/ipfsAdd.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// tslint:disable:no-magic-numbers
+// tslint:disable:no-unused-expression
+describe('ipfsAdd', () => {
+  before(async () => {
+    requestNodeInstance = new requestNode();
+    await requestNodeInstance.initialize();
+
+    // Any port number can be used since we use supertest
+    server = requestNodeInstance.listen(3000, () => 0);
```
benjlevesque: this is not necessary with `supertest`

packages/request-node/test/ipfsAdd.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  it('responds with status 500 to requests with badly formatted value', async () => {
+    await request(server)
+      .post('/ipfsAdd')
+      .send({
+        data: 'not parsable',
+      })
+      .set('Accept', 'application/json')
+      .expect(httpStatus.INTERNAL_SERVER_ERROR);
```
benjlevesque: imo, a server should never send a 500 for "expected" errors. This case is a `400 BAD_REQUEST`: the given data is not properly formatted, and the user should not retry this request.  
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

packages/request-node/test/ipfsAdd.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      'ipfsSize request result should always be empty',
+    ).to.be.equal('1026');
+  });
+
+  it('responds with status 422 to requests with no value', async () => {
+    await request(server)
+      .post('/ipfsAdd')
+      .send({})
+      .set('Accept', 'application/json')
+      .expect(httpStatus.UNPROCESSABLE_ENTITY);
```
benjlevesque: `422 UNPROCESSABLE_ENTITY` means this request is valid, but the state doesn't allow it. Imo, this should be a `400 BAD REQUEST` 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

packages/request-node/test/ipfsAdd.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    expect(
+      serverResponse.body.ipfsHash,
+      'ipfsHash request result should always be empty',
+    ).to.be.equal('QmaViWwahWwCU7DgYBLYwfvBuEU9bj3F3rmLDoAS5ujqXX');
+    expect(
+      serverResponse.body.ipfsSize,
+      'ipfsSize request result should always be empty',
+    ).to.be.equal('1026');
+  });
```
benjlevesque: Just fyi, this can be written like this:
```js 
const serverResponse = await request(server)
      .post('/ipfsAdd')
      .send({ data: blockString })
      .set('Accept', 'application/json')
      .expect(httpStatus.OK)
      .expect({
         ipfsHash: "QmaViWwahWwCU7DgYBLYwfvBuEU9bj3F3rmLDoAS5ujqXX",
         ipfsSize: '1026'
     });
```

benjlevesque: why is the size a string, not a number?

packages/request-node/src/requestNode.ts
```diff
@@ -154,6 +158,15 @@ class RequestNode {
       }
     });
 
+    // Route for ipfs-add request
+    router.post('/ipfsAdd', (clientRequest: any, serverResponse: any) => {
+      if (this.initialized) {
+        return ipfsAdd(clientRequest, serverResponse, this.ethereumStorage, this.logger);
+      } else {
+        return serverResponse.status(httpStatus.SERVICE_UNAVAILABLE).send(NOT_INITIALIZED_MESSAGE);
```
benjlevesque: for later refactoring: this could be moved to a middleware

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -225,6 +225,49 @@ export default class EthereumStorage implements StorageTypes.IStorage {
     };
   }
 
+  /**
+   * Append content into the storage: add the content to ipfs and the hash on Ethereum
```
benjlevesque: since it's prefixed with `_`, I guess it's "unstable"? should maybe be specified in the comments...

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -225,6 +225,49 @@ export default class EthereumStorage implements StorageTypes.IStorage {
...
+    } catch (error) {
+      throw Error(`Ipfs add request error: ${error}`);
+    }
+
+    // Get content length from ipfs
+    let ipfsSize;
+    try {
+      ipfsSize = (await this.ipfsManager.getContentLength(ipfsHash)).toString();
+    } catch (error) {
+      throw Error(`Ipfs get length request error: ${error}`);
```
benjlevesque: consistency 
```suggestion
      throw new Error(`Ipfs get length request error: ${error}`);
```

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -225,6 +225,49 @@ export default class EthereumStorage implements StorageTypes.IStorage {
...
+    try {
+      ipfsHash = await this.ipfsManager.add(data);
+    } catch (error) {
+      throw Error(`Ipfs add request error: ${error}`);
+    }
+
+    // Get content length from ipfs
+    let ipfsSize;
+    try {
+      ipfsSize = (await this.ipfsManager.getContentLength(ipfsHash)).toString();
```
benjlevesque: why `toString`?

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -225,6 +225,49 @@ export default class EthereumStorage implements StorageTypes.IStorage {
     };
   }
 
+  /**
+   * Append content into the storage: add the content to ipfs and the hash on Ethereum
+   * @param content Content to add into the storage
+   * @returns Promise resolving id used to retrieve the content
+   */
+  public async _ipfsAdd(data: string): Promise<StorageTypes.IIpfsMeta> {
```
benjlevesque: just curious, don't we have this anywhere else? adding stuff to IPFS isn't new...

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Creates an instance of HttpDataAccess.
+   * @param nodeConnectionConfig Configuration options to connect to the node. Follows Axios configuration format.
+   */
+  constructor(
+    {
+      nodeConnectionConfig,
+      web3,
+      urlProvider,
```
benjlevesque: `web3providerUrl` ?

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+   */
+  public async persistTransaction(
+    transactionData: DataAccessTypes.ITransaction,
+    channelId: string,
+    topics?: string[],
+  ): Promise<DataAccessTypes.IReturnPersistTransaction> {
+    if (!this.submitterContract) {
+      const network = await this.provider.getNetwork();
+
+      this.networkName =
```
benjlevesque: not used outside of this function, shouldn't be a class field

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+   */
+  public cache: {
+    [channelId: string]: {
+      [ipfsHash: string]: { block: DataAccessTypes.IBlock; storageMeta: any } | null;
+    };
+  } = {};
+
+  /**
+   * Configuration that will be sent to axios for each request.
+   * We can also create a AxiosInstance with axios.create() but it dramatically complicates testing.
```
benjlevesque: just curious, why?

packages/request-client.js/test/index.test.ts
```diff
@@ -173,6 +173,40 @@ describe('index', () => {
     expect(spy).to.have.been.called.once;
   });
 
+  it('uses http://localhost:3000 with persist from local', async () => {
+    const mock = new mockAdapter(axios);
+    const callback = (): any => {
+      return [200, {ipfsSize: 100, ipfsHash: 'QmZLqH4EsjmB79gjvyzXWBcihbNBZkw8YuELco84PxGzQY'}];
+    };
+
+    const spy = chai.spy();
```
benjlevesque: give more explicit name (`spyPersistTransaction`?)

packages/request-client.js/test/http-metamask-data-access.test.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  DataAccessTypes,
+  //   DecryptionProviderTypes,
+  //   EncryptionTypes,
+  //   IdentityTypes,
+  //   RequestLogicTypes,
+  //   SignatureProviderTypes,
+  //   SignatureTypes,
```
benjlevesque: remove?

packages/request-client.js/test/http-metamask-data-access.test.ts
```diff
@@ -0,0 +1,78 @@
...
+} from '@requestnetwork/types';
+
+import { Block } from '@requestnetwork/data-access';
+
+// import Utils from '@requestnetwork/utils';
```
benjlevesque: remove?

packages/request-client.js/test/http-metamask-data-access.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import 'mocha';
+import HttpMetamaskDataAccess from '../src/http-metamask-data-access';
+// import * as Types from '../src/types';
```
benjlevesque: remove ?

packages/request-client.js/test/http-metamask-data-access.test.ts
```diff
@@ -0,0 +1,78 @@
...
+let block2: DataAccessTypes.IBlock = Block.pushTransaction(
+  Block.createEmptyBlock(),
+  { data: 'data11' },
+  'channel1',
+);
+block2 = Block.pushTransaction(block2, { data: 'data22' }, 'channel2');
+
+/* tslint:disable:no-unused-expression */
+describe('HttpMetamaskDataAccess', () => {
+  describe('getChacheAndClean()', () => {
```
benjlevesque:
```suggestion
  describe('getCacheAndClean()', () => {
```

packages/request-client.js/src/http-request-network.ts
```diff
@@ -24,23 +25,34 @@ export default class HttpRequestNetwork extends RequestNetwork {
...
       decryptionProvider?: DecryptionProviderTypes.IDecryptionProvider;
       nodeConnectionConfig?: AxiosRequestConfig;
       signatureProvider?: SignatureProviderTypes.ISignatureProvider;
       useMockStorage?: boolean;
+      persistFromLocal?: boolean;
+      web3?: any;
+      urlProvider?: string;
     } = {
       nodeConnectionConfig: {},
+      persistFromLocal: false,
```
benjlevesque: not conviced about this naming, but can't come up with a better one yet... anyone has a better idea?

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Gets the transactions cached and remove the ones that have been retrieved from the node
+   * (public for easier testing)
+   *
+   * @param channelId The channel id to search for
+   * @param storageLocationFromNode location retrieved from the node
+   * @param timestampBoundaries filter timestamp boundaries
+   */
+  public getTransactionsCachedAndCleanCache(
```
benjlevesque: getCachedTransactionsAndCleanCache

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+// Delay between retry in ms
+const HTTP_REQUEST_RETRY_DELAY = 100;
+
+/**
+ * Exposes a Data-Access module over HTTP
+ */
+export default class HttpMetamaskDataAccess implements DataAccessTypes.IDataAccess {
+  /**
+   * Cache block persisted directly (in case the node did not have the time to retrieve it)
+   * (public for easier testing)
```
benjlevesque: another option could be to pass a cache object to the constructor, so that, in tests, you could pass a mocked cache

packages/request-client.js/package.json
vrolland: No, but we share you concern with @rittme.
I Jira this, so we can handle it later.

packages/request-client.js/src/http-metamask-data-access.ts
vrolland: It is not necessary a web3 provider, What about `ethereumProviderUrl`

packages/request-client.js/src/http-metamask-data-access.ts
vrolland: It avoids to retreive the network information everytime you call the function. but rather do it once for good.

packages/request-client.js/src/http-metamask-data-access.ts
vrolland: I think it is from @romaric-juniet (I copied/past it)

packages/request-client.js/src/http-request-network.ts
vrolland: @rittme ? @ltacker ? Any suggestion ?

packages/ethereum-storage/src/ethereum-storage.ts
vrolland: Normally, ipfs is called internally in `ethereum-storage` the newness is the fact that it is in the interface now.

packages/request-client.js/src/http-metamask-data-access.ts
benjlevesque: hum, it's a very straightforward condition, not really an optimization... having it as a field, we might want to use it in other methods, where it would not be initialized.

packages/request-client.js/src/http-metamask-data-access.ts
benjlevesque: I think this complexifies a lot usage of Axios... it might be related to the mock, not sure. anyway, we can cleanup later...

packages/request-node/test/ipfsAdd.test.ts
vrolland: good point.

packages/ethereum-storage/src/ethereum-storage.ts
vrolland: Good point.

packages/ethereum-storage/src/ethereum-storage.ts
benjlevesque: OK but this function is not called by the existing code, is it?

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+   *
+   * @returns nothing
+   */
+  public async initialize(): Promise<void> {
+    // no-op, nothing to do
+    return;
+  }
+
+  /**
+   * Persists a new transaction on a node through HTTP.
```
ltacker: The Node doesn't fully persist the transaction isn't it ? only IPFS

packages/request-client.js/package.json
```diff
@@ -49,6 +49,7 @@
     "@requestnetwork/epk-signature": "0.5.7",
     "@requestnetwork/multi-format": "0.2.4",
     "@requestnetwork/request-logic": "0.8.3",
+    "@requestnetwork/smart-contracts": "0.3.0",
```
rittme: I did some checks to clear my conscience about the size increase of the library.
from `1460818` to `1475340`. It's a lot less than I expected 👍

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+        maxRetries: HTTP_REQUEST_MAX_RETRY,
+        retryDelay: HTTP_REQUEST_RETRY_DELAY,
+      },
+    )();
+
+    return data;
+  }
+
+  /**
+   * Gets the transactions cached and remove the ones that have been retrieved from the node
```
ltacker:
```suggestion
   * Gets the cached transactions and remove the ones that have been retrieved from the node
```

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+      channelId,
+      topics,
+    );
+
+    // store on ipfs the block and the the ipfs hash and size
+    const {
+      data: { ipfsHash, ipfsSize },
+    } = await axios.post('/ipfsAdd', { data: block }, this.axiosConfig);
+
+    // get the fee require to submit the hash
```
rittme:
```suggestion
    // get the fee required to submit the hash
```

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+
+    // create a block and add the transaction in it
+    const block: DataAccessTypes.IBlock = Block.pushTransaction(
+      Block.createEmptyBlock(),
+      transactionData,
+      channelId,
+      topics,
+    );
+
+    // store on ipfs the block and the the ipfs hash and size
```
rittme:
```suggestion
    // store the block on ipfs and get the the ipfs hash and size
```

packages/request-client.js/src/http-request-network.ts
```diff
@@ -24,23 +25,34 @@ export default class HttpRequestNetwork extends RequestNetwork {
...
   ) {
-    // useMockStorage === true => use mock data-access
-    // useMockStorage === false => use http data-access
     const dataAccess: DataAccessTypes.IDataAccess = useMockStorage
-      ? new MockDataAccess(new MockStorage())
-      : new HttpDataAccess(nodeConnectionConfig);
+      ? // useMockStorage === true => use mock data-access
+        new MockDataAccess(new MockStorage())
+      : // useMockStorage === false
+      persistFromLocal
```
ltacker: We should describe what this option is used for

packages/request-client.js/src/http-request-network.ts
ltacker: `manualEthereumPersist`
`onlyIPFSNode`

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -0,0 +1,306 @@
...
+// Maximum number of retries to attempt when http requests to the Node fail
+const HTTP_REQUEST_MAX_RETRY = 3;
+
+// Delay between retry in ms
+const HTTP_REQUEST_RETRY_DELAY = 100;
+
+/**
+ * Exposes a Data-Access module over HTTP
+ */
+export default class HttpMetaMaskDataAccess implements DataAccessTypes.IDataAccess {
```
rittme: I'm wondering if we can't ommit a good part of this class if we extend HttpDataAccess instead.

packages/request-client.js/src/http-request-network.ts
rittme: `directEthereumBroadcast` `ethereumBroadcast` `localEthereumBroadcast` ?

vrolland: I go for `useLocalEthereumBroadcast`. Feels right for you guys?

ltacker: `useLocalEthereumBroadcast` or just `localEthereumBroadcast` is also fine for me

PR #121, opened 2020-01-31
(OPINIONS) WIP doc: new usage example format

rittme:
The goal of this PR is to discuss the new format for usage example files.
The new format should be:
- Smaller files with more precise examples.
- More files.
- Over-document, more as a beginner guide than as normal code. Imagine the comments could be exported and auto generate a doc page with the example (which is vaguely the long term plan).


packages/usage-examples/src/1-create-request-mock-storage.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * REQUEST INFORMATION
+ *
+ * In the next section of code we declare the request information.
+ */
+
+// The main request info, with the currency, amount (in the smallest denominator), payee identity and payer identity
+const requestInfo: RequestNetwork.Types.IRequestInfo = {
+  currency: 'REQ',
+  expectedAmount: 1e18,
```
benjlevesque: I didn't know you could do that 😮

packages/usage-examples/src/3-create-declarative-request.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * REQUEST INFORMATION
+ *
+ * In the next section of code we declare the request information.
+ */
+
+// The main request info, with the currency, amount (in the smallest denominator), payee identity and payer identity
+const requestInfo: RequestNetwork.Types.IRequestInfo = {
+  currency: 'BTC',
```
benjlevesque: declarative request in BTC? forgot to update maybe?

packages/usage-examples/src/3-create-declarative-request.ts
```diff
@@ -0,0 +1,110 @@
...
+ * It allows us to declare how we will detect and update the payment status for this request.
+ *
+ * For this request, we will use a declarative payment network. It allows the payer to declare it paid the request,
+ * and the payee to confirm the payment was received. It's a simple, manual way of updating the request.
+ *
+ * To create the payment network we declare the network ID (from one of the available here: https://docs.request.network/request-protocol/payment-detection)
+ * and the payment address where we want the payer to send the money to.
+ */
+const paymentNetwork: RequestNetwork.Types.IPaymentNetworkCreateParameters = {
+  id: RequestNetwork.Types.PAYMENT_NETWORK_ID.BITCOIN_ADDRESS_BASED,
```
benjlevesque: that should be declarative, forgot to change?

packages/usage-examples/src/3-create-declarative-request.ts
```diff
@@ -0,0 +1,110 @@
+/**
+ * Creating a declarative Request using a public gateway.
+ * ======================================================
+ *
+ * In this example we will create a declarative request, meaning there is no manual payment detection, the payment information update is manual.
```
benjlevesque:
```suggestion
 * In this example we will create a declarative request, meaning there is no automatic payment detection, the payment information update is manual.
```

packages/usage-examples/src/1-create-request-mock-storage.ts
vrolland: We should have a string here. 
Plus, having a huge number like without using "BigNumber" is a risk to have operation issues.

packages/usage-examples/src/3-create-declarative-request.ts
```diff
@@ -0,0 +1,110 @@
...
+ * The request will be created at a node running locally (https://docs.request.network/request-protocol/getting-started-1/deploy-a-request-node).
+ */
+
+/**
+ * IMPORTS
+ *
+ * First we import the 2 packages we will need to create the request:
+ */
+// The signature provider allow us to sign the request
```
vrolland: Should we add a small sentence to explain that they can use another (even develop) signature provider.

packages/usage-examples/src/3-create-declarative-request.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * SIGNATURE PROVIDER
+ *
+ * The ethereum private key signature provider allow a user to pass in their private ethereum key to sign a request.
+ * The signature is a proof of who created the request and of it's integrity (that no data changed after it was signed).
+ * This process is similar to the signature of an Ethereum transaction.
+ */
+
+// The signature info requires the request creator private key.
+// Please be careful with how you store and handle your private key, since it's a very sensitive piece of data.
```
vrolland: Maybe explain that the `epk-signature-provider` is better to test purpose.

packages/usage-examples/src/1-create-request-mock-storage.ts
benjlevesque: Javascript now has BigInts (notation: `1n`), don't they solve that issue?

rittme: @benjlevesque they do, but also don't. They allow us to use big numbers, but they don't support scientific notation (so basically it's declaring a string to start with).

PR #137, opened 2020-02-12
feat: Payment detection for payment network eth-proxy-contract

vrolland:


vrolland:
pn-eth-input-data and pn-eth-proxy-contract will be merged.

PR #143, opened 2020-02-19
doc: add code2md and usage examples

rittme:
## Description of the changes
Add code2md script for generating code.

Also add the new usage examples that can be used as inspirations for writing some of the articles.

### DIFF
--- a/packages/usage-examples/scripts/code2md/index.js
+++ b/packages/usage-examples/scripts/code2md/index.js
@@ -1,0 +1,53 @@
+const parse = require('./parser');
+const { findTitle, generateHeader } = require('./helpers.js');
+
+/**
+ * Converts the code string to a markdown string
+ * @param {string} content
+ */
+function code2Md(content) {
+  const parsedArray = parse(content);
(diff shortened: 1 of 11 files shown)


packages/usage-examples/scripts/code2md/parser.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
```
benjlevesque: if found online, should add a reference
if not, should publish this!

packages/usage-examples/scripts/code2md/index.js
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * If loaded from the command line.
+ *
+ * Options:
+ * --out: the output path (default to 'docs')
+ *
+ * All arguments will be parsed as files.
+ */
+const fromCLI = !module.parent;
+if (fromCLI) {
```
benjlevesque: CLI usage is a bit counter intuitive, I would have expected to be able to pass a glob (it actually works because ZSH, unsure how it would work without it)

PR #149, opened 2020-02-25
docs: add typedocs generated documentation

rittme:
## Description of the changes
Add typedocs and generate the request-client js API automatically.

Please don't try to read every generated file. Just review the first commit. You can filter it through this menu:
image


benjlevesque:
is it necessary to store the generated files?

yomarion:
Shouldn't we git ignore all the cache markdown files?

packages/docs/docs/client/index.md
```diff
@@ -0,0 +1,41 @@
...
+id: "index"
+title: "@requestnetwork/request-client.js"
+sidebar_label: "README"
+---
+
+[@requestnetwork/request-client.js](index.md) › [Globals](globals.md)
+
+# Website
+
+This website is built using [Docusaurus 2](https://v2.docusaurus.io/), a modern static website generator.
```
yomarion: Shouldn't we include such a mention in the footer instead?

PR #157, opened 2020-03-02
docs: api apps

benjlevesque:


packages/docs/docs/guides/3-API/0-portal-intro.md
```diff
@@ -16,6 +16,23 @@ Our API accepts JSON-encoded request bodies, returns JSON-encoded responses, and
...
+# Authentication
+
+All API endpoints are authenticated.
+Two mechanisms are currently allowed:
+
+- API Key, explained below, to be used for scripting and test purposes
+- OAuth, explain in the [Apps](3-api-apps) section
+
+:::important
+API Keys don't enable the best security and UX for your users, we might drop their support in the future.
```
AdamDowson: I agree we should have a disclaimer here but maybe it should be reworded? I feel it could potentially scare people.

benjlevesque: indeed, any idea?

AdamDowson: Maybe we should say

At Request we are always reviewing our security policy and following the best security practises. Because of this, there is a possibility we may deprecate usage of API keys in the future. 

In general though, I don't think this warning is really necessary.

packages/docs/docs/guides/3-API/3-api-apps.md
```diff
@@ -4,10 +4,37 @@ keywords: [Request, Apps, API]
...
 
 # How it works
 
-TODO: Get in touch with us
+An application built on top of Request API needs to authenticate with the OAuth2 mechanism.
+We recommend using [Auth0 SDKs](https://auth0.com/docs/libraries#sdks) to integrate it, but you are free to choose whatever you want.
+
+# Getting started
+
+TODO: unlock the Apps feature for all users??
```
ltacker: What do you mean "for all users" ?

benjlevesque: currently, the Apps section of the settings is behind a user-based feature flag, handled by auth0. If we add this to the docs, we might want to open the feature to all.

PR #163, opened 2020-03-10
feat: payment detection error does not throw

vrolland:
## Description of the changes

Getting the balance does not throw but feed an attribut named `balance.error`


packages/payment-detection/src/btc/address-based.ts
```diff
@@ -79,42 +80,49 @@ export default class PaymentNetworkBTCAddressBased {
     networkId: number,
   ): Promise<PaymentTypes.BTCBalanceWithEvents> {
     if (!request.extensions[paymentNetworkId]) {
-      throw new Error(`The request do not have the extension : ̀${paymentNetworkId}`);
+      return getBalanceErrorObject(
+        `The request do not have the extension : ${paymentNetworkId}`,
```
benjlevesque:
```suggestion
        `The request does not have the extension: ${paymentNetworkId}`,
```

packages/payment-detection/src/btc/address-based.ts
```diff
@@ -79,42 +80,49 @@ export default class PaymentNetworkBTCAddressBased {
     networkId: number,
   ): Promise<PaymentTypes.BTCBalanceWithEvents> {
     if (!request.extensions[paymentNetworkId]) {
-      throw new Error(`The request do not have the extension : ̀${paymentNetworkId}`);
+      return getBalanceErrorObject(
```
rittme: I'm wondering if errors like this one shouldn't still throw, since it would only happen for an error in the code itself.
What do you think?

packages/payment-detection/src/btc/address-based.ts
```diff
@@ -79,42 +80,49 @@ export default class PaymentNetworkBTCAddressBased {
...
-    return {
-      balance,
-      events,
-    };
+      return {
+        balance,
+        events,
+      };
+    } catch (error) {
+      return getBalanceErrorObject(error.message);
```
rittme: I guess we could have an error code on these. Something like `DETECTION_FAILED` ?

vrolland: The error code will be `UNKNOWN` for now. as it is a code error only for the balance. We will enhance this later.

PR #166, opened 2020-03-10
docs: pay with multisig

benjlevesque:
## Link to Jira


packages/docs/package.json
```diff
@@ -38,8 +38,12 @@
     "@docusaurus/preset-classic": "2.0.0-alpha.43",
     "@docusaurus/theme-live-codeblock": "2.0.0-alpha.43",
     "@docusaurus/utils": "2.0.0-alpha.43",
+    "@requestnetwork/types": "0.12.0",
```
benjlevesque: @rittme should I include the dependencies I need in the doc here, when using aurelius?

rittme: for now we're not running them, so it will work without. But, yes, long term this is the right approach.

packages/docs/docs/guides/4-request-payment/2-multisig.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * # Handle multisig payments
+ *
+ * ## Pay with a multisig
+ * Imports
+ */
+
+import { Contract, ContractTransaction, Signer } from 'ethers';
+
+// TODO export this from @requestnetwork/payment-processor
```
yomarion: Did you forget this TODO?

packages/docs/docs/guides/4-request-payment/2-multisig.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * # Handle multisig payments
+ *
+ * ## Pay with a multisig
```
yomarion:
```suggestion
 * # Handle multisig payments for ERC20
 *
 * ## Pay through a proxy-contract with a multisig
```

packages/docs/docs/guides/4-request-payment/2-multisig.ts
benjlevesque: no, for later :)

PR #172, opened 2020-03-13
docs: Adding the introduction to the Request Protocol

ltacker:
## Description of the changes

Adding the page "Introduction to the Request Protocol" to the documentation


packages/docs/docs/guides/7-protocol/0-intro.md
```diff
@@ -1,10 +1,249 @@
...
+
 # Overview
 
-TODO by Lucas
+The Request Protocol has one basic purpose: **to persist, on a distributed ledger, data representing requests and to be able to retrieve these data in an efficient way**.
+
+To organize these different purposes, the Request Protocol follows the layered architecture pattern. Each layer is responsible for a specific task and a specific level of abstraction. This layered architecture also simplifies the understandability of the code, we believe it's an important matter for an open-source project.
+
+The protocol is composed of four layers:
+- Request logic
```
vrolland: missing  ̀advanced logic`

ltacker: As I already stated on Notion, advanced logic is not a layer in the Layered Architecture Pattern point of viewx

ltacker:
> I agree with @yomarion. the introduction should be very short and the rest must be split.
> 
> I would make the draw more clear. It is difficult to do so through github, I am up for a call about this. @ltacke

We can discuss tomorrow 👍but I think it's already a version pushable, if there are optimization to make we can push them after the doc is released

PR #179, opened 2020-03-25
feat: emits "error" event when the confirmation fails

vrolland:


packages/request-client.js/src/api/request.ts
```diff
@@ -26,6 +26,8 @@ export default class Request {
   private contentDataExtension: ContentDataExtension | null;
   private emitter: EventEmitter;
 
+  private confirmationErrorOccurred: boolean = false;
```
ltacker: Could be explained here why this variable is used

packages/request-client.js/src/api/request.ts
```diff
@@ -133,9 +143,13 @@ export default class Request {
...
 
-    acceptResult.on('confirmed', async () => {
-      requestData.emit('confirmed', await this.refresh());
-    });
+    acceptResult
+      .on('confirmed', async () => {
+        requestData.emit('confirmed', await this.refresh());
+      })
+      .on('error', error => {
+        this.emitter.emit('error', error);
```
ltacker: Mhh why `confirmationErrorOccurred` is not set here ?

packages/request-client.js/src/api/request.ts
```diff
@@ -171,9 +185,13 @@ export default class Request {
...
 
-    cancelResult.on('confirmed', async () => {
-      requestData.emit('confirmed', await this.refresh());
-    });
+    cancelResult
+      .on('confirmed', async () => {
+        requestData.emit('confirmed', await this.refresh());
+      })
+      .on('error', error => {
+        this.emitter.emit('error', error);
```
ltacker: Same as above comment (and same comment for all these kind of line), why not setting `confirmationErrorOccurred` ?

packages/request-client.js/src/api/request.ts
```diff
@@ -574,6 +624,10 @@ export default class Request {
    * @returns The updated request data
    */
   public getData(): Types.IRequestDataWithEvents {
+    if (this.confirmationErrorOccurred) {
+      throw Error('request confirmation failed');
```
ltacker: The message should be more explicit
We could define `confirmationErrorOccurred` as a string and insert the error occurred inside to print it here

packages/request-client.js/src/api/request.ts
```diff
@@ -603,6 +657,9 @@ export default class Request {
   public async refresh(
     requestAndMeta?: RequestLogicTypes.IReturnGetRequestFromId,
   ): Promise<Types.IRequestDataWithEvents> {
+    if (this.confirmationErrorOccurred) {
+      throw Error('request confirmation failed');
```
ltacker: Same as above

packages/request-client.js/src/api/request.ts
vrolland: It is only for the creation, I renamed the variable

packages/request-client.js/src/api/request.ts
vrolland: same same

packages/request-client.js/src/api/request.ts
vrolland: documented and renamed

packages/request-client.js/src/api/request.ts
vrolland: We get the error from an event already. I think it's enough.

PR #192, opened 2020-04-08
feat(request-node): Add Request Node version and Request Client version to requests header

ltacker:
## Description of the changes

Added the Request Node version in the response's header of each request to the Request Node.

## Link to Jira

https://requestnetwork.atlassian.net/browse/PROT-1150


packages/request-node/src/requestNode.ts
```diff
@@ -15,11 +15,15 @@ import ipfsAdd from './request/ipfsAdd';
...
 import { getEthereumStorage } from './storageUtils';
 
+const packageJson = require('../package.json');
+
 const NOT_FOUND_MESSAGE =
   'Not found\nAvailable endpoints:\n/POST persistTransaction\n/GET getTransactionsByChannelId\n/GET getChannelsByTopic\n/POST /ipfsAdd\nGET getConfirmedTransaction';
 
 const NOT_INITIALIZED_MESSAGE = 'The node is not initialized';
 
+const REQUEST_NODE_VERSION_HEADER = 'Request-Node-Version';
```
rittme, line 25: Most non-standard headers start with `X-` for historical reasons. We should probably call it `X-Request-Node-Version`

packages/request-client.js/src/http-data-access.ts
```diff
@@ -4,6 +4,9 @@ import axios, { AxiosRequestConfig } from 'axios';
 
 import { EventEmitter } from 'events';
 
+const packageJson = require('../package.json');
```
benjlevesque, line 7: you could do `import { version } from "../package.json"`

packages/request-node/src/requestNode.ts
```diff
@@ -141,6 +149,12 @@ class RequestNode {
       });
     }
 
+    // Set the Request Node version to the header
+    this.express.use((_: any, res: any, next: any) => {
+      res.header(REQUEST_NODE_VERSION_HEADER, this.requestNodeVersion),
```
benjlevesque, line 154: I'm confused by this last comma 😅

packages/request-node/test/requestNode.test.ts
```diff
@@ -99,6 +102,16 @@ describe('requestNode server', () => {
       .expect('x-custom-test-header', 'test-passed');
   });
 
+  it('the response header contains the Request Node version', async () => {
+    // Import directly requestNode to create a server
+    requestNodeInstance = new requestNode();
+    server = requestNodeInstance.listen(3002, () => 0);
```
benjlevesque, line 108: you don't have to use `listen` in unit tests, aren't we using `supertest`?

packages/request-client.js/src/http-data-access.ts
```diff
@@ -4,6 +4,9 @@ import axios, { AxiosRequestConfig } from 'axios';
...
+const REQUEST_CLIENT_VERSION_HEADER = 'X-Request-Client-Version';
```
vrolland, line 8: `X-Request-Client-Version` is this name not too generic?
maybe `X-Request-Network-Client-Version` to avoid collisions. 
It's a genuine question maybe the name is already specific enough.

yarn.lock
```diff
@@ -4,14 +4,14 @@
 
 "@babel/code-frame@7.8.3", "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
   version "7.8.3"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
+  resolved "http://localhost:4873/@babel%2fcode-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
```
vrolland, line 7: You should do a `yarn` and `git push`, your "yarn.lock" is not right.

packages/request-client.js/src/http-data-access.ts
ltacker: You can't import a json file with `import` unless you declare it
```
declare module "*.json" {
  const value: any;
  export default value;
}
```

I personally prefer using require

packages/request-client.js/src/http-data-access.ts
ltacker: Good remark

packages/request-node/test/requestNode.test.ts
benjlevesque: ping :)

PR #201, opened 2020-04-17
feat: add an option to disable payment detection in the request client

vrolland:


packages/request-client.js/src/api/request.ts
```diff
@@ -674,7 +681,9 @@ export default class Request {
         )}`,
       );
     }
-    if (this.paymentNetwork && requestAndMeta.result.request) {
+    if (this.skipPaymentDetection) {
+      this.balance = null;
+    } else if (this.paymentNetwork && requestAndMeta.result.request) {
```
benjlevesque, line 686: what if `skipPaymentDetection === false` and `requestAndMeta.result.request === null`. Will `balance` be undefined? is this wished?

packages/request-client.js/src/api/request.ts
```diff
@@ -692,4 +701,18 @@ export default class Request {
 
     return this.getData();
   }
+
+  /**
+   * Enables the payment detection
+   */
+  public enablePaymentDetection(): void {
```
benjlevesque, line 708: this mean to get the balance you need to do `request.enablePaymentDetection()`, then `request.refresh()`? 
Would it make sense to add a utility like `loadBalance()`? (thinking out loud, no opinion)

packages/request-client.js/test/index.test.ts
```diff
@@ -1461,6 +1461,83 @@ describe('index', () => {
       );
       sinon.restore();
     });
+
+    it('can skip the get the balance of a request', async function(): Promise<void> {
```
benjlevesque, line 1465: I would add a check that `getBalance` was not called. We do this to improve performance, if the `balance` field is null but `getBalance` was still called, we didn't achieve much!

packages/request-client.js/src/api/request-network.ts
```diff
@@ -138,9 +140,13 @@ export default class RequestNetwork {
    * Create a Request instance from an existing Request's ID
    *
    * @param requestId The ID of the Request
+   * @param disablePaymentDetection if true, skip the payment detection
    * @returns the Request
    */
-  public async fromRequestId(requestId: RequestLogicTypes.RequestId): Promise<Request> {
+  public async fromRequestId(
+    requestId: RequestLogicTypes.RequestId,
+    disablePaymentDetection: boolean = false,
```
benjlevesque, line 148: for this one and all others: would it be more flexible having an `options` object instead of properties? we might have other options in the future...
```
fromRequestId(
    requestId: RequestLogicTypes.RequestId,
    options: {    disablePaymentDetection: boolean },
)
```

packages/request-client.js/src/api/request.ts
vrolland: after rethinking it, we will return the previous balance (to avoid loosing data)

packages/request-client.js/test/index.test.ts
vrolland: after rethinking it, We return the previous balance to avoid loosing information

packages/request-client.js/src/api/request-network.ts
```diff
@@ -163,7 +169,14 @@ export default class RequestNetwork {
     );
 
     // create the request object
-    const request = new Request(this.requestLogic, requestId, paymentNetwork, this.contentData);
+    const request = new Request(
+      this.requestLogic,
+      requestId,
+      paymentNetwork,
+      this.contentData,
+      undefined,
```
rittme, line 177: This undefined here looks weird :/
I'm ok with leaving it for now, but we have to change it in the future (maybe add it to the options?)

packages/request-client.js/test/index.test.ts
rittme: I don't understand. Can't you add a spy to `getBalance` and make sure it's not called?

benjlevesque: I think so too

vrolland: the payment network is nested in the class as private.

packages/request-client.js/src/api/request-network.ts
vrolland: I put it this way because I was thinking that it was a breaking change. but the request constructor is not meaned to be public. so I can arrange this !

packages/request-client.js/test/index.test.ts
rittme: private stuff sucks because of this :/

benjlevesque: Can't you solve this with any ?

PR #205, opened 2020-04-22
fix: enhance node synchronization and storing of ignored data

vrolland:
# Description

- Store the ignored data only at the end of the synchronization
- Store more information on the ignored data to prepare the next step
- Add more tests
- Split a big function in two

# Next step

A mechanism to retrieved previous ignored data through time


packages/types/src/storage-types.ts
```diff
@@ -176,6 +185,14 @@ export interface IPinRequestConfiguration {
   timeout: number;
 }
 
+/** Configuration for the pinRequest method */
+export interface IDataIdIgnored {
```
benjlevesque, line 189: `IIngoredDataId`?

packages/types/src/storage-types.ts
```diff
@@ -176,6 +185,14 @@ export interface IPinRequestConfiguration {
...
+  reason: string;
+  timeoutLastTry: number;
```
benjlevesque, line 191: not clear what this field is

packages/types/src/storage-types.ts
```diff
@@ -73,6 +73,15 @@ export interface IEthereumEntry extends IWithMeta<IEthereumMetadata> {
   hash: string;
   /** parameters used to compute fees */
   feesParameters: IFeesParameters;
+  /** error encounter */
+  error?: { type: ErrorEntries; message: string };
+}
+
+/** Enum of state possible for data */
+export enum ErrorEntries {
+  ipfsConnectionError,
```
benjlevesque, line 82: I think enums are upper case in the rest of the codebase?

packages/types/src/storage-types.ts
benjlevesque: oh, did you mean `lastTryTimestamp`?

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -464,6 +464,128 @@ export default class EthereumStorage implements StorageTypes.IStorage {
     };
   }
 
+  /**
+   * Verify the hashes are present on IPFS for the corresponding ethereum entry
+   * Filtered incorrect hashes
+   * public to facilitate the tests
```
benjlevesque, line 470: If you make a method public to facilitate tests, I think it's not at the right place... maybe should be a helper outside this class, or moved to its own class

packages/ethereum-storage/test/synhronization.test.ts
```diff
@@ -0,0 +1,290 @@
+import 'mocha';
```
benjlevesque, line 1: typo in this file name

benjlevesque: also, not very explicit name

packages/types/src/storage-types.ts
vrolland: yep it's a typo. :+1:

vrolland:
> mind the typo in the commit :)
> 
> could you explain how this will behave? I mean, will it slow down the initialization or retry at a later point in time. Also, does this require clearing existing index files?

1. The behavior remains the same. It just fix the storing of ignored data. (the retry through time will be added in another PR)
2. As the name of the ignored dataId storing has changed, probably not.

packages/ethereum-storage/src/ethereum-storage.ts
vrolland: I moved it in another file :) hope it is better !

packages/ethereum-storage/src/ethereum-entries-to-ipfs-content.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }
+
+  // Save the entries not successfully retrieved after the retries
+  for (const entryRemaining of ethereumEntriesToProcess) {
+    // store the ipfs ignored after the retried
+    await ignoredDataIdsIndex.save(
+      entryRemaining.hash,
+      entryRemaining.error!.message,
+      true,
```
rittme, line 124: I'm not a big fan of booleans as arguments because of this. By reading this file I have no idea what true means. 
I would suggest having a thin function over save, like `saveWithRetry`, that will set this argument to true (and maybe `saveWithoutRetry` for the other cases, but I guess in this context we could just use the default value).

rittme: Since TS and an IDE help a bit with that, I'm not going to say it's necessary to merge the PR. But it's a nice to have.

packages/ethereum-storage/src/ignored-dataIds.ts
```diff
@@ -0,0 +1,152 @@
...
+          lastTryTimestamp: Date.now(),
+          reason,
+          toRetry,
+        });
+      }
+    }
+  }
+
+  /**
+   * Removes in the cache the ignored dataId
```
rittme, line 70:
```suggestion
   * Removes the ignored dataId from the cache
```

packages/ethereum-storage/src/ignored-dataIds.ts
```diff
@@ -0,0 +1,152 @@
...
+   * Constructor
+   * @param store a Keyv store to persist the metadata
+   */
+  public constructor(store?: Keyv.Store<any>) {
+    this.ignoredDataIds = new Keyv<StorageTypes.IIgnoredDataId>({
+      namespace: 'dataIdIgnored',
+      store,
+    });
+
+    this.listIgnoredDataIds = new Keyv<string[]>({
```
rittme, line 26: I don't understand why we need this and what the code related to this do :(

packages/ethereum-storage/src/ethereum-entries-to-ipfs-content.ts
vrolland: Spoiler alert: This will be removed in the next PR.
But I keep this in mind!

packages/ethereum-storage/src/ignored-dataIds.ts
vrolland: I ve added a comment and JIRA it.

PR #206, opened 2020-04-23
feat: add getIgnoredData() to the ethereum storage

vrolland:
## Description of the changes

Add an entry to retry to get previous ignored data

## Next step

Call this entry point from the data-access


packages/ethereum-storage/src/ethereum-entries-to-ipfs-content.ts
```diff
@@ -21,7 +21,7 @@ const SAFE_MAX_HEADER_SIZE: number = 500;
  * @param ethereumEntries Ethereum entries from the smart contract
  * @returns Filtered list of dataId with metadata
  */
-export default async function ethereumEntriesToEntries(
+export default async function œ(
```
benjlevesque, line 24: I don't think this is what you meant 😁

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -391,6 +391,53 @@ export default class EthereumStorage implements StorageTypes.IStorage {
...
+      this.ipfsManager,
+      this.ignoredDataIds,
+      this.logger,
+      this.maxConcurrency,
+    );
+
+    const ids = entries.map(entry => entry.id) || [];
+    // Pin data asynchronously
+    // tslint:disable-next-line:no-floating-promises
+    this.pinDataToIPFS(ids);
```
benjlevesque, line 427: why asynchronously? and what happens if it fails?

packages/ethereum-storage/src/ignored-dataIds.ts
```diff
@@ -121,6 +146,17 @@ export default class IgnoredDataIds {
     return result;
   }
 
+  /**
+   * Check if it is the time to retry the entry
+   * @param entry to check
+   * @returns true if it is time to retry
+   */
+  private shouldBeRetry(
```
benjlevesque, line 154:
```suggestion
  private shouldRetry(
```

packages/ethereum-storage/src/ignored-dataIds.ts
```diff
@@ -121,6 +146,17 @@ export default class IgnoredDataIds {
...
+    entry: StorageTypes.IIgnoredDataId,
+  ): boolean {
+    return entry.toRetry && (entry.lastTryTimestamp as number + Math.floor(Math.exp(entry.iteration)) * INTERVAL_RETRY_MS) <= Date.now();
```
benjlevesque, line 157: you should explain this calculation in plain english

packages/ethereum-storage/test/ethereum-storage.test.ts
```diff
@@ -809,6 +809,68 @@ describe('EthereumStorage', () => {
...
+      ethereumStorage.ipfsManager.read = chai.spy(
+        async (_hash: string): Promise<StorageTypes.IIpfsObject> => ({
+          content: 'ok',
+          ipfsLinks: [],
+          ipfsSize: 2,
+        }),
+      );
+
+      const entries = await ethereumStorage.getIgnoredData();
+      expect(entries.length, 'config wrong').to.be.equal(1);
```
benjlevesque, line 848:
```suggestion
      expect(entries.length, 'config wrong').to.equal(1);
```

packages/ethereum-storage/test/ethereum-storage.test.ts
```diff
@@ -809,6 +809,68 @@ describe('EthereumStorage', () => {
...
+      expect(entries[0], 'config wrong').to.be.deep.equal({
```
benjlevesque, line 849:
```suggestion
      expect(entries[0], 'config wrong').to.deep.equal({
```

benjlevesque:
It's alarming that tests passed with the typo in `packages/ethereum-storage/src/ethereum-entries-to-ipfs-content.ts`

packages/ethereum-storage/src/ethereum-entries-to-ipfs-content.ts
vrolland: interesting typo !

packages/ethereum-storage/src/ethereum-storage.ts
vrolland: Because the pining can be done afterward, If it fails it will just be warning in the logger.

benjlevesque: then we should track the warnings in the logger I guess ;) there are ways to create metrics or alerts for that in GCP logs

vrolland: Yes, it would a very nice to have! I Jira it !

PR #210, opened 2020-04-29
[WIP] feat: encode content data

benjlevesque:
## Link to Jira


benjlevesque:
This PR will probably be abandonned. After discussion with @vrolland we think there is a less obstrusive way to handle this. 

Indeed, at the moment `contentData` is changed from an object to a string, which might have side effects

Options:
 - the best is to handle this at the signature level, where the issue occurs (fix normalize)
 - if not possible, we need to bump spec version so that we can do `Semver.lte(req.version, versionBeforeFix)` and have a proper retro-compatibility

PR #218, opened 2020-05-06
feat: add the identity ethereumSmartContract to the request logic

vrolland:


packages/integration-test/test/layers.test.ts
```diff
@@ -140,7 +140,10 @@ describe('Request system', () => {
     });
 
     const payer = {
-      type: IdentityTypes.TYPE.ETHEREUM_ADDRESS,
+      extra: {
+        network: 'private',
+      },
+      type: IdentityTypes.TYPE.ETHEREUM_SMART_CONTRACT,
```
benjlevesque, line 146: isn't this altering an existing test? did you mean to add a new one?

packages/types/src/identity-types.ts
```diff
@@ -4,9 +4,21 @@ export interface IIdentity {
   type: TYPE;
   // the identification itself
   value: string;
+  // extra information
+  extra?: IExtraInfoIdentity;
```
benjlevesque, line 8: alternative proposition:
- Leave IIdentity untouched
- Create a ISmartContractIdentity:
```typescript
interface ISmartContractIdentity extends IIdentity {
   network:string;
}
```

usage: 
```typescript
if(type === TYPE.ETHEREUM_SMART_CONTRACT){
  const network =(identity as ISmartContractIdentity).network;
}
```

packages/integration-test/test/node-client.test.ts
```diff
@@ -244,6 +244,63 @@ describe('Request client using a request node', () => {
...
+    const requestNetwork = new RequestNetwork({ signatureProvider });
+
+    const payerSmartContract = {
+      type: IdentityTypes.TYPE.ETHEREUM_SMART_CONTRACT,
+      value: '0xf17f52151ebef6c7334fad080c5704d77216b732',
+    };
+
+    const timestampCreation = Utils.getCurrentTimestampInSecond();
+
+    // create request 1
```
benjlevesque, line 257: you should explain why you create 2 request, not trivial :)

packages/request-client.js/src/api/request-network.ts
```diff
@@ -190,7 +190,10 @@ export default class RequestNetwork {
     updatedBetween?: Types.ITimestampBoundaries,
     options?: { disablePaymentDetection: boolean },
   ): Promise<Request[]> {
-    if (identity.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS) {
+    if (
+      identity.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS &&
```
rittme, line 194: this would require less maintenance in the future if we check a list of supported identities with array.includes

packages/request-client.js/src/api/request-network.ts
```diff
@@ -210,7 +213,9 @@ export default class RequestNetwork {
     options?: { disablePaymentDetection: boolean },
   ): Promise<Request[]> {
     const identityNotSupported = identities.find(
-      identity => identity.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS,
+      identity =>
+        identity.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS &&
```
rittme, line 217: and if the previous comment is in a function, it can be reused here already :)

packages/utils/src/identity.ts
```diff
@@ -39,7 +39,10 @@ function normalizeIdentityValue(value: string): string {
  * @returns the error or null if valid
  */
 function hasError(id: IdentityTypes.IIdentity): string | null {
-  if (id.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS) {
+  if (
+    id.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS &&
```
rittme, line 43: Again :)

packages/request-logic/src/config.ts
```diff
@@ -1,6 +1,6 @@
 export default {
   specificationVersion: {
-    current: '2.0.2',
+    current: '2.0.3',
```
rittme, line 3: Why are we bumping the version? Is adding this new identity making things not retrocompatible? 
If so, should we be checking the version when checking for ETHEREUM_SMART_CONTRACT type?

packages/types/src/identity-types.ts
rittme: This looks good to me. We only have to make sure that doing this the network will be correctly added to the transaction data. I guess so, since in runtime we don't know about types anymore, but worth having a look (or even better, writing a test).

packages/request-client.js/src/api/request-network.ts
```diff
@@ -26,6 +26,7 @@ import localUtils from './utils';
  */
 export default class RequestNetwork {
   public bitcoinDetectionProvider?: PaymentTypes.IBitcoinDetectionProvider;
+  public supportedIdentity: IdentityTypes.TYPE[] = Utils.identity.supportedIdentity;
```
benjlevesque, line 29:
```suggestion
  public supportedIdentities: IdentityTypes.TYPE[] = Utils.identity.supportedIdentities;
```

benjlevesque: is this only an alias of `Utils.identity.supportedIdentity`? If so maybe not necessary to have it as a class field

packages/utils/src/identity.ts
```diff
@@ -1,12 +1,18 @@
 import { IdentityTypes } from '@requestnetwork/types';
 
+const supportedIdentity: IdentityTypes.TYPE[] = [
```
benjlevesque, line 3:
```suggestion
const supportedIdentities: IdentityTypes.TYPE[] = [
```

packages/utils/src/identity.ts
```diff
@@ -1,12 +1,18 @@
...
+];
+
 /**
  * Module to manage Request Logic Identity
  */
 export default {
   areEqual,
   hasError,
   normalizeIdentityValue,
+  supportedIdentity,
```
benjlevesque, line 15:
```suggestion
  supportedIdentities,
```

packages/utils/src/identity.ts
```diff
@@ -39,10 +45,7 @@ function normalizeIdentityValue(value: string): string {
  * @returns the error or null if valid
  */
 function hasError(id: IdentityTypes.IIdentity): string | null {
-  if (
-    id.type !== IdentityTypes.TYPE.ETHEREUM_ADDRESS &&
-    id.type !== IdentityTypes.TYPE.ETHEREUM_SMART_CONTRACT
-  ) {
+  if (!supportedIdentity.includes(id.type)) {
```
benjlevesque, line 48:
```suggestion
  if (!supportedIdentities.includes(id.type)) {
```

packages/request-client.js/src/api/request-network.ts
vrolland: I kept it like in case the library and the utils have a gap.

PR #219, opened 2020-05-19
fix: topic can be any not only string

vrolland:


vrolland:
> If topics can by any, the utilities to get requests by topic should be updated too
It is already `any`

PR #223, opened 2020-05-21
feat: amount are only number or string

vrolland:


packages/request-logic/test/unit/actions/create.test.ts
```diff
@@ -571,7 +570,7 @@ describe('CreateAction', () => {
       ).to.throw('Signer must be the payee or the payer');
     });
 
-    it('can create with amount as integer, bigNumber or zero', async () => {
+    it('can create with amount as integer or zero', async () => {
```
rittme, line 573: shouldn't it be a string?

rittme: and, this is internal, right? do we need to change it too?

PR #239, opened 2020-06-24
feat: resubmit stuck transaction with more gas

rittme:
## Description of the changes
Resubmit transactions stuck because of low gas.

PS Upgrading web3.js to 1.2.9 would give us more possibilities, but the change in the way they handle block confirmations breaks many tests in a deep way.

## Link to Jira

https://requestnetwork.atlassian.net/browse/PROT-1210


packages/ethereum-storage/src/smart-contract-manager.ts
```diff
@@ -282,16 +289,63 @@ export default class SmartContractManager {
...
+              const newGasPrice = new bigNumber(
+                await gasPriceDefiner.getGasPrice(StorageTypes.GasPriceType.FAST, this.networkName),
+              );
+
+              // If the new gas price is higher than the previous, resubmit the transaction
+              if (newGasPrice.gt(new bigNumber(gasPriceToUse))) {
+                // Retry transaction with the new gas price and propagate back the result
+                try {
+                  resolve(
+                    await this.addHashAndSizeToEthereum(
```
AdamDowson, line 330: It could be safer to use a retry method in Web3 but this works just as well!

packages/ethereum-storage/src/smart-contract-manager.ts
```diff
@@ -282,16 +289,63 @@ export default class SmartContractManager {
...
+                nonce = tx.nonce;
+              }
+
+              // Get the new gas price for the transaction
+              const newGasPrice = new bigNumber(
+                await gasPriceDefiner.getGasPrice(StorageTypes.GasPriceType.FAST, this.networkName),
+              );
+
+              // If the new gas price is higher than the previous, resubmit the transaction
+              if (newGasPrice.gt(new bigNumber(gasPriceToUse))) {
```
AdamDowson, line 326: At this stage could we have a fallback to add an arbitrary amount to the gasPriceToUse to force it through quicker? e.g. gasPriceToUse + 20 GWEI. Only useful for extreme cases

rittme: Here we will get the new `fast` gas price. For example, if the last gas price was 17, but it got stuck because now it's 50, we will retry with 50.
But I think you have a good point, we can maybe use the new fast gas price + something like 20% to compensate for the case our gas price APIs are lagging behind.

packages/ethereum-storage/src/smart-contract-manager.ts
rittme: I wish I could use it, but web3.js does not implement this endpoint.
It's [specified in geth](https://github.com/ethereum/go-ethereum/blob/9371b2f70c70f887999d10750440fb09fe1e184f/internal/web3ext/web3ext.go#L485), but not web3.js

PR #246, opened 2020-07-15
WIP forex extension

yomarion:
## Description of the changes

Specs: parameters and interpretation
Tests: main things we should test (empty body)

### DIFF
--- a/packages/advanced-logic/specs/payment-context-exchange-rate.md
+++ b/packages/advanced-logic/specs/payment-context-exchange-rate.md
@@ -23,14 +23,14 @@ and to compute a balance that is very close to the one payers and payee expect.
 
 ## Properties
 
-| Property             | Type     | Description                               | Requirement   |
-|----------------------|----------|-------------------------------------------|---------------|
-| **id**               | String   | constant value: "pc-exchange-rate"        | **Mandatory** |
-| **type**             | String   | constant value: "paymentContext"          | **Mandatory** |
-| **version**          | String   | constant value: "0.1.0"                   | **Mandatory** |
-| **events**                       | Array  | List of the actions performed by the extension | **Mandatory** |
-| **values**           | Object   | List of payment context options                                          |               |
-| **values.paymentContextOption**           | Array   | List of payment context options                                          |               |
+| Property                        | Type   | Description                                    | Requirement   |
+|---------------------------------|--------|------------------------------------------------|---------------|
+| **id**                          | String | constant value: "pc-exchange-rate"             | **Mandatory** |
+| **type**                        | String | constant value: "paymentContext"               | **Mandatory** |
+| **version**                     | String | constant value: "0.1.0"                        | **Mandatory** |
+| **events**                      | Array  | List of the actions performed by the extension | **Mandatory** |
+| **values**                      | Object |                                                |               |
+| **values.paymentContextOption** | Array  | List of payment context options                | **Mandatory** |
 
 ### paymentContextOption
 
@@ -51,7 +51,7 @@ extensions currencies.
 The first step for payment initation is to pick one `values.currency` and its related exchange rate extension.
 There should be a payment network for this currency bound to the request, to get required details.
 
-For the selected currency, the payment processor should look at the exchange rate given by `values.oracle`.
+For the selected currency, the payment processor should look at the exchange rate given by the `values.oracle`.
 Once the exchange rate is fetched, the payment processor has a maximum `values.timeframe`to execute the payment, 
 in seconds. The timeframe will be computed based on the payment transaction datetime, so the payment processor 
 should anticipate network delays by adding a time margin.
@@ -124,4 +124,5 @@ A extension state is created with the following properties:
 ### Updates
 
 None.
+
 
--- a/packages/advanced-logic/specs/payment-context-exchange-rate.md
+++ b/packages/advanced-logic/specs/payment-context-exchange-rate.md
@@ -14,12 +14,12 @@ The request denomination currency is used to compute the amount in another curre
 
 The typical scenario is a request denominated in fiat and paid in ERC20.
 
-Many Exchange rates extensions can be associated with one request, together they describe:
+As a payment context, many triplets of oracle, currency and timeframe can be provided in order to describe:
 * Before payment: what are the possible payment currencies and how to compute corresponding amounts
 * After payment: how to compare payments in one currency with a balance in another currency
 
-The version 0.1.0 is not designed to handle properly multiple payments, although it will detect them and
-compute a balance that is very close to the one payers and payee expect.
+The version 0.1.0 is not designed to handle properly multiple payments, although it allows their detection
+and to compute a balance that is very close to the one payers and payee expect.
 
 ## Properties
 
@@ -28,7 +28,13 @@ compute a balance that is very close to the one payers and payee expect.
 | **id**               | String   | constant value: "pc-exchange-rate"        | **Mandatory** |
 | **type**             | String   | constant value: "paymentContext"          | **Mandatory** |
 | **version**          | String   | constant value: "0.1.0"                   | **Mandatory** |
-| **values**           | Object   |                                           |               |
+| **events**                       | Array  | List of the actions performed by the extension | **Mandatory** |
+| **values**           | Object   | List of payment context options                                          |               |
+| **values.paymentContextOption**           | Array   | List of payment context options                                          |               |
+
+### paymentContextOption
+
+| Property             | Type     | Description                               | Requirement   |
 | **values.oracle**    | String   | TODO:how can that be identified properly? | **Mandatory** |
 | **values.timeframe** | Integer  | Exchange rate timespan                    | **Mandatory** |
 | **values.currency**  | Currency | Currency of the expected amount           | **Mandatory** |
--- a/packages/advanced-logic/specs/payment-context-exchange-rate.md
+++ b/packages/advanced-logic/specs/payment-context-exchange-rate.md
@@ -1,0 +1,120 @@
+# Payment Context: Exchange rates
+
+You can be interested in this document if you want to understand or implement how the network deals with payments in a different
+currency than the request currency.
+
+Prerequisite: Having read the advanced logic specification (see [here](./advanced-logic-specs-0.1.0.md))
+
+## Description
+
+This extension describes how a request denominated in one currency should be paid in another currency.
+
+When this extension exists, its currency overrides the request currency from a payment point of view.
+The request denomination currency is used to compute the amount in another currency at the payment.
+
+The typical scenario is a request denominated in fiat and paid in ERC20.
+
+Many Exchange rates extensions can be associated with one request, together they describe:
+* To the payment processor what are the possible payment currencies and how to compute corresponding amounts
+* To the payment detection how to compare payments in one currency with a balance in another currency
+
+The version 0.1.0 is not designed to handle properly multiple payments, although it will detect them and
+compute a balance that is very close to the one payers and payee expect.
+
+## Properties
+
+| Property             | Type     | Description                         | Requirement   |
+|----------------------|----------|-------------------------------------|---------------|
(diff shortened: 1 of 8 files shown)


packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Standard case: one payment
+
+The first step is to detect the payment as described in the payment network extensions.
+
+If one payment is detected, we can compute an observed exchange rate, that we can compare to the oracle rate.
+The observed rate should be between the minimum and maximum oracle exchange rate of the period:
+`[paymentDateTime - values.timeframe, paymentDateTime]`.
+
+#### Edge case: many payments
```
rittme, line 63: I don't see the issue with this one.
I would say that for each payment, we need to get the payment time and apply the rate at that time, then sum everything to get the balance.

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+### To detect payments
+
+#### Standard case: one payment
+
+The first step is to detect the payment as described in the payment network extensions.
+
+If one payment is detected, we can compute an observed exchange rate, that we can compare to the oracle rate.
+The observed rate should be between the minimum and maximum oracle exchange rate of the period:
+`[paymentDateTime - values.timeframe, paymentDateTime]`.
```
rittme, line 61: What happens in this case:

> I make a payment of 4, since that's the current value of the exchange for me. The min on my timeframe was 1, the max was 10. 
What will be the balance value on the request?
How do we make sure that the balance value is not greater or smaller than the request value (making us show the request as underpaid or overpaid)?

packages/advanced-logic/src/extensions/payment-context-exchange-rate.ts
```diff
@@ -0,0 +1,77 @@
...
+  createCreationAction,
+};
+export default exchangeRate;
+
+const CURRENT_VERSION = '0.1.0';
+
+/**
+ * Creates the extensionsData to create the extension pc-exchange-rate
+ *
+ * @param extensions IAdvancedLogicExtensionsCreationParameters extensions parameters to create
```
rittme, line 19: this doesn't match the function signature.

packages/advanced-logic/src/extensions/payment-context-exchange-rate.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @returns IExtensionCreationAction the extensionsData to be store in the request
```
rittme, line 21: The type here should be `IAction`.
ps: yes, it seems it's broken on all other extensions :)

packages/advanced-logic/src/extensions/payment-context-exchange-rate.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Applies the extension action to the request
+ * Is called to interpret the extensions data when applying the transaction
+ *
+ * @param extensionsState IExtensionStates previous state of the extensions
+ * @param extensionAction IAction action to apply
+ * @param requestState IRequest request state read-only
+ *
+ * @returns state of the request updated
+ */
+function applyActionToExtension(
```
rittme, line 46: Can't review this function for now.

packages/advanced-logic/test/extensions/payment-context-exchange-rate.test.ts
```diff
@@ -0,0 +1,37 @@
+//import { ExtensionTypes, RequestLogicTypes } from '@requestnetwork/types';
+//import Utils from '@requestnetwork/utils';
+import 'mocha';
+
+//import erc20ProxyContract from '../../src/extensions/payment-network/erc20/proxy-contract';
```
rittme, line 5: I don't think these commented imports will be useful on the future, you can remove them.

packages/types/src/extension-types.ts
```diff
@@ -51,10 +52,12 @@ export enum ID {
...
   PAYMENT_NETWORK_ETH_INPUT_DATA = 'pn-eth-input-data',
   PAYMENT_NETWORK_ANY_DECLARATIVE = 'pn-any-declarative',
+  PAYMENT_CONTEXT_EXCHANGE_RATE = 'pc-exchange-rate',
 }
 
 /** Type of extensions */
 export enum TYPE {
   CONTENT_DATA = 'content-data',
   PAYMENT_NETWORK = 'payment-network',
+  PAYMENT_CONTEXT = 'payment-context',
```
rittme, line 62: I like the name.

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+---
+
+## Interpretation
+
+When this extension is present at least once, the request payment should be made in one of the exchange rate
+extensions currencies.
+
+### To initiate a payment
+
+The first step for payment initation is to pick one `values.currency` and its related exchange rate extension.
```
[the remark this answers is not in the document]
yomarion: One per pc-exchange-rate, but there can be several pc-exchange-rate.

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+## Description
+
+This extension describes how a request denominated in one currency should be paid in another currency.
+
+When this extension exists, its currency overrides the request currency from a payment point of view.
+The request denomination currency is used to compute the amount in another currency at the payment.
+
+The typical scenario is a request denominated in fiat and paid in ERC20.
+
+Many Exchange rates extensions can be associated with one request, together they describe:
```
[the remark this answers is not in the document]
yomarion: Fixed

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Interpretation
+
+When this extension is present at least once, the request payment should be made in one of the exchange rate
+extensions currencies.
+
+### To initiate a payment
+
+The first step for payment initation is to pick one `values.currency` and its related exchange rate extension.
+There should be a payment network for this currency bound to the request.
```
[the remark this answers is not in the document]
yomarion: Just that you don't know what to pay :) I have precised.

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Properties
+
+| Property             | Type     | Description                         | Requirement   |
+|----------------------|----------|-------------------------------------|---------------|
+| **id**               | String   | constant value: "pc-exchange-rate"  | **Mandatory** |
+| **type**             | String   | constant value: "paymentContext"    | **Mandatory** |
+| **version**          | String   | constant value: "0.1.0"             | **Mandatory** |
+| **values**           | Object   |                                     |               |
+| **values.oracle**    | String   | TODO:should be fixed for the moment | **Mandatory** |
```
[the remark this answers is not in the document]
yomarion: The term `oracle` comes from provider of truth, I think. I does not have to be bound to blockchains.

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+### To initiate a payment
+
+The first step for payment initation is to pick one `values.currency` and its related exchange rate extension.
+There should be a payment network for this currency bound to the request.
+
+For the selected currency, the payment processor should look at the exchange rate given by the `values.oracle`.
+Once the exchange rate is fetched, the payment processor has a maximum `values.timeframe`to execute the payment, 
+in seconds. The timeframe will be computed based on the payment transaction datetime, so the payment processor 
+should anticipate network delays by adding a time margin.
```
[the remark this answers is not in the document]
yomarion: What's wrong with this precision then?
`The timeframe will be computed based on the payment transaction datetime, so the payment processor 
should anticipate network delays by adding a time margin.`

packages/advanced-logic/specs/payment-context-exchange-rate.md
```diff
@@ -0,0 +1,122 @@
...
+
+### To detect payments
+
+#### Standard case: one payment
+
+The first step is to detect the payment as described in the payment network extensions.
+
+If one payment is detected, we can compute an observed exchange rate, that we can compare to the oracle rate.
+The observed rate should be between the minimum and maximum oracle exchange rate of the period:
```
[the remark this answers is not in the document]
yomarion: Cf. https://min-api.cryptocompare.com/data/v2/histominute?fsym=BTC&tsym=GBP&limit=10
You can see highs and lows per time slice

packages/advanced-logic/specs/payment-context-exchange-rate.md
yomarion: The issue is that binding a payment rate to a payment time creates a different between what the payer wants to pay in fiat and what he actually pays. Basically because of the time it takes for his transaction to be mined. 
An ideal solution would be to apply the same mechanism to every sub-payment. Ideal = for later.

packages/advanced-logic/specs/payment-context-exchange-rate.md
yomarion: The balance is 100% because 4 is between 1 and 10. 
4 is the observed rate, it is between min and max [1, 10] of the timeframe.

packages/advanced-logic/src/extensions/payment-context-exchange-rate.ts
yomarion: OK, fixed locally, I'll commit soon.

In the meantime, from `extension-types.ts`:
```
/** Creation action object */
export interface IAction {
  action: string;
  id: ID;
  parameters?: any;
  version?: string;
}
```

`Creation action object` is wrong, isn't it? It should be `Extension action interface`?

packages/advanced-logic/specs/payment-context-exchange-rate.md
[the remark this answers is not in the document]
rittme: Nothing wrong with the precision. But we will have to be creative when implementing this.

packages/advanced-logic/specs/payment-context-exchange-rate.md
rittme: I'm still not completelly convinced. The amount of the request is an important accounting value for both sides, and mainly for the tool using request. I could clearly see this creating mismatches between different values in an app (e.g.: how did I have 10, pay 4, and am left with 3 on my wallet?)
I wonder if having a magin at the balance object would not be interesting for precision: IBalanceWithEvents could show the precise balance of the request (in our example, 4), and the rate extension could add an event (or something else) that shows the margin of values this amount could represent (in our example, 1 to 10).

packages/advanced-logic/specs/payment-context-exchange-rate.md
rittme: I don't see how it differs from the mechanism for one payment. If it's the same mechanism, but done for every payment, that would actually be simpler, since we already iterate over all the payments, and we wouldn't need a special case.

packages/advanced-logic/specs/payment-context-exchange-rate.md
AdamDowson: For me, having the rate timeframe validity in here can make for a breaking UX, if I make a payment and then because of the volatility it shows as underpaid / overpaid this is not an acceptable UX.

Personally, I think we should keep it more simple and just have a timeboxed fixed amount that the user must pay (say 30 minutes) and this value must be regenerated after expiry. (a lot do this with BTC payments).

As this is all handled off-chain we don't have a way of reverting the payment like we could with a smart contract, if we implemented a payment method like this using on-chain oracles with an upper / lower limit it could suffice as we could revert the transaction but potentially having under/over payments for me isn't a good UX.

vrolland (commented):
just to quiet the alerts

PR #248, opened 2020-07-20
docs: create specs for ERC20 fee proxy-contract PN

rittme:
## Description of the changes
Create specs for a Payment Network that allow the payment and detection of fees
on ERC20, using a proxy smart-contract.

## Link to Jira

https://requestnetwork.atlassian.net/browse/PROT-1218

### DIFF
--- a/packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
+++ b/packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
@@ -1,0 +1,328 @@
+# Payment Network - ERC20 with Fee
+
+You may be interested in this document if:
+
+- you want to create your own implementation of the Request protocol
+- you are curious enough to dive and see what is under the hood of the Request protocol
+
+Prerequisite: Having read the advanced logic specification (see [here](./advanced-logic-specs-0.1.0.md)).
+
+## Description
+
+This extension allows the payments and the refunds to be made in ERC20 tokens on the Ethereum blockchain.
+This Payment Network is similar to the [ERC20 Proxy Contract](./payment-network-erc20-proxy-contract-0.1.0.md) extension, with the added feature of allowing a fee to be taken from the payment.
+
+The payment is made through a proxy contract. This proxy contract does the ERC20 token transfer on behalf of the user. The contract ensures a link between an ERC20 transfer and a request through a `paymentReference`.
+
+This `paymentReference` consists of the last 8 bytes of a salted hash of the requestId: `last8Bytes(hash(lowercase(requestId + salt + address)))`:
+
+The contract also ensures that the `feeAmount` amount of the ERC20 transfer will be forwarded to the `feeAddress`.
+
+- `requestId` is the id of the request
+- `salt` is a random number with at least 8 bytes of randomness. It must be unique to each request
+- `address` is the payment address for payments, the refund address for refunds
+- `feeAmount` is the amount of the transfer that should be paid in fees
+- `feeAddress` is the address where the fee will be sent to
+- `lowercase()` transforms all characters to lowercase
+- `hash()` is a keccak256 hash function
+- `last8Bytes()` take the last 8 bytes
+
+As a payment network, this extension allows to deduce a payment `balance` for the request. (see
+[Interpretation](#Interpretation))
+
+## Contract
+
+The contract contains one function called `transferFromWithReferenceAndFee` which takes 6 arguments:
+
+- `tokenAddress` is the address of the ERC20 contract
+- `to` is the destination address for the tokens
+- `amount` is the amount of tokens to transfer to the destination address
+- `paymentReference` is the reference data used to track the transfer (see `paymentReference`)
+- `feeAmount` is the amount of tokens to transfer to the fee destination address
+- `feeAddress` is the destination address for the fee
+
+The `TransferWithReferenceAndFee` event is emitted when the tokens are transfered. This event contains the same 6 arguments as the `transferFromWithReferenceAndFee` function.
+
+TODO: [See smart contract source]()
+
+| Network | Contract Address |
+| ------- | ---------------- |
+| Mainnet | TODO             |
+| Rinkeby | TODO             |
+| Local   | TODO             |
+
+## Properties
+
+| Property                  | Type   | Description                                    | Requirement   |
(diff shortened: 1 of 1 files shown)


rittme:
@benjlevesque, these are good points.
>    Should the Smart Contract call `transferFromWithReference` on the other proxy contract, so that the `TransferWithReference` event is sent? This would simplify the integration of services like ETH watch, but probably increase the cost...

This would simplify things on that end, but it would make the gas price higher for the payer. It would also require allowances for both contracts (if we use DELEGATECALL, but if we don't we lose the sender). I'm open to discussions on this subject, but 
I decided to go this way since both options seem to implicate as much work, but this one is cheaper.
 
>   Shoudn't there be a post-payment verification on the fee address? It feels like the payer can send the fees to themselves... Alternatively, there could be a kind of signature passed to the Smart contract to ensure this address is authorized for that given request; Another possibility would be to have a whitelist of feeAddresses. But then, the feeAmount becomes a problem...

By following the same strategy of the rest of the protocol, we leave the smart-contract as flexible as possible, but impose rules when parsing the Request. In this case, we don't have anything on the contract imposing correctness, but when veryfing the payment we will verify the rules were followed. 
This is similar to ERC20 proxy contract, where the payer can send the payment to himself, but use the correct reference ID. When detecting the payment, we will verify that the payment was sent to the correct address.

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
```diff
@@ -0,0 +1,328 @@
...
+
+This action is valid, if:
+
+- The extension state with the id "pn-erc20-fee-proxy-contract" exists
+- The signer is the `payee`
+- The extension property `paymentAddress` is undefined
+
+##### Warnings
+
+None.
```
yomarion, line 165: Shouldn't there be a warning if the payer tries to add a payment address?

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
```diff
@@ -0,0 +1,328 @@
...
+
+This action is valid if:
+
+- The extension state with the id "pn-erc20-fee-proxy-contract" exists
+- The signer is the `payer`
+- The extension property `refundAddress` is undefined
+
+##### Warnings
+
+None.
```
yomarion, line 205: Shouldn't there be a warning if the payee tries to add a refund address?

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
```diff
@@ -0,0 +1,328 @@
...
+
+This action is valid, if:
+
+- The extension state with the id "pn-erc20-fee-proxy-contract" exists
+- The signer is the `payee`
+- The extension property `feeAddress` is undefined
+
+##### Warnings
+
+None.
```
yomarion, line 245: There should be a warning if the payer tries to perform this action

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
```diff
@@ -0,0 +1,328 @@
...
+
+This action is valid, if:
+
+- The extension state with the id "pn-erc20-fee-proxy-contract" exists
+- The signer is the `payee`
+- The extension property `feeAmount` is undefined
+
+##### Warnings
+
+This action must trigger the warnings:
```
yomarion, line 285: Missing a 2nd warning if the method is called by the payer

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
```diff
@@ -0,0 +1,328 @@
...
+The payment is made through a proxy contract. This proxy contract does the ERC20 token transfer on behalf of the user. The contract ensures a link between an ERC20 transfer and a request through a `paymentReference`.
+
+This `paymentReference` consists of the last 8 bytes of a salted hash of the requestId: `last8Bytes(hash(lowercase(requestId + salt + address)))`:
+
+The contract also ensures that the `feeAmount` amount of the ERC20 transfer will be forwarded to the `feeAddress`.
+
+- `requestId` is the id of the request
+- `salt` is a random number with at least 8 bytes of randomness. It must be unique to each request
+- `address` is the payment address for payments, the refund address for refunds
+- `feeAmount` is the amount of the transfer that should be paid in fees
```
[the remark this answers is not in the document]
rittme: The extension is currency agnostic (like the other ERC20). It will use the request currency (or the payment-context extension currency, I suppose).

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
rittme: Not necessarilly, because this case will never pass the conditions above.

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
rittme: Same as previous, the conditions will make sure to filter this case out.

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
rittme: Same as previous.

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
rittme: Same as previous.

packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md
rittme: To clarify this and the following points about the warnings:
We will show warnings only when a request follow the conditions, but is still something to worry about. 
For example, when the payer creates the request and sets the feeAddress, it can be a valid use-case, but it can be a scam. So we allow that usecase, but we warn the builder that this can be something sketchy.

PR #269, opened 2020-08-19
feat: swap to pay contract

yomarion:
Prepare the swap-to-pay feature from the smart contracts level.
It does not include swapping ETH (but works with WETH) nor adding a fee.


packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
```diff
@@ -0,0 +1,166 @@
...
+    //address _tokenAddress,
+    address _to,
+    uint256 _amount,
+    bytes calldata _paymentReference
+  )
+    external 
+    payable
+  {
+    IERC20 erc20 = IERC20(multiDaiKovan);
+    convertEthToDai(_amount);
```
benjlevesque, line 152: shouldn't this be checked with `require` ?

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
```diff
@@ -0,0 +1,166 @@
...
+    return path;
+  }
+  
+  /**
+  * @notice Performs a DAI token transfer from ETH with a reference
+  * @param _to Transfer recipient
+  * @param _amount Amount to transfer
+  * @param _paymentReference Reference of the payment related
+  */
+  function transferDAIFromETHWithReference(
```
benjlevesque, line 142: can this be made (more) generic, so that we don't have to deploy a contract for new each token pair?

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
```diff
@@ -0,0 +1,166 @@
...
+  // Event to declare a transfer with a reference
+  event TransferWithReference(
+    address tokenAddress,
+    address to,
+    uint256 amount,
+    bytes indexed paymentReference
+  );
+
+
+  address internal constant UNISWAP_ROUTER_ADDRESS = 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D ;
```
benjlevesque, line 110: I think there are parameters that can be given when creating a contract, I guess those hard coded values could go there?

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
```diff
@@ -0,0 +1,166 @@
...
+  )
+    external 
+    payable
+  {
+    IERC20 erc20 = IERC20(multiDaiKovan);
+    convertEthToDai(_amount);
+    require(erc20.transfer(_to, _amount), "transfer() failed");
+    //require(erc20.transferFrom(msg.sender, _to, _amount), "transferFrom() failed");
+    
+    emit TransferWithReference(
```
benjlevesque, line 156: would be interested to know how much more expensive it would be to call our existing contract instead of doing transfer + emit event

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
yomarion: Yes I hope, POC mode :)

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
yomarion: Yes but I need to understand if there can be a good reason to (not) do so. It's just how they did in the example I saw. I would like to investigate a bit further, but it's probably better as you say.

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
yomarion: Yes, that's my next step

packages/smart-contracts/src/pn-ETH-to-DAI-proxy.sol
yomarion: Yes, maybe. What's weird is that the Uniswap contract will reject if you pay... **too much** ETH. I have not tried yet to pay not enough. Many edge cases to consider!

yomarion: Now I understand better how everything works so I should be able to test the necessity of require and be sure.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  
+
+  constructor(address _uniswapRouterAddress, address _paymentProxyAddress) public {
+    uniswapRouter = IUniswapV2Router02(_uniswapRouterAddress);
+    paymentProxy = IFeePaymentNetwork(_paymentProxyAddress);
+    admin = msg.sender;
+  }
+
+  function approvePaymentProxyToSpend(address erc20Address) public {
+    IERC20 erc20 = IERC20(erc20Address);
```
rittme, line 155: With the import from open-zeppelin lib this will have to become:
```suggestion
    IERC20 erc20 = ERC20(erc20Address);
```

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  address public admin;
+  
+
+  constructor(address _uniswapRouterAddress, address _paymentProxyAddress) public {
+    uniswapRouter = IUniswapV2Router02(_uniswapRouterAddress);
+    paymentProxy = IFeePaymentNetwork(_paymentProxyAddress);
+    admin = msg.sender;
+  }
+
+  function approvePaymentProxyToSpend(address erc20Address) public {
```
rittme, line 154: The goal of this function is to approve the payment proxy to spend for the current smart contract, right?

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  }
+  
+  /**
+  * @notice Performs a token swap between a payment currency and a request currency, and then
+  *         calls a payment proxy to pay the request, including fees.
+  * @param _to Transfer recipient = request issuer
+  * @param _amount Amount to transfer in request currency
+  * @param _amountInMax Maximum amount allowed to spend for currency swap, in payment currency.
+            This amount should take into account the fees.
+    @param _path, path of ERC20 tokens to swap from requestedToken to spentToken. The first
```
rittme, line 173: Do we have a plan on how to find the routes?
For example, currently if I want to change USDC to DAI on Uniswap V2 website the cheapest route is:
USDC -> WETH -> MKR -> DAI

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+            address of the path should be the payment currency. The last element should be the
+            request currency.
+  * @param _paymentReference Reference of the payment related
+  * @param _feeAmount Amount of the fee in request currency
+  * @param _feeAddress Where to pay the fee
+  * @param _deadline Deadline for the swap to be valid
+  *
+  * TODO test the behaviour when the deadline is passed
+  */
+  function transferMkrFromDaiWithReference(
```
rittme, line 183: What about this?
```suggestion
  function swapTransferWithReference(
```

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  constructor(address _uniswapRouterAddress, address _paymentProxyAddress) public {
+    uniswapRouter = IUniswapV2Router02(_uniswapRouterAddress);
+    paymentProxy = IFeePaymentNetwork(_paymentProxyAddress);
+    admin = msg.sender;
+  }
+
+  function approvePaymentProxyToSpend(address erc20Address) public {
+    IERC20 erc20 = IERC20(erc20Address);
+    uint256 max = 2**256 - 1;
+    erc20.approve(address(paymentProxy), max);
```
rittme, line 157: we probably want to `require` here

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  }
+  
+  function approveRouterToSpend(address erc20Address) public {
+    IERC20 erc20 = IERC20(erc20Address);
+    uint256 max = 2**256 - 1;
+    erc20.approve(address(uniswapRouter), max);
```
rittme, line 163: we should `require` here too

packages/smart-contracts/src/contracts/SwapToPay.sol
rittme: we should check if approve don't have non-standard messy versions...

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+    uint256 _deadline
+  )
+    external
+  {
+    IERC20 spentToken = IERC20(_path[0]);
+    IERC20 requestedToken = IERC20(_path[_path.length-1]);
+    
+    uint256 requestedTotalAmount = _amount + _feeAmount;
+
+    spentToken.transferFrom(msg.sender, address(this), _amountInMax);
```
rittme, line 200: We probably want to use the SafeTransfer function from the fee proxy smart contract here too, to handle non-standard tokens.

rittme: Also, we need to require.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+
+    // Allow the router to spend all this contract's spentToken
+    if (spentToken.allowance(address(this),address(uniswapRouter)) < _amountInMax) {
+        approveRouterToSpend(address(spentToken));
+    }
+    
+    // TODO? Add require
```
rittme, line 207: We most probably need a check to see if worked. So, I would say yes.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+  {
+    IERC20 spentToken = IERC20(_path[0]);
+    IERC20 requestedToken = IERC20(_path[_path.length-1]);
+    
+    uint256 requestedTotalAmount = _amount + _feeAmount;
+
+    spentToken.transferFrom(msg.sender, address(this), _amountInMax);
+
+    // Allow the router to spend all this contract's spentToken
+    if (spentToken.allowance(address(this),address(uniswapRouter)) < _amountInMax) {
```
rittme, line 203: Do we really need to send the tokens to this contract and from this contract to the swap? Can't we send directly from the `msg.sender`? There is a risk of something failing and the tokens getting stuck in the contract.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+    
+    // Give the change back
+    spentToken.transfer(msg.sender, spentToken.balanceOf(address(this)));
+  }
+  
+
+  /*
+  * Admin functions to edit the admin, router address or proxy address
+  */
+  modifier onlyAdmin() {
```
rittme, line 240: We should probably use the `WhitelistedRole` contract that we use at the RequestHashStorage.
`WhitelistAdminRole` is also an option, but for consistency we could use the same.
Both are from the openzeppelin-solidity package.

rittme (approved):

packages/smart-contracts/src/contracts/SwapToPay.sol
rittme: Ok, reading more I got why we need to do this. But I'll leave the comment because I think the discussion can be interesting.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+
+pragma solidity ^0.6.2;
+
+import "https://raw.githubusercontent.com/Uniswap/uniswap-v2-periphery/master/contracts/interfaces/IUniswapV2Router02.sol";
+
+/**
+ * @dev Interface of the ERC20 standard as defined in the EIP. Does not include
+ * the optional functions; to access them see {ERC20Detailed}.
+ */
+interface IERC20 {
```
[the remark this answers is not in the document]
yomarion: Yes, my mistake, something did not work at first and then I got used to it :D

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
...
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    /**
+     * @dev Emitted when the allowance of a `spender` for an `owner` is set by
+     * a call to {approve}. `value` is the new allowance.
+     */
+    event Approval(address indexed owner, address indexed spender, uint256 value);
+}
+
+interface IFeePaymentNetwork {
```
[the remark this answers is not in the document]
yomarion: Completely agree, I already renamed variables in this direction. 👍

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: Yes 👍

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: What about requiring that the allowance is `max` afterwards, instead of requiring approve? It would make it less dependent on `approve` implementations.

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: Yes and in the meantime I have changed my mind again about returning also the swapped token. In theory only the amount we need is given by the swap contract, but we never know and indeed we don't want any token to be stuck on the contract. I'll add that back, with a condition on balance to minimize fees.

packages/smart-contracts/src/contracts/SwapToPay.sol
```diff
@@ -0,0 +1,259 @@
+/**
+ * v Deployed here: https://rinkeby.etherscan.io/address/0x35177580d2f12c94cb274ec762121f4f7dfd451e
+ * (i) This version was deployed with an additional change refund in requested currency, being always 0 it is removed.
+*/
+
+pragma solidity ^0.6.2;
+
+import "https://raw.githubusercontent.com/Uniswap/uniswap-v2-periphery/master/contracts/interfaces/IUniswapV2Router02.sol";
```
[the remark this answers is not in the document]
yomarion: I now only use the interface, which is enough I think. As the interface is very short, do you think it's worth it moving to a dedicated file?

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: Thanks, done! I have moved the safeTransferFrom to a library for next contracts of this kind. I don't think it's a good idea to update the ProxyFee contract in this direction, but for a later update!

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: I used isOnwable instead, it looks like the new WhitelistedRole

packages/smart-contracts/src/contracts/SwapToPay.sol
yomarion: Since USDT does not comply with the ERC20 require signature, and this function is useless if it fails, I suggest to leave it like this for now. OK for you?

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -0,0 +1,139 @@
...
+        _feeAddress
+    );
+    
+    // Give the change back to the payer, in both currencies (only spent token should remain)
+
+    if (spentToken.balanceOf(address(this)) > 0) {
+      spentToken.transfer(msg.sender, spentToken.balanceOf(address(this)));
+    }    
+    if (requestedToken.balanceOf(address(this)) > 0) {
+      requestedToken.transfer(msg.sender, requestedToken.balanceOf(address(this)));
```
rittme, line 123: This transfer and the previous one seem to be the riskiest part of the contract.
We need to be 100% sure this will only transfer the funds from the current user/payment.

packages/smart-contracts/test/contracts/ERC20SwapToPay.js
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    const finalBuilderBalance = await requestErc20.balanceOf(builder);
+    const finalIssuerBalance = await requestErc20.balanceOf(to);
+    expect(finalBuilderBalance.toNumber()).to.equals(1);
+    expect(finalIssuerBalance.toNumber()).to.equals(10);
+    // Everything should behave exactly the same, except gas usage.
+    // 50k gas units are saved by approving the router and proxy in advance for first users, per token
+    // console.log(`${maxGasUsed - gasUsed} gas units saved by approving in advance`);
+    expect(gasUsed).to.below(maxGasUsed);
```
benjlevesque, line 143: I wonder of the side effects this could have when running tests... it's usually best to keep each test isolated. maybe not a big deal...

benjlevesque (approved):
Great work! 

+ the tests seem very good

packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -34,7 +38,7 @@ module.exports = async function(deployer) {
     console.log('requestSubmitter Whitelisted in requestHashDeclaration');
 
     // Deploy the ERC20 contract
-    const instanceTestERC20 = await deployer.deploy(erc20, 1000); // 1000 initial supply
+    const instanceTestERC20 = await deployer.deploy(erc20, 10000); // 10000 initial supply
```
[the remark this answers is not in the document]
yomarion: It did not break any, I paid attention

packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -94,8 +110,11 @@ module.exports = async function(deployer) {
       ERC20False:               ${ERC20False.address}
       ERC20NoReturn:            ${ERC20NoReturn.address}
       ERC20Revert:              ${ERC20Revert.address}
-    `);
-  } catch (e) {
+      ERC20Alpha:               ${erc20Alpha.address}
```
AdamDowson, line 113: erc20Alpha > ERC20Alpha to keep it uniform

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
yomarion: Yes, I thought about that as well. According to me, the contract is never supposed to hold any token. It manipulates 2 per transaction and at the end gives both back. Thus, the only case when the contract would give back "additional" tokens is if someone previously sent tokens to the contract, with a normal erc20.transfer. I don't see good reasons to do that, so I assume it would only be done by mistake. 

Do you see other possible flaws?

packages/smart-contracts/test/contracts/ERC20SwapToPay.js
yomarion: How can I make it more isolated?

yomarion: Exchange in DM: not a prio, we should compute "maxGasused" in the same test case.

packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -79,7 +83,19 @@ module.exports = async function(deployer) {
 
     await deployer.deploy(ERC20Revert);
     console.log('ERC20Revert Contract deployed: ' + ERC20Revert.address);
-
+    
+    // Swap-to-pay related contracts
+    // Payment erc20: ALPHA
+    const erc20AlphaInstance = await deployer.deploy(erc20Alpha, 100000); // 100000 initial supply
```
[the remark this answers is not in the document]
yomarion: My bad, I was sure it did not work but it does. Thanks!

packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -5,12 +5,16 @@ const EthereumProxy = artifacts.require('./EthereumProxy.sol');
...
 const erc20 = artifacts.require('./TestERC20.sol');
+const erc20Alpha = artifacts.require('./ERC20Alpha.sol');
 const BadERC20 = artifacts.require('./BadERC20.sol');
 const ERC20True = artifacts.require('ERC20True');
 const ERC20False = artifacts.require('ERC20False');
 const ERC20NoReturn = artifacts.require('ERC20NoReturn');
 const ERC20Revert = artifacts.require('ERC20Revert');
 
+const FakeSwapRouter = artifacts.require('FakeSwapRouter');
+const ERC20SwapToPay = artifacts.require('ERC20SwapToPay');
```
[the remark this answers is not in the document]
yomarion: True, I did not pay attention! NB: we can actually remove all the imports and deployments from this migration file, as they are only used during unit testing and we redeploy them before each unit test. I'm only 99.99% sure of the collateral damages so I'll handle that later.

packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -5,12 +5,16 @@ const EthereumProxy = artifacts.require('./EthereumProxy.sol');
 const ERC20FeeProxy = artifacts.require('./ERC20FeeProxy.sol');
 
 const erc20 = artifacts.require('./TestERC20.sol');
+const erc20Alpha = artifacts.require('./ERC20Alpha.sol');
```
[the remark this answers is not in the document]
yomarion: This is one of the only exceptions actually, we need to deploy a second ERC20 for integration tests from payment-processor for instance. But not from a separate contract, I fixed that already, thanks for your feedback.

NB: since we now have 2 ERC20 tokens, it would probably help debugging to name them differently.

packages/smart-contracts/src/contracts/SwapToPay.sol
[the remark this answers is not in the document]
rittme: it's fine like that.

packages/smart-contracts/src/contracts/SwapToPay.sol
rittme: OK

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
rittme: There is nothing obvious, at least.

PR #271, opened 2020-08-22
feat: add etherscan gas oracle provider

rittme:
## Description of the changes
Add etherscan as a gas oracle provider

### DIFF
--- a/packages/ethereum-storage/src/gas-price-definer.ts
+++ b/packages/ethereum-storage/src/gas-price-definer.ts
@@ -1,6 +1,7 @@
 import * as config from './config';
 import EthereumUtils from './ethereum-utils';
 import EtherchainProvider from './gas-price-providers/etherchain-provider';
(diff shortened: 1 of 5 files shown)


benjlevesque (approved):

yomarion (commented):
If the 3 oracles give a price that is higher than the cap, I think isGasPriceSafe will return false every time, and we will ignore the 3 of them. I think we should first compute an average and then cap it to the configured maximum.

rittme:
> If the 3 oracles give a price that is higher than the cap, I think isGasPriceSafe will return false every time, and we will ignore the 3 of them. I think we should first compute an average and then cap it to the configured maximum.

Yes, that will be the case.
We're not using an average anymore, since it was creating problems in the past (probably with a 0 gas price return).
We need a better logic overall, maybe simply the median, or some logic to ignore extremes.

PR #277, opened 2020-08-31
fix: revert failed token transfer

rittme:
## Description of the changes

When an ERC20 transfer fails, revert the transaction.


packages/smart-contracts/src/contracts/ERC20FeeProxy.sol
```diff
@@ -86,6 +86,10 @@ contract ERC20FeeProxy {
             revert(0, 0)
         }
     }
+
+    require(success, "transferFrom() failed");
```
vrolland, line 90: as message I would put something more explicit: "transferFrom() has been reverted"

vrolland (commented):
It seems to be working well on remix. But we need to add more tests with edge cases like: 

`
contract ERC20True {
    function transferFrom(address _from, address _to, uint _value) public returns (bool) {
        return true;
    }
}

contract ERC20False {
    function transferFrom(address _from, address _to, uint _value) public returns (bool) {
        return false;
    }
}

contract ERC20NoReturn {
    function transferFrom(address _from, address _to, uint _value) public {
    }
}

contract ERC20Revert {
    function transferFrom(address _from, address _to, uint _value) public {
        require(false);
    }
}
`

benjlevesque (approved):

PR #284, opened 2020-09-09
docs: how to test payment-processor

yomarion:


benjlevesque (approved):

packages/payment-processor/README.md
```diff
@@ -6,6 +6,20 @@ It contains client-side payment methods for:
 - ERC20 proxy contract
 - ETH input data
 
+### Test
+
```
rittme, line 10: Maybe a quick line to explain why we do it could be nice. Like:
```suggestion
To run the payment-processor tests we need a local running ganache with all our smart contracts deployed. You can open two terminals and do:
```

rittme (approved):

PR #323, opened 2020-09-29
feat: payment processor support for swap-to-pay

yomarion:


packages/payment-processor/src/payment/index.ts
```diff
@@ -55,58 +68,102 @@ export async function payRequest(
 }
 
 /**
+ * TODO: refactor to re-use the payment currency?
```
yomarion, line 71: TODO: refactor this function

packages/payment-processor/src/payment/index.ts
```diff
@@ -55,58 +68,102 @@ export async function payRequest(
...
  * Verifies the address has enough funds to pay the request. For ERC20
  * Supported networks: ERC20_PROXY_CONTRACT, ETH_INPUT_DATA
  *
  * @throws UnsupportedNetworkError if network isn't supported
  * @param request the request to verify.
  * @param address the address holding the funds
+ * @param paymentCurrency if different from the requested currency
  * @param provider the Web3 provider. Defaults to Etherscan.
  */
 export async function hasSufficientFunds(
```
yomarion, line 82: TODO: test with swap settings

packages/payment-processor/src/payment/index.ts
```diff
@@ -33,20 +41,25 @@ export class UnsupportedNetworkError extends Error {
...
   request: ClientTypes.IRequestData,
   signerOrProvider: Web3Provider | Signer = getProvider(),
   amount?: BigNumberish,
+  swapSettings?: ISwapSettings,
   overrides?: ITransactionOverrides,
 ): Promise<ContractTransaction> {
   const signer = getSigner(signerOrProvider);
   const paymentNetwork = getPaymentNetwork(request);
+  if (swapSettings && paymentNetwork !== ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT) {
+    throw new Error(`Payment network: ${paymentNetwork} is not supported by swapToPay contract`);
```
yomarion, line 57: TODO throw UnsupportedPN

packages/payment-processor/src/payment/index.ts
```diff
@@ -55,12 +68,15 @@ export async function payRequest(
 }
 
 /**
+ * TODO: refactor to re-use the payment currency?
+ * 
  * Verifies the address has enough funds to pay the request. For ERC20
```
yomarion, line 73: Document better to be specific to request payment only

packages/payment-processor/src/payment/index.ts
yomarion: OR canSwapToPay

yomarion:
I realize that @types did not depend on ethers until now and that it's the first usage of BigNumber in this lib... Is it a mistake to use this type?

yomarion:
I cannot configure lint to accept `gte`... Any hint?

packages/payment-processor/src/payment/index.ts
```diff
@@ -25,28 +32,47 @@ export class UnsupportedNetworkError extends Error {
...
  * @param signerOrProvider the Web3 provider, or signer. Defaults to window.ethereum.
  * @param amount optionally, the amount to pay. Defaults to remaining amount of the request.
+ * @param swapSettings optionally, the information of how to swap from another payment token.
  * @param overrides optionally, override default transaction values, like gas.
  */
 export async function payRequest(
   request: ClientTypes.IRequestData,
   signerOrProvider: Web3Provider | Signer = getProvider(),
   amount?: BigNumberish,
+  swapSettings?: ISwapSettings,
```
benjlevesque, line 64: changing the order of the parameters makes this PR a breaking change.

benjlevesque: It also makes it pretty dependant to a payment method, which is kind of the opposite of the desired behaviour of this generic `payRequest` method... 

an idea would be to allow `swapSettings` on `overrides`? 
Something like this:
`overrides?: ITransactionOverrides | ISwapSettings`. Or, to be even more generic, `overrides?: ITransactionOverrides  | {swapSettings: ISwapSettings}`

packages/payment-processor/src/payment/index.ts
```diff
@@ -25,28 +32,47 @@ export class UnsupportedNetworkError extends Error {
...
   overrides?: ITransactionOverrides,
 ): Promise<ContractTransaction> {
   const signer = getSigner(signerOrProvider);
   const paymentNetwork = getPaymentNetwork(request);
+  if (swapSettings && !canSwapToPay(request)) {
+    throw new UnsupportedNetworkError(paymentNetwork);
```
benjlevesque, line 70: I would add a more specific error messsage... Especially if we keep the breaking change, since this error might happen to people upgrading the library

packages/payment-processor/src/payment/index.ts
```diff
@@ -69,44 +95,86 @@ export async function hasSufficientFunds(
...
+/**
+ * Verifies the address has enough funds to pay an amount in a given currency.
+ *
+ * @param fromAddress the address willing to pay
+ * @param amount
+ * @param currency
+ * @param provider the Web3 provider. Defaults to Etherscan.
+ * @throws UnsupportedNetworkError if network isn't supported
+ */
+export async function isSolvent(
```
benjlevesque, line 127: great change!

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -55,18 +66,70 @@ export async function hasErc20Approval(
...
+    const signer = getSigner(provider);
+    const encodedTx = encodeApproveErc20(request, signer);
+    const tokenAddress = request.currencyInfo.value;
+    const tx = await signer.sendTransaction({
+      data: encodedTx,
+      to: tokenAddress,
+      value: 0,
+      ...overrides,
+    });
+    return tx;
```
benjlevesque, line 123: this seems exactly the same as `approveErc20`?

benjlevesque (approved):
Great work, that was a tough one!

Regarding my comments on the breaking change of `payRequest`, I think I would actually avoid having the swap possibility in `payRequest`, as it is intended to be the generic default behaviour. Paying with swap requires additional, specific parameters. Trying to make the method generic for all cases actually makes it more complicated, I would simply add another method `payRequestWithSwap` or something like this...

packages/payment-processor/src/payment/erc20.ts
yomarion: Oopsie 👍  Thanks

packages/payment-processor/src/payment/index.ts
yomarion: I'll go with your suggestion above and make it a different function then.

packages/payment-processor/src/payment/erc20-fee-proxy.ts
```diff
@@ -44,49 +46,105 @@ export async function payErc20FeeProxyRequest(
...
+      tokenAddress,
+      paymentAddress,
+      amountToPay,
+      `0x${paymentReference}`,
+      feeToPay,
+      feeAddress || constants.AddressZero,
+    ]);
+  }
+
+  // Swap to pay
```
rittme, line 127: This makes this function a lot more complex.
I would suggest creating another module for the swap payments.
Something like `erc20-swap-payment.ts`, and having all the swap logic there, if possible.

Changes like this one, `payErc20Request()`, `approveErc20ForSwapToPayIfNeeded()`, `approveErc20ForSwapToPay()`, the interface `ISwapSettings`, and all things that concern only swapping should be encapsulated on it's own module as much as possible.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -18,34 +18,45 @@ import {
...
 /**
  * Processes a transaction to pay an ERC20 Request.
  * @param request
  * @param signerOrProvider the Web3 provider, or signer. Defaults to window.ethereum.
  * @param amount optionally, the amount to pay. Defaults to remaining amount of the request.
  * @param feeAmount optionally, the fee amount to pay. Only applicable to ERC20 Fee Payment network. Defaults to the fee amount.
+ * @param swapSettings optionally, the settings to swap a maximum amount of currency, through a swap path, before a deadline, to pay
  * @param overrides optionally, override default transaction values, like gas.
  */
 export async function payErc20Request(
```
rittme, line 33: This function signature is starting to look too big. All the `undefined` we have to declare `payRequest` shows an issue.
I think we need something like an `options` object parameter.
The CodeClimate check is not mandatory, but has good hints about this kind of things.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -55,18 +66,61 @@ export async function hasErc20Approval(
...
+
+/**
+ * Checks if a spender has enough allowance from an ERC20 token owner to pay an amount.
+ * @param ownerAddress address of the owner
+ * @param spenderAddress address of the spender
+ * @param provider the web3 provider. Defaults to Etherscan.
+ * @param paymentCurrency ERC20 currency
+ * @param amount
+ */
+export async function checkErc20Allowance(
```
rittme, line 86: Nice change. But mixing allowance and approval can be confusing. I think we should stick to only one name.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -55,18 +66,61 @@ export async function hasErc20Approval(
...
 
 /**
  * Processes the approval transaction of the targeted ERC20.
  * @param request request to pay
  * @param provider the web3 provider. Defaults to Etherscan.
  * @param overrides optionally, override default transaction values, like gas.
- * @param paymentNetworkId the payment network id
- * @param proxyContractAddress the address of the proxy contract to set the approval.
+ */
+export async function approveErc20IfNeeded(
```
rittme, line 107: nice helper :)

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -86,10 +140,67 @@ export async function approveErc20(
...
+ * Processes the approval transaction of a given payment ERC20 to be spent by the swap router,
+ * if the current approval is missing or not sufficient.
+ * @param request request to pay, used to know the network
+ * @param ownerAddress address of the payer
+ * @param paymentCurrency ERC20 currency used for the swap
+ * @param signerOrProvider the web3 provider. Defaults to Etherscan.
+ * @param minAmount ensures the approved amount is sufficient to pay this amount
+ * @param overrides optionally, override default transaction values, like gas.
+ */
+export async function approveErc20ForSwapToPayIfNeeded(
```
rittme, line 152: As I mentioned before, I think this and the next function (among other swap specific code) deserve their own module.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -101,19 +212,32 @@ export function encodeApproveErc20(
...
-  const erc20interface = ERC20Contract.connect(tokenAddress, signer).interface;
-  const encodedApproveCall = erc20interface.functions.approve.encode([
-    proxyAddress,
+/**
+ * Encodes the approval call to approve any erc20 token to be spent, with no limit.
+ * @param tokenAddress the ERC20 token address to approve
+ * @param spenderAddress the address granted the approval
+ * @param signerOrProvider the signer who owns ERC20 tokens
+ */
+export function encodeApproveAnyErc20(
```
rittme, line 228: I'm not a fan of the name.
It encodes the approve method for the token in the parameter, not for any token. This sounds to me that I could use this to approve all tokens, or that it would give me something generic I could use against any token.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -127,7 +251,21 @@ export async function getErc20Balance(
...
+  return getAnyErc20Balance(request.currencyInfo.value, address, provider);
+}
+
+/**
+ * Gets any ERC20 balance of an address
+ * @param anyErc20Address the currency address
+ * @param address the address to check
+ * @param provider the web3 provider. Defaults to Etherscan
+ */
+export async function getAnyErc20Balance(
```
rittme, line 263: Same as previous.

packages/payment-processor/src/payment/index.ts
```diff
@@ -69,44 +117,86 @@ export async function hasSufficientFunds(
...
+  }
+
+  let feeAmount = new BigNumber(0);
+  if (paymentNetwork === ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT) {
+    feeAmount = request.extensions[paymentNetwork].values.feeAmount || 0;
+  }
+  return isSolvent(
+    address,
+    request.currencyInfo,
+    new BigNumber(request.expectedAmount).add(feeAmount),
```
rittme, line 135: Any reason why we dropped bigNumberify and started using BigNumber instead?

packages/payment-processor/src/payment/index.ts
```diff
@@ -69,44 +117,86 @@ export async function hasSufficientFunds(
...
+ * @param fromAddress the address willing to pay
+ * @param amount
+ * @param currency
+ * @param provider the Web3 provider. Defaults to Etherscan.
+ * @throws UnsupportedNetworkError if network isn't supported
+ */
+export async function isSolvent(
+  fromAddress: string,
+  currency: ICurrency,
+  amount: BigNumber,
```
rittme, line 152: For compatibility it's better if we use BigNumberish

packages/payment-processor/src/payment/index.ts
```diff
@@ -69,44 +117,86 @@ export async function hasSufficientFunds(
...
+ * @param address the address holding the funds
+ * @param paymentCurrency if different from the requested currency
+ * @param provider the Web3 provider. Defaults to Etherscan.
+ * @throws UnsupportedNetworkError if the currency is not implemented.
+ */
+async function getBalanceInAnyCurrency(
+  address: string,
+  paymentCurrency: ICurrency,
+  provider: Provider,
+): Promise<BigNumber> {
```
rittme, line 177: For compatibility it's better if we use `BigNumberish`

rittme (commented):
Good stuff! I left a few comments, let me know if you want to discuss.

packages/payment-processor/src/payment/index.ts
```diff
@@ -69,44 +117,86 @@ export async function hasSufficientFunds(
...
-      );
+  if (!provider) {
+    provider = getNetworkProvider(request);
+  }
+
+  let feeAmount = new BigNumber(0);
+  if (paymentNetwork === ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT) {
+    feeAmount = request.extensions[paymentNetwork].values.feeAmount || 0;
+  }
+  return isSolvent(
```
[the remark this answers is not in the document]
yomarion: It's exported and used by any app willing to swap & pay with a decent UX.

packages/payment-processor/src/payment/utils.ts
```diff
@@ -33,7 +34,7 @@ export function getNetworkProvider(request: ClientTypes.IRequestData): Provider
   if (request.currencyInfo.network === 'rinkeby') {
     return getDefaultProvider('rinkeby');
   }
-  throw new Error('unsupported network');
+  throw new UnsupportedNetworkError(request.currencyInfo.network);
```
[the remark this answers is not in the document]
yomarion: I created a `UnsupportedCurrencyNetwork ` error instead

packages/payment-processor/src/payment/erc20.ts
yomarion: Naming-wise, I tried my best to circumvent all the previous asumptions that we pay in request currency. For example `hasErc20Approval` should be renamed `hasRequestErc20Approval` or `hasDefaultErc20Approval` or something like that.

Do you prefer `hasAnyErc20Approval` ?

packages/payment-processor/src/payment/erc20.ts
yomarion: Again, I tried my best to work around all the previous asumptions that we pay in request currency. For example `encodeApproveErc20` should be renamed `encodeApproveRequestErc20`or `encodeApproveDefaultErc20` or something like that.

What do you suggest? I wanted to avoid breaking changes.

packages/payment-processor/src/payment/erc20.ts
yomarion: Suggestions welcome ;)

packages/payment-processor/src/payment/index.ts
yomarion: Laziness... haha

benjlevesque (approved):

packages/payment-processor/src/payment/erc20-fee-proxy.ts
```diff
@@ -50,34 +50,24 @@ export async function payErc20FeeProxyRequest(
  * @param signerOrProvider the Web3 provider, or signer. Defaults to window.ethereum.
  * @param amount optionally, the amount to pay. Defaults to remaining amount of the request.
  * @param feeAmountOverride optionally, the fee amount to pay. Defaults to the fee amount of the request.
+ * @param swapSettings settings for the swap:
```
rittme, line 53: Should be removed

packages/payment-processor/src/payment/erc20-fee-proxy.ts
```diff
@@ -50,34 +50,24 @@ export async function payErc20FeeProxyRequest(
...
  */
 export function encodePayErc20FeeRequest(
   request: ClientTypes.IRequestData,
   signerOrProvider: Web3Provider | Signer = getProvider(),
   amount?: BigNumberish,
   feeAmountOverride?: BigNumberish,
 ): string {
-  validateRequest(request, PaymentTypes.PAYMENT_NETWORK_ID.ERC20_FEE_PROXY_CONTRACT);
-  const signer = getSigner(signerOrProvider);
+  validateErc20FeeProxyRequest(request, amount, feeAmountOverride);
```
rittme, line 61: nice :)

rittme (approved):

PR #344, opened 2020-10-16
feat: add tax amount alongside tax percentage

AdamDowson:
## Description of the changes
Added tax amount (fixed) alongside tax percentage

## Link to Jira

https://app.asana.com/0/1194620076282706/1196258233315537


vnnkl (approved):

yomarion (changes requested):
I think we should update the version from 0.0.2 to 0.0.3.

Also, any idea why the diff is so big compared to the changes?

packages/data-format/src/format/rnf_invoice/rnf_invoice-0.0.2.json
```diff
@@ -133,6 +133,9 @@
           "taxPercent": {
             "type": "number"
           },
+          "taxAmount": {
```
benjlevesque, line 136: open question: should we leave the taxPercent a required field with this new one?

packages/data-format/test/data/example-invalid-0.0.2.json
```diff
@@ -1,80 +1,82 @@
...
-      "note": "payment before chrismas"
-    }
-  },
-
-  "miscellaneous": {
-    "manufacturerCompany": "Victor Hugo & Co.",
-    "deliveryCompany": "Gavroche Express"
-  }
-}
+{
```
benjlevesque, line 1: only formatting for these files?

benjlevesque (commented):
I feel this should be a new version of the format, what do you think @rittme ?

packages/data-format/src/format/rnf_invoice/rnf_invoice-0.0.2.json
rittme: I agree, we have to handle it better. If possible, having both being mutualy exclusive and mandatory could work.
Another option would be having a `taxAmount` field, and a `taxType` field, where we specify if it's percentage or amount.

AdamDowson:
> I feel this should be a new version of the format, what do you think @rittme ?

Yes. I forgot to post a comment, do you guys think this be a new version? Pros and cons to both @rittme @vnnkl @yomarion

packages/data-format/test/data/example-meta-unknown.json
```diff
@@ -1,78 +1,80 @@
...
-      "note": "payment before chrismas"
-    }
-  },
-
-  "miscellaneous": {
-    "manufacturerCompany": "Victor Hugo & Co.",
-    "deliveryCompany": "Gavroche Express"
-  }
-}
+{
```
rittme, line 1: why are we changing everything on these files?

packages/data-format/test/data/example-invalid-0.0.2.json
AdamDowson: I'll try and figure out why the diffs are so crazy here. It was a one line change to the tests

packages/data-format/test/data/example-meta-unknown.json
AdamDowson: It's a formatting issue - taking a look to see why its happened

PR #347, opened 2020-10-20
fix: UDST.transfer

yomarion:
## Description of the changes

* Fix the transfer of USDT in the case of USDT > ERC20 swaps to pay

* Also fix a bug that would occur if we needed to increase the approval of USDT for one of the 2 contracts

### DIFF
--- a/packages/smart-contracts/src/contracts/lib/SafeERC20.sol
+++ b/packages/smart-contracts/src/contracts/lib/SafeERC20.sol
@@ -49,13 +49,6 @@ library SafeERC20 {
     return result;
   }
   
-  /**
-    * @dev Deprecated. This function has issues similar to the ones found in
-    * {IERC20-approve}, and its usage is discouraged.
-    *
-    * Whenever possible, use {safeIncreaseAllowance} and
-    * {safeDecreaseAllowance} instead.
-    */
   function safeApprove(IERC20 _token, address _spender, uint256 _amount) internal returns (bool result) {
     address tokenAddress = address(_token);
     /* solium-disable security/no-inline-assembly */
@@ -89,5 +82,24 @@ library SafeERC20 {
 
     /* solium-enable security/no-inline-assembly */
     return result;
+  }
+
+  
+  bytes4 private constant SELECTOR = bytes4(keccak256(bytes('transfer(address,uint256)')));
+
+  function safeTransfer(IERC20 _token, address _to, uint256 _amount) internal {
+      address tokenAddress = address(_token);
+
+      // check if the address is a contract
+      assembly {
+        if iszero(extcodesize(tokenAddress)) { revert(0, 0) }
+      }
+      
+      (bool success, bytes memory data) = tokenAddress.call(abi.encodeWithSelector(
+          SELECTOR, 
+          _to, 
+          _amount
+      ));
+      require(success && (data.length == 0 || abi.decode(data, (bool))), 'Transfer failed');
   }
 }
--- a/packages/smart-contracts/src/contracts/lib/SafeERC20.sol
+++ b/packages/smart-contracts/src/contracts/lib/SafeERC20.sol
@@ -2,6 +2,7 @@ pragma solidity ^0.5.0;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
+
 /**
  * @title SafeERC20
  * @notice Works around implementations of ERC20 with transferFrom not returning success status.
@@ -13,7 +14,13 @@ library SafeERC20 {
    * @dev This is necessary because of non-standard ERC20 tokens that don't have a return value.
    * @return The return value of the ERC20 call, returning true for non-standard tokens
    */
-  function safeTransferFrom(IERC20  _token, address _from, address _to, uint256 _amount) internal returns (bool result) {
+  function safeTransferFrom(
+    IERC20  _token,
+    address _from,
+    address _to,
+    uint256 _amount
+  ) internal returns (bool result)
+  {
     // solium-disable-next-line security/no-low-level-calls
     (bool success, bytes memory data) = address(_token).call(abi.encodeWithSignature(
       "transferFrom(address,address,uint256)",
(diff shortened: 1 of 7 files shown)


yomarion:
TODO before merging: rinkeby & mainnet artifacts

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
```diff
@@ -90,4 +83,23 @@ library SafeERC20 {
...
+
+  function safeTransfer(IERC20 _token, address _to, uint256 _amount) internal {
+      address tokenAddress = address(_token);
+
+      // check if the address is a contract
+      assembly {
+        if iszero(extcodesize(tokenAddress)) { revert(0, 0) }
+      }
+      
+      (bool success, bytes memory data) = tokenAddress.call(abi.encodeWithSelector(
```
rittme, line 98: Any reason we don't use encodeWithSignature like the others? It's more readable.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
```diff
@@ -90,4 +83,23 @@ library SafeERC20 {
...
   }
+
+  
+  bytes4 private constant SELECTOR = bytes4(keccak256(bytes('transfer(address,uint256)')));
+
+  function safeTransfer(IERC20 _token, address _to, uint256 _amount) internal {
+      address tokenAddress = address(_token);
+
+      // check if the address is a contract
+      assembly {
```
rittme, line 94: we're missing some solium-disable on this method. It will fail on thing like assembly or low level 
calls. 
I recommend adding the solidity package to vscode, so you can have these warnings inline.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
```diff
@@ -90,4 +83,23 @@ library SafeERC20 {
...
+        if iszero(extcodesize(tokenAddress)) { revert(0, 0) }
+      }
+      
+      (bool success, bytes memory data) = tokenAddress.call(abi.encodeWithSelector(
+          SELECTOR, 
+          _to, 
+          _amount
+      ));
+      require(success && (data.length == 0 || abi.decode(data, (bool))), 'Transfer failed');
```
rittme, line 103: Is this replacing what we're doing with assembly on the other functions on this library?
If we're sure it's the same it's great, and we should drop the assembly from the others.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
```diff
@@ -49,13 +49,6 @@ library SafeERC20 {
     return result;
   }
   
-  /**
-    * @dev Deprecated. This function has issues similar to the ones found in
```
rittme, line 53: Why are we removing this? It's not the case anymore?

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: I tried, it did not work... I did not push investigations further on this one.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: I think the deprecation does not concern us because we approve the max every time... But I can put it back.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: It's how uniswap dealt with USDT, fully tested but I cannot say if it's 100% the same as the assembly lines. What I can do is remove the assembly lines of the 2 others, and test (with automated tests but also playing around).

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: I probably went a bit fast on this. But on the Uniswap package, they don't have these warning mutes... And I don't really know how they work TBH, if you do, I will just add it back 👍

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
rittme: It's fine, just trying to understand the context.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
rittme: Those lines are to disable the linter, so it won't fail on these. As long as the lint test doesn't fail, I'm fine with it (and CircleCI too).

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
rittme: EncodeWithSignature should be a short version of what you're doing on the selector.
It would be wise to make sure we understand what we're doing here.

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
rittme: From my understanding it looks the same. But I may be missing something.
If this can replace the assembly it would be great.

rittme (approved):

vnnkl (approved):

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: Done 👍

packages/smart-contracts/src/contracts/lib/SafeERC20.sol
yomarion: Indeed 👍  I got confused by something else apparently. Thanks for the review

packages/smart-contracts/test/contracts/ERC20SwapToPay.js
```diff
@@ -76,44 +76,13 @@ contract('SwapToPay', function(accounts) {
     const finalFromBalance = await paymentErc20.balanceOf(from);
     expect(finalFromBalance.toNumber()).to.equals(initialFromBalance.toNumber());
   }
-  
-  it('can approve bad ERC20 to be swapped by the router', async function() {
```
yomarion, line 80: Moved in a section for bad ERC20 checks

rittme (approved):
Nice. A lot cleaner :)

vnnkl (approved):

PR #351, opened 2020-10-22
WIP: Add smartcontract for proxy change from FIAT to Crypo

vrolland:
## Description of the changes

Proxy Smartcontract to pay a FIAT request (USD, AUD, CHF, EUR, GBP, JPY) in some crypto (DAI, USDT, USDC or SUSD).


packages/smart-contracts/migrations/2_deploy_contracts.js
```diff
@@ -94,7 +100,21 @@ module.exports = async function(deployer) {
     // SwapToPay
     await deployer.deploy(ERC20SwapToPay, FakeSwapRouter.address, ERC20FeeProxy.address);
     console.log('SwapToPay Contract deployed: ' + ERC20SwapToPay.address);
-    
+
+    // Mock Chainlink aggregator
+    await deployer.deploy(MockChainlinkDAIUSD);
```
yomarion, line 105: I think it's time to kick-off a 3rd migration file in order to keep things separate.

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
```diff
@@ -0,0 +1,127 @@
...
+    uint256 amountToPay = computeConversion(_amountFiat, _currencyFiat, _currencyCrypto);
+    require(amountToPay <= _maxCryptoToSpend, "Amount to pay is over the user limit");
+
+    // Make the ERC20 transfer
+    IERC20 erc20 = IERC20(getTokenAddress(_currencyCrypto));
+    require(
+      erc20.allowance(msg.sender, address(this)) >= amountToPay,
+      "allowance needed for the token transfer"
+    );
+    require(erc20.transferFrom(msg.sender, _to, amountToPay), "transferFrom() failed");
```
yomarion, line 62: Isn't it possible to call the existing proxy contract to do do the transfer, let it throw the usual transfer with reference event, so the main blockchain payment detection is kept the same. This contract would send an event to match fiat amount and crypto amount. 

I think there are pros and cons for both kinds of solutions, but we need to consider them and decide.

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
```diff
@@ -0,0 +1,103 @@
...
+
+  event AnswerUpdated(int256 indexed current, uint256 indexed roundId, uint256 updatedAt);
+  event NewRound(uint256 indexed roundId, address indexed startedBy, uint256 startedAt);
+}
+
+
+/**
+ * @title ChainlinkAggregatorCaller for Rinkeby
+ */
+contract ChainlinkAggregatorCaller {
```
yomarion, line 22: Should be a library? Shouldn't we have one library instantiating values depending on the network?

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
```diff
@@ -0,0 +1,127 @@
...
+        .div(1e8);
+    }
+
+    // Then, compute the amount in the right crypto
+    uint256 finalAmount;
+    if (_currencyCrypto == CryptoEnum.ETH || _currencyCrypto == CryptoEnum.DAI) {
+      // Compute the amount from USD to the right crypto
+      // amount * decimal / rate
+      finalAmount = amountInUSD.mul(1e18).div(
+        uint256(getChainlinkAggregatorCryptoToUsd(_currencyCrypto).latestAnswer())
```
yomarion, line 103: We should probably first check that it's not zero

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
```diff
@@ -0,0 +1,127 @@
...
+      );
+    } else {
+      // Compute the amount from USD to ETH
+      // amount * decimal / rate
+      uint256 amountInETH = amountInUSD.mul(1e18).div(
+        uint256(getChainlinkAggregatorCryptoToUsd(CryptoEnum.ETH).latestAnswer())
```
yomarion, line 109: Also here we should check that the rate is not zero

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
```diff
@@ -0,0 +1,127 @@
...
+  event TransferWithReference(
+    address tokenAddress,
+    address to,
+    uint256 amount,
+    bytes indexed paymentReference,
+    uint256 amountinCrypto
+  );
+
+  /**
+   * @notice Performs a ERC20 token transfer with a reference computing the amount regarding a fiat amount
```
yomarion, line 32:
```suggestion
   * @notice Performs an ERC20 token transfer with a reference computing the amount based on a fiat amount
```

yomarion (commented):
Exciting move! Good job, congrats! I'd be happy to discuss a few things, it's very promising.

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
vrolland: The `.div()` is a safeMath function so it is already checked.

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
vrolland: same

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
vrolland: This piece of code is meant to be used only once. Plus, using a library would require two deployments. 
Maybe I am missing something, tho?

packages/smart-contracts/migrations/2_deploy_contracts.js
vrolland: I agree. But this would mean to rearrange the order of deployment and change the address of the contracts. We should do it in a separate PR, no?

packages/smart-contracts/src/contracts/ProxyChangeCryptoFiat.sol
vrolland: We should have an IRL discussion about it

yomarion: IRL
Tradeoff between:
- more info on blockchain
- or more calls to the blockchain nodes (Infura)

To be challenged

packages/smart-contracts/src/contracts/test/MockChainlinkUSDTETH.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.5.0;
+
+
+contract MockChainlinkUSDTETH {
+  function latestAnswer() external view returns (int256) {
+    return 2639000000000000;
```
benjlevesque, line 8: could you add a comment to explain? for all Mocks

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
```diff
@@ -0,0 +1,103 @@
...
+   */
+  function getChainlinkAggregatorFiatToUsd(FiatEnum _currencyFiat)
+    internal
+    pure
+    returns (AggregatorInterface)
+  {
+    if (_currencyFiat == FiatEnum.AUD) {
+      // AUD/USD
+      return AggregatorInterface(0x21c095d2aDa464A294956eA058077F14F66535af);
+    }
```
benjlevesque, line 73: would it be possible to make this contract administrable by its owner? otherwise  adding currencies and changing addresses will be a nightmare

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * @notice Get token address from _currencyCrypto
+   * @param _currencyCrypto crypto currency wanted
+   */
+  function getTokenAddress(CryptoEnum _currencyCrypto) internal pure returns (address) {
+    if (_currencyCrypto == CryptoEnum.DAI) {
+      return 0x5592EC0cfb4dbc12D3aB100b257153436a1f0FEa;
+    }
+    revert("Crypton not supported");
```
benjlevesque, line 101: same as above

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
```diff
@@ -0,0 +1,103 @@
+pragma solidity ^0.5.0;
+
+interface AggregatorInterface {
```
benjlevesque, line 3: could this be in a separate file and imported instead of redefined 3 times?

packages/smart-contracts/src/contracts/lib/RinkebyChainlinkAggregatorCaller.sol
vrolland: I would rather upload a new contract. 
Adding administration there would make the contract less secure and more expensive to call. 
We can open an IRL discussion about it :)

PR #357, opened 2020-11-06
chore: use ethers instead of web3 for Web3 Signature Provider

vrolland:
## Description of the changes

Used of ethers instead of web3 for Web3 signature for Metamask compatibility


AdamDowson (approved):

vnnkl (approved):

packages/web3-signature/package.json
```diff
@@ -43,7 +43,7 @@
   "dependencies": {
     "@requestnetwork/types": "0.27.0",
     "@requestnetwork/utils": "0.26.0",
-    "web3-eth": "1.3.0"
+    "ethers": "5.0.19"
```
benjlevesque, line 46: ethers v5 is a pretty big change compared to v4 afaik, I would recommend using the same version as the other packages in order to avoid conflicts.

benjlevesque (approved):
nice change! ethers rock...

packages/web3-signature/package.json
vrolland: Good point :)

benjlevesque (approved):

packages/web3-signature/test/web3-signature-provider.test.ts
```diff
@@ -16,34 +18,46 @@ const id1Raw = {
...
+  getSigner: jest.fn().mockImplementation(() => ({signMessage: () => {return signatureValueExpected;} }))
+}
 
 /* tslint:disable:no-unused-expression */
 describe('web3-signature-provider', () => {
   describe('sign', () => {
     it('can sign', async () => {
-      const spy = jest.spyOn(mockEth.personal, 'sign');
-      const signProvider = new Web3SignatureProvider('http://localhost:8545');
+      const signProvider = new Web3SignatureProvider(new providers.InfuraProvider());
```
yomarion, line 33: If Infura is only used for the object creation but not actually used, you should precise that in a comment so we don't incriminate this external service if, for example, unit tests fail some day.

yomarion (approved):
Good job

packages/web3-signature/test/web3-signature-provider.test.ts
```diff
@@ -16,35 +18,44 @@ const id1Raw = {
 };
 
 const data = { What: 'ever', the: 'data', are: true };
-const normalizedData = Utils.crypto.normalize(data);
+// const normalizedData = Utils.crypto.normalize(data);
```
rittme, line 21: remove this line?

packages/web3-signature/package.json
```diff
@@ -43,7 +43,7 @@
   "dependencies": {
     "@requestnetwork/types": "0.27.0",
     "@requestnetwork/utils": "0.26.0",
-    "web3-eth": "1.3.0"
+    "ethers": "4.0.48"
```
rittme, line 46: Why not use the latest version?
Ethers 5 is out of beta and is the stable ethers version now.

rittme (approved):
The change looks good to me.
I'm curious about why we needed to drop web3 for this change. Do we have a requirement that web3 doesn't fulfill anymore?

There is an interesting debate on using ethers vs web3 on the entire product. It seems this PR starts to make the argument for ethers.

packages/web3-signature/package.json
vrolland: Because We use the 4.0.48 version in the rest of the code. I followed a @benjlevesque 's suggestion.

If we want to update to the 5. We should do it for all the repo.

PR #360, opened 2020-11-10
fix: use force push to fix lerna broken nightly updates

rittme:
## Description of the changes

Tries to fix lerna canary release (for nightlies) by force-publishing all the changed packages since last release (and not only the ones changed at the last commit).
Based on this thread: https://github.com/lerna/lerna/issues/2060#issuecomment-619403848

This is a hack and should be addressed in the future, probably by migrating to a better tool (like Nx, Rush, multi-semantic-release) or using our own publishing script (simple if we want to use fixed versions for all packages).


yomarion (approved):

vrolland (approved):
I trust you on this.

benjlevesque (approved):

PR #364, opened 2020-11-20
feat: smart contract wallets do not need eth to pay gas

yomarion:
## Description of the changes
* Smart contract wallets do not need to have ETH for the payment are the gas is paid by an EOA

## Link to Asana
https://app.asana.com/0/1199139651874074/1199241446145269/f

### DIFF
--- a/packages/payment-processor/src/payment/index.ts
+++ b/packages/payment-processor/src/payment/index.ts
@@ -139,30 +139,7 @@ export async function isSolvent(
   provider: Provider,
 ): Promise<boolean> {
   const ethBalance = await provider.getBalance(fromAddress);
-  let needsGas:boolean;
-  try {
-    /**
-     * proxyProvider represents some additional data given by Wallet Connect
-     */
-    type proxyProvider = {
-      provider: {
-        wc: {
-          _peerMeta: {
-            name: string;
-          }
-        }
-      }
-    }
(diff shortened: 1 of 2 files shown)


AdamDowson (approved):

packages/payment-processor/src/payment/index.ts
```diff
@@ -139,12 +139,36 @@ export async function isSolvent(
...
+    const thisProvider:proxyProvider = provider as any;
+    // Smart Contract wallets do not pay for the gas
+    const smartContractWallets = [
+      'Safe Multisig WalletConnect',
+    ]
+    needsGas = !smartContractWallets.includes(thisProvider.provider.wc._peerMeta.name);
+  } catch {
+    // Other providers need gas for the transaction
+    needsGas = true;
+  }
```
benjlevesque, line 165: or, in one line: 
```ts
const needsGas  =  (provider as any)?.provider?.wc?._peerMeta?.name != "Safe Multisig WalletConnect";
```

I feel the type is overkill here, and actually creates complexity, since you don't know what can be null. The try catch also feels unnecessary, since the only thing that can fail is null checking, which is easy to solve.

The array for future smart wallets is a good idea, but we can add it when we have other wallets...

benjlevesque (changes requested):

benjlevesque (approved):

PR #365, opened 2020-11-23
WIP: add chainlink conversion path smartcontract

vrolland:
## Description of the changes

Smartcontract to get the rate between two currencies through a path of chainlink aggregators.
This is the first step to pay request in different currencies.

### DIFF
--- a/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
+++ b/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
@@ -15,6 +15,7 @@ module.exports = async function(deployer) {
   await deployer.deploy(AggUSDT_ETH);
   await deployer.deploy(USDT_fake);
 
+  // all these addresses are for test purposes
   const ETH_address = "0x0000000000000000000000000000000000000000";
   const USD_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'USD' });
   const EUR_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'EUR' });
@@ -23,6 +24,7 @@ module.exports = async function(deployer) {
 
   const conversionPathInstance = await deployer.deploy(ChainlinkConversionPath);
 
+  // all these aggregators are for test purposes
   await conversionPathInstance.updateListAggregators( [DAI_address,         EUR_address,        ETH_address,        USDT_address], 
                                                       [USD_address,         USD_address,        USD_address,        ETH_address], 
                                                       [AggDAI_USD.address,  AggEUR_USD.address, AggETH_USD.address, AggUSDT_ETH.address]);
--- a/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
+++ b/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
@@ -1,9 +1,9 @@
 const Utils = require('@requestnetwork/utils').default;
-const AggDAI_USD = artifacts.require("./test/Agg_DAI_USD.sol");
-const AggETH_USD = artifacts.require("Agg_ETH_USD");
-const AggEUR_USD = artifacts.require("Agg_EUR_USD");
-const AggUSDT_ETH = artifacts.require("Agg_USDT_ETH");
-const USDT_fake = artifacts.require("USDT_fake");
+const AggDAI_USD = artifacts.require("AggDaiUsd.sol");
+const AggETH_USD = artifacts.require("AggEthUsd");
+const AggEUR_USD = artifacts.require("AggEurUsd");
+const AggUSDT_ETH = artifacts.require("AggUsdtEth");
+const USDT_fake = artifacts.require("UsdtFake");
 
 const ChainlinkConversionPath = artifacts.require("ChainlinkConversionPath");
 
--- a/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
+++ b/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
@@ -16,16 +16,16 @@ module.exports = async function(deployer) {
   await deployer.deploy(USDT_fake);
 
   // all these addresses are for test purposes
-  const ETH_address = "0x0000000000000000000000000000000000000000";
-  const USD_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'USD' });
-  const EUR_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'EUR' });
-  const DAI_address = '0x6b175474e89094c44da98b954eedeac495271d0f';
+  const ETH_address = Utils.currency.getCurrencyHash({type: 'ETH', value:'ETH'});
+  const USD_address = Utils.currency.getCurrencyHash({type: 'ISO4217', value: 'USD' });
+  const EUR_address = Utils.currency.getCurrencyHash({type: 'ISO4217', value: 'EUR' });
+  const DAI_address = Utils.currency.getCurrencyHash({type: 'ERC20', value:'0x6b175474e89094c44da98b954eedeac495271d0f'});
   const USDT_address = USDT_fake.address;
 
   const conversionPathInstance = await deployer.deploy(ChainlinkConversionPath);
 
   // all these aggregators are for test purposes
-  await conversionPathInstance.updateListAggregators( [DAI_address,         EUR_address,        ETH_address,        USDT_address], 
+  await conversionPathInstance.updateAggregatorsList( [DAI_address,         EUR_address,        ETH_address,        USDT_address], 
                                                       [USD_address,         USD_address,        USD_address,        ETH_address], 
                                                       [AggDAI_USD.address,  AggEUR_USD.address, AggETH_USD.address, AggUSDT_ETH.address]);
 };
--- a/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
+++ b/packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
@@ -1,0 +1,29 @@
+const Utils = require('@requestnetwork/utils').default;
+const AggDAI_USD = artifacts.require("./test/Agg_DAI_USD.sol");
+const AggETH_USD = artifacts.require("Agg_ETH_USD");
+const AggEUR_USD = artifacts.require("Agg_EUR_USD");
+const AggUSDT_ETH = artifacts.require("Agg_USDT_ETH");
+const USDT_fake = artifacts.require("USDT_fake");
+
+const ChainlinkConversionPath = artifacts.require("ChainlinkConversionPath");
+
+// Deploys, set up the contracts
+module.exports = async function(deployer) {
+  await deployer.deploy(AggDAI_USD);
+  await deployer.deploy(AggETH_USD);
+  await deployer.deploy(AggEUR_USD);
+  await deployer.deploy(AggUSDT_ETH);
+  await deployer.deploy(USDT_fake);
+
+  const ETH_address = "0x0000000000000000000000000000000000000000";
+  const USD_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'USD' });
+  const EUR_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'EUR' });
+  const DAI_address = '0x6b175474e89094c44da98b954eedeac495271d0f';
+  const USDT_address = USDT_fake.address;
+
+  const conversionPathInstance = await deployer.deploy(ChainlinkConversionPath);
+
+  await conversionPathInstance.updateListAggregators( [DAI_address,         EUR_address,        ETH_address,        USDT_address], 
+                                                      [USD_address,         USD_address,        USD_address,        ETH_address], 
+                                                      [AggDAI_USD.address,  AggEUR_USD.address, AggETH_USD.address, AggUSDT_ETH.address]);
+};
--- a/packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
+++ b/packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
@@ -18,14 +18,14 @@ interface AggregatorFraction {
 /**
  * @title ChainlinkConversionPath
  *
- * @notice ChainlinkConversionPath is a contract allowing to compute conversion from a path of Chainlink
+ * @notice ChainlinkConversionPath is a contract allowing to compute conversion from a Chainlink path
  */
 contract ChainlinkConversionPath is WhitelistAdminRole {
   using SafeMath for uint256;
 
   uint constant DECIMALS = 1e18;
 
-  // Mapping of Chainlink aggrefators (input currency => output currency => contract address)
+  // Mapping of Chainlink aggregators (input currency => output currency => contract address)
   // input & output currencies are the addresses of the ERC20 contracts OR the sha3("currency code")
   mapping(address => mapping(address => address)) public allAggregators;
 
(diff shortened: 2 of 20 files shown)


packages/utils/src/crypto.ts
```diff
@@ -60,6 +61,20 @@ function keccak256Hash(data: string): string {
...
 
+/**
+ * Hashes with the keccak256 algorithm with a normalization before and formats it
+ *
+ * @notice It will sort the object by keys before hashing
+ *
+ * @param data The data to hash
+ * @returns The hashed data multi-formatted
+ */
+function last20bytesOfNormalizeKeccak256Hash(data: any): string {
```
benjlevesque, line 72:
```suggestion
function last20bytesOfNormalizedKeccak256Hash(data: any): string {
```

benjlevesque: or even, `getCurrencyHash`? isn't it what this does? in which case, `data` can be typed `ICurrency`

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+ * @title ChainlinkConversionPath
+ *
+ * @notice ChainlinkConversionPath is a contract allowing to compute conversion from a path of Chainlink
+ */
+contract ChainlinkConversionPath is WhitelistAdminRole {
+  using SafeMath for uint256;
+
+  uint constant DECIMALS = 1e18;
+
+  // Mapping of Chainlink aggrefators (input currency => output currency => contract address)
```
benjlevesque, line 28:
```suggestion
  // Mapping of Chainlink aggregators (input currency => output currency => contract address)
```

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+  // input & output currencies are the addresses of the ERC20 contracts OR the sha3("currency code")
+  mapping(address => mapping(address => address)) public allAggregators;
+
+  // rate must have been updated before the last... 10min
+  uint256 public maxTimestampDeltaAcceptable = 600;
```
benjlevesque, line 33: can this become an issue? when ethereum is cluttered for instance

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+    emit UpdateAggregator(_input, _output, _aggregator);
+  }
+
+  /**
+    * @notice Update a list of aggregators
+    * @param _inputs list of addresses representing the input currencies
+    * @param _outputs list of addresses representing the output currencies
+    * @param _aggregators list of addresses of the aggregator contracts
+  */
+  function updateListAggregators(address[] calldata _inputs, address[] calldata _outputs, address[] calldata _aggregators)
```
benjlevesque, line 58:
```suggestion
  function updateAggregatorList(address[] calldata _inputs, address[] calldata _outputs, address[] calldata _aggregators)
```

or maybe `updateAggregatorsList`? What's better @AdamDowson?

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+    onlyWhitelistAdmin
+  {
+    require(_inputs.length == _outputs.length, "arrays must have the same length");
+    require(_inputs.length == _aggregators.length, "arrays must have the same length");
+
+    // For every conversions of the path
+    for (uint i; i < _inputs.length; i++) {
+      allAggregators[_inputs[i]][_outputs[i]] = _aggregators[i];
+      emit UpdateAggregator(_inputs[i], _outputs[i], _aggregators[i]);
+    }
```
benjlevesque, line 69: what happens if we want to remove a currency? not sure if it's something that will happen,but it seems here that it's not handled

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+  * @notice Computes the conversion from an amount through a list of conversion
+  * @param _amountIn Amount to convert
+  * @param _path List of addresses representing the currencies for the conversions
+  * @return result the result after all the conversions
```
benjlevesque, line 76: missing doc about `oldestTimestampRate`

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+    // initialize the result
+    result = _amountIn.mul(rate).div(decimals);
+
+    oldestTimestampRate = oldestTimestampRateTemp;
+  }
+
+  /**
+  * @notice Computes the rate from a list of conversion
+  * @param _path List of addresses representing the currencies for the conversions
+  * @return result the result after all the conversions, the oldest rate of the path and the decimals (always 1e18)
```
benjlevesque, line 97: not sure -should this be 3 lines? 
```
@return result
@return oldestTimestampRate
@return decimals
```

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+  function getRate(
+    address[] memory _path
+  )
+    public
+    view
+    returns (uint256 result, uint256 oldestTimestampRate, uint256 decimals)
+  {
+    // initialize the result with 1e18 decimals (for more precision)
+    result = DECIMALS;
+    decimals = DECIMALS;
```
benjlevesque, line 108: if I get this right, you are converting currencies that have less than 18 decimals (USDT for instance) later in the code. What's the point of returning this, if it's always 18?

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
benjlevesque: btw, if `oldestTimestampRate` represents the timestamp of the oldest rate, the name should be `oldestRateTimestamp`.  If it is the Rate of the oldest timestamp, the name is correct.

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    // For every conversions of the path
+    for (uint i; i < _path.length - 1; i++) {
+      (AggregatorFraction aggregator, bool reverseAggregator, uint256 decimalsInput, uint256 decimalsOutput) = getAggregatorAndDecimals(_path[i], _path[i + 1]);
+
+      // store the lastest timestamp of the path
```
benjlevesque, line 114:
```suggestion
      // store the latest timestamp of the path
```
But didn't you mean oldest instead of latest?

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+
+      // mul with the difference of decimals before the rate computation (for more precision)
+      if (decimalsAggregator > decimalsInput) {
+        result = result.mul(10**(decimalsAggregator-decimalsInput));
+      }
+      if (decimalsAggregator < decimalsOutput) {
+        result = result.mul(10**(decimalsOutput-decimalsAggregator));
+      }
+
+      // Apply the rate (if path use an aggregator in the reverse way, div instead of mul)
```
benjlevesque, line 133:
```suggestion
      // Apply the rate (if path uses an aggregator in the reverse way, div instead of mul)
```

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+        result = result.div(10**(decimalsAggregator-decimalsOutput));
+      }
+    }
+  }
+
+  /**
+  * @notice Gets aggregators and decimals of two currencies
+  * @param _input input Address
+  * @param _output output Address
+  * @return aggregator and decimals
```
benjlevesque, line 154: it's worth saying something about reverseAggregators imo

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+  function decimals() external view returns (uint8);
+  function latestAnswer() external view returns (int256);
+  function latestTimestamp() external view returns (uint256);
+}
+
+
+/**
+ * @title ChainlinkConversionPath
+ *
+ * @notice ChainlinkConversionPath is a contract allowing to compute conversion from a path of Chainlink
```
rittme, line 21:
```suggestion
 * @notice ChainlinkConversionPath is a contract allowing to compute conversion from a Chainlink path
```

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+
+interface AggregatorFraction {
+  function decimals() external view returns (uint8);
+  function latestAnswer() external view returns (int256);
+  function latestTimestamp() external view returns (uint256);
+}
+
+
+/**
+ * @title ChainlinkConversionPath
```
rittme, line 19: I'm not sure what exactly we mean by conversion, but it's too generic to be clear. Whould it be clearer if we call it `rate`, `exchange rate` or `conversion rate`?

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
rittme: I feel the idea was to update this later, with `result` (so we know the amount of decimals we should use), but it has been forgotten.

rittme (approved):

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
yomarion: There is a choice here: when the network is cluttered, should we reduce the risk of abuse (and UX) by forcing the payer to wait or the opposite? 
I think that at the contract level we should keep a trust-base logic, and maybe this threshold should be set in upper layers.

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+
+  // Mapping of Chainlink aggrefators (input currency => output currency => contract address)
+  // input & output currencies are the addresses of the ERC20 contracts OR the sha3("currency code")
+  mapping(address => mapping(address => address)) public allAggregators;
+
+  // rate must have been updated before the last... 10min
+  uint256 public maxTimestampDeltaAcceptable = 600;
+
+  // declare a new aggregator
+  event UpdateAggregator(address _input, address _output, address _aggregator);
```
yomarion, line 36:
```suggestion
  event AggregatorUpdate(address _input, address _output, address _aggregator);
```

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
yomarion: I agree, we should anticipate that

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+  */
+  function getConversion(
+    uint256 _amountIn,
+    address[] calldata _path
+  )
+    external
+    view
+    returns (uint256 result, uint256 oldestTimestampRate)
+  {
+    (uint256 rate, uint256 oldestTimestampRateTemp, uint256 decimals) = getRate(_path);
```
yomarion, line 86:
```suggestion
    (uint256 rate, uint256 timestamp, uint256 decimals) = getRate(_path);
```
just for an easier read

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -0,0 +1,211 @@
...
+    // by default we assume it is FIAT so 8 decimals
+    decimalsOutput = 8;
+    // if address is 0, then it's ETH
+    if (_output == address(0x0)) {
+      decimalsOutput = 18;
+    } else if (isContract(_output)) {
+      // otherwise, we get the decimals from the erc20 directly
+      decimalsOutput = ERC20fraction(_output).decimals();
+    }
+  }
```
yomarion, line 192: I think it's worth factorizing this part of the code

packages/smart-contracts/test/contracts/ChainlinkConversionPath.js
```diff
@@ -0,0 +1,180 @@
...
+const address1 = "0x1111111111111111111111111111111111111111";
+const address2 = "0x2222222222222222222222222222222222222222";
+const address3 = "0x3333333333333333333333333333333333333333";
+const address4 = "0x4444444444444444444444444444444444444444";
+const address5 = "0x5555555555555555555555555555555555555555";
+const address6 = "0x6666666666666666666666666666666666666666";
+
+const ETH_address = "0x0000000000000000000000000000000000000000";
+const USD_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'USD' });
+const EUR_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'EUR' });
```
yomarion, line 16: Shoulnd't we have a util for fiat currencies "addresses"?

packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
```diff
@@ -0,0 +1,31 @@
...
+  await deployer.deploy(AggEUR_USD);
+  await deployer.deploy(AggUSDT_ETH);
+  await deployer.deploy(USDT_fake);
+
+  // all these addresses are for test purposes
+  const ETH_address = "0x0000000000000000000000000000000000000000";
+  const USD_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'USD' });
+  const EUR_address = Utils.crypto.last20bytesOfNormalizeKeccak256Hash({type: 'ISO4217', value: 'EUR' });
+  const DAI_address = '0x6b175474e89094c44da98b954eedeac495271d0f';
+  const USDT_address = USDT_fake.address;
```
yomarion, line 23: It would be cleaner to deploy these only on the local network. Even if we don't use deployers for actual deployments, it helps understanding.

yomarion (changes requested):
Very happy that this subject gets tackled finally! Once the feedback is integrated, it seems future-proof 👍

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
AdamDowson: Yes, `updateAggregatorsList` fits here (sorry btw I just saw this review)

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
benjlevesque: my understanding is that this smart contract is no longer about converting rates, but providing the value, which btw, could be an issue: we will probably want to display the rate on the UI

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
vrolland: This constant is actually not used. I forgot to remove it.

This mechanism will be set up on the next contract. 

I will open a discussion about it.

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
vrolland: to remove an aggregator: 
`updateAggregator(_input, _output, '0x0')`

packages/smart-contracts/migrations/3_deploy_chainlink_contracts.js
vrolland: We do the same in `2_deploy_contracts.js`

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
vrolland: We can open a discussion about it. 
I wanted to add it in case later the decimals change regarding the currencies but we want to keep the same interface.

vnnkl (approved):

PR #371, opened 2020-12-03
build: get rid of duplicate open-zepellin contract dependency

rittme:
## Description of the changes
`@openzeppelin/contracts` and `openzeppelin-solidity` are the same package.
We were using both on our codebase. This PR removes the older alias `openzeppelin-solidity`.


benjlevesque (approved):
I think the reason I left it there was because i didn't want to touch the contract - since we will not release a new version just for that, it made sense to leave it as it is deployed 

But if it's only renaming, no problem

PR #394, opened 2020-12-22
refactor: externalize currency tools in a new packages

vrolland:
## Description of the changes

- Move the currency tools to a new package
- Remove depedency to  ̀currency-iso`.


packages/currency/package.json
```diff
@@ -0,0 +1,66 @@
...
+  "devDependencies": {
+    "@types/jest": "26.0.13",
+    "@typescript-eslint/parser": "4.1.1",
+    "eslint": "7.9.0",
+    "eslint-plugin-spellcheck": "0.0.17",
+    "eslint-plugin-typescript": "0.14.0",
+    "jest": "26.4.2",
+    "lint-staged": "10.3.0",
+    "mocha": "8.1.3",
+    "nyc": "15.1.0",
```
benjlevesque, line 57: Those 2 are no longer used. We should remove them from `utils` too, it's a mistake that they're still there.

packages/currency/src/iso4217/data-iso4217.ts
```diff
@@ -0,0 +1,1356 @@
+/*
+	Follows ISO 4217, https://www.iso.org/iso-4217-currency-codes.html
+	See https://www.currency-iso.org/dam/downloads/lists/list_one.xml
+	Data last updated 2018-08-29
+*/
+/* eslint-disable spellcheck/spell-checker */
+export default [
```
benjlevesque, line 7: any reason not to no longer use the iso currency package ? would avoid us to require to update this list manually, even if I guess this won't change any time soon :)

packages/currency/src/others/others.ts
```diff
@@ -0,0 +1,14 @@
+export default {
```
benjlevesque, line 1: maybe the folder is unnecessary?  `packages/currency/src/others.ts`

packages/currency/src/iso4217/data-iso4217.ts
```diff
@@ -0,0 +1,1356 @@
+/*
```
benjlevesque, line 1: maybe the folder is necessary? `packages/currency/src/iso4217.ts`

yomarion (approved):
Well done!

benjlevesque (commented):
Nice! This will be very useful....

I think you miss the update to the CI (unit tests)

Some suggestions on easier code browsing (most of it comes from previous structure, but it's a good opportunity to do some cleanup) 
- rename `erc20/networks/networks.ts` to `erc20/networks/index.ts`
- rename `erc20/erc20.ts` to `erc20/index.ts`
- delete file extra-data.ts, and move the value to `erc20/networks/mainnet.ts` (this requires some small changes for consistance with the other "custom" currencies)

packages/currency/package.json
vrolland: @benjlevesque What about the file .nycrc?

packages/currency/src/iso4217/data-iso4217.ts
vrolland: As we are getting closer to the protocol this must not change or with a new version of the request logic.
So we cannot depend on an external library.
I don't have the same concern on the ERC20 currencies as the number of decimal is define directly on the smart contract of the token.

AdamDowson (approved):

packages/currency/src/iso4217/data-iso4217.ts
benjlevesque: fair enough, but that means reviewing this list every now and then. Maybe we should subscribe to changes to the original repo we were using

benjlevesque (approved):

packages/currency/test/currency.test.ts
```diff
@@ -44,6 +44,7 @@ describe('api/currency', () => {
     });
 
     it('returns INDA', () => {
+      // console.log(getAllSupportedCurrencies().ERC20);
```
yomarion, line 47: Forgot that?

yomarion (commented):
Great job, it can save lives 👍 Isn't it time to give the currency logic an object oriented dev approach? (new Currency('DAI', 'rinkeby')).decimals would help me a lot

AdamDowson (approved):

PR #410, opened 2021-02-17
fix: mock storage confirming too fast causes concurrency issue

benjlevesque:
I have the case where I reuse the mock several times for unit testing, and the `confirmed` event is sent before we listen for it


yomarion (approved):
I don't really know what can the impacts be, so we'd rather wait for Vincent's review

vrolland (approved):
LGTM :)

benjlevesque:
Closing for now, the issue is more complex and increasing the timeout isn't always sufficient

benjlevesque:
Closing for now, the issue is more complex and increasing the timeout isn't always sufficient

PR #411, opened 2021-02-17
fix: enable tree-shaking on data-format

benjlevesque:
When using a bundler (like [ncc](https://github.com/vercel/ncc)), the JSON files are not included in the final bundle. This prevents important optimizations

NB: the same should be done for smart-contracts artifacts later


vrolland (approved):
LGTM

PR #430, opened 2021-03-04
chore: Enums to Unions

benjlevesque:
Unions are more powerful than enums in TS.
To avoid breaking changes, we keep the export of the former enums as variables.

Recommended reading (and inspiration): https://rlee.dev/writing/stop-misusing-typescript-string-enums


yomarion (commented):
💯

PR #438, opened 2021-03-12
Documenting Dai based req burner.

GrandGarcon:
## Description of the changes
did documentation of the functions , data vars  being used in the   DAI-REQ burner and   readme doc


GrandGarcon:
as its being  completed on the side of @yomarion  , i will be removing the branch

PR #439, opened 2021-03-12
Dai lock for burner

GrandGarcon:
## Description of the changes

Modified the  Open hash submitter and associated contracts on Xdai .


GrandGarcon (commented):
All good on the cleaning and deployment part . thanks for the assistance .

PR #443, opened 2021-03-17
feat: tool to list all the aggregators supported by the protocol

vrolland:


GrandGarcon (approved):

benjlevesque (approved):

packages/toolbox/README.md
```diff
@@ -37,6 +37,40 @@ yarn run:create
...
+#   network: {
+#      "currencyIn": {
+#        "currencyOut": "aggregator"
+#        ...
+#      }
+#      ...
+#   }
+#   ...
+# }
+# All aggregators nodes for dijkstra: (Used to populate @requestnetwork/currency)
```
yomarion, line 60:
```suggestion
# All aggregators nodes for the currency pairs graph: (Used to populate @requestnetwork/currency)
```
I know that I am picky with that but Dijkstra is how we do it. One could populate the graph without knowing about Dijkstra algo.

packages/toolbox/src/chainlinkConversionPathTools.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+  }
+
+  // tslint:disable:no-console
+  console.log('#####################################################################');
+  console.log('All aggregators:');
+  console.log(allAggregators);
+  console.log('#####################################################################');
+  console.log('All aggregators nodes for dijkstra:');
```
yomarion, line 137:
```suggestion
  console.log('All aggregators nodes for currency pairs graph:');
```

yomarion (approved):
Nice tooling ! It will prove super handful, we should do the same for swap to pay proxies.

PR #450, opened 2021-03-19
chore: prettier & lint

benjlevesque:
- run prettier on all files to avoid formatting issues that became a problem on many PRs
- remove `tslint`, which is deprecated, and replace with `eslint`. ⚠I did not try to replicate all rules. Instead I decided to go with simpler, recommended, rules. 
- CI now runs the lint once for all package, instead of once per package. This will ensure an early failure in case of lint issue 
- CI now runs prettier check as well as linting. 
- removed spellcheck, it was a nightmare
- as much as possible, replaced `require` with `import`

Relevant changes (no, I don't expect a review on all 315 changed files...)
- [CI config changes](https://github.com/RequestNetwork/requestNetwork/pull/450/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47)
- [ES lint changes](https://github.com/RequestNetwork/requestNetwork/pull/450/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.log&file-filters%5B%5D=.ts#diff-e7e195d3d8fa5b82fa51ed952f3a179d429cbc46c0432ce1f7b357c54c162822)
- Check package.json & other config changes on 1 package, they are the same for all packages.


package.json
```diff
@@ -17,23 +17,39 @@
     "build": "lerna run build",
     "clean": "lerna run clean",
     "build:tsc": "tsc -b packages/advanced-logic packages/data-access packages/data-format packages/epk-decryption packages/epk-signature packages/ethereum-storage packages/integration-test packages/multi-format packages/payment-detection packages/prototype-estimator packages/request-client.js packages/request-logic packages/request-node packages/smart-contracts packages/toolbox packages/transaction-manager packages/types packages/usage-examples packages/utils packages/web3-signature packages/currency",
-    "lint": "lerna run lint",
-    "lint-staged": "lerna run lint-staged",
+    "lint": "eslint packages/**/src/**/*.ts --quiet",
```
benjlevesque, line 20: `quiet` is there to ignore warnings for now, there are hundreds of them. We could disable them from config but I feel it's good to have them in the IDE

yomarion (approved):
LGTM, I don't have relevant background on lint to challenge further but the removal of spellcheck is a life saver, and the rest looks cleaner now ! Nice

vrolland (approved):
Very good PR. (hard to read tho... :smile:)

Less require. No more spellcheck. CI simplify. No more tslint.

It's a go for me ! LGTM !

PR #453, opened 2021-03-23
feat: add the graph token

siibars:
## Description of the changes
Adding the graph token 

## Etherscan Link for the graph token
https://etherscan.io/token/0xc944e90c64b2c07662a292be6244bdf05cda44a7


vrolland (changes requested):
The token is already on metamask: https://github.com/MetaMask/contract-metadata/blob/af1cb193183b0e474564d55bdb3657de13136faf/contract-map.json#L2575

There is no need to add it here.

PR #458, opened 2021-03-25
build: fix next-release

benjlevesque:


benjlevesque:
seems it is not required

PR #463, opened 2021-03-26
feat: graceful shutdown of request-node

benjlevesque:
## Description of the changes
Currently, shuting down a node running in Docker takes long time. 

It is best practice to handle a graceful shutdown of an application, and reduces the deployment time significantly

unrelated change: removed some `any` that itched me 😁


packages/request-node/docker/docker-tsconfig.json
```diff
@@ -14,7 +14,12 @@
     "noUnusedLocals": true,
```
benjlevesque, line 14: fixes the Docker build (with the Dockerfile available in the monorepo)

yomarion (approved):
Good thing!

vrolland (approved):

PR #467, opened 2021-03-26
chore: use ethers for ethereum-storage

benjlevesque:
Motivations
- `ethers` is used everywhere in the codebase
- web3 causes the node to crash if connection breaks
- types and code readability

Additionally, this allows to remove hd-wallet-provider which limits the number of dependencies to maintain


packages/request-node/src/storageUtils.ts
```diff
@@ -29,33 +28,33 @@ export function getEthereumStorage(
...
-          onTimeout: false,
-        },
-      }),
-    });
+
+  let provider: providers.Provider;
+  const url = config.getStorageWeb3ProviderUrl();
+  if (url.match('^wss?://.+')) {
+    provider = new providers.WebSocketProvider(url);
+    // FIXME previous WS config
```
benjlevesque, line 36: @alexandre-abrioux any idea what we should do here?

packages/integration-test/test/node-client.test.ts
```diff
@@ -494,6 +495,31 @@ describe('Request client using a request node', () => {
     const requests = await badRequestNetwork.fromTopic(myRandomTopic);
     expect(requests).toHaveLength(0);
   });
+
+  it('can create multiple requests in parallel', async () => {
```
benjlevesque, line 499: I added this test because it seemed that the node wouldn't accespt multiple transactions at once.

packages/request-node/src/storageUtils.ts
```diff
@@ -29,33 +28,33 @@ export function getEthereumStorage(
...
+    //   reconnect: {
+    //     auto: true,
+    //     delay: 3000, // ms
+    //     maxAttempts: 5,
+    //     onTimeout: false,
+    //   },
+    // }
   } else {
-    provider = new HDWalletProvider(mnemonic, config.getStorageWeb3ProviderUrl());
+    provider = new providers.StaticJsonRpcProvider(url);
```
benjlevesque, line 51: using a StaticJsonRpcProvider allows to reduce the number of calls (eth_chainId for instance)

packages/ethereum-storage/src/smart-contract-manager.ts
```diff
@@ -272,133 +260,82 @@ export default class SmartContractManager {
...
-      (await gasPriceDefiner.getGasPrice(StorageTypes.GasPriceType.FAST, this.networkName));
+      (await gasPriceDefiner.getGasPrice(StorageTypes.GasPriceType.STANDARD, this.networkName));
 
     // parse the fees parameters to hex bytes
-    const feesParametersAsBytes = web3Utils.padLeft(
-      web3Utils.toHex(feesParameters.contentSize),
-      LENGTH_BYTES32_STRING,
+    const feesParametersAsBytes = utils.hexZeroPad(
+      utils.hexlify(feesParameters.contentSize),
+      LENGTH_BYTES32_STRING / 2,
```
benjlevesque, line 268: because of the way hexZeroPad works

packages/request-node/src/storageUtils.ts
alexandre-abrioux: You can do this:
```ts
provider = new providers.Web3Provider(
  new Web3WsProvider(url, {
    //  previous WS config
  }),
)
```
; or, if we want to completely remove web3js dependencies, I can take care of it in another PR, for now just remove the `if (wss)` block or throw an "unsupported" error) as `ethers.js` does not handle auto-reconnect, cf. https://github.com/ethers-io/ethers.js/issues/1053

benjlevesque: the point is to remove web3 completely yes

alexandre-abrioux: We should use this solution then: https://github.com/ethers-io/ethers.js/issues/1053#issuecomment-808736570 ,  we just need to abstract the provider in it's own module.

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -71,8 +71,8 @@ export default class EthereumStorage implements StorageTypes.IStorage {
    */
   public constructor(
     externalBufferUrl: string,
-    ipfsGatewayConnection?: StorageTypes.IIpfsGatewayConnection,
-    web3Connection?: StorageTypes.IWeb3Connection,
+    ipfsGatewayConnection: StorageTypes.IIpfsGatewayConnection,
```
vrolland, line 74: No risk of breaking backward compatibility ?

packages/request-node/src/storageUtils.ts
vrolland: I discovered it with ct. I should have told you so.

vrolland (approved):
Awesome PR ! Well done!

packages/ethereum-storage/src/ethereum-storage.ts
benjlevesque: ethereum storage is only used by the node, so no

packages/ethereum-storage/src/smart-contract-manager.ts
```diff
@@ -272,133 +260,82 @@ export default class SmartContractManager {
...
     // use it for the different events (error, transactionHash, receipt and confirmation)
-    return new Promise((resolve, reject) => {
-      // This boolean is set to true once the ethereum metadata has been created and the promise has been resolved
-      // When set to true, we use it to ignore next confirmation event function call
-      let ethereumMetadataCreated = false;
-
-      // Keep the transaction hash for future needs
-      let transactionHash = '';
-      const transactionParameters = {
+    // This boolean is set to true once the ethereum metadata has been created and the promise has been resolved
```
yomarion, line 274: ??

packages/ethereum-storage/test/smartcontract-manager.test.ts
```diff
@@ -184,94 +175,87 @@ const txPerBlockConfiguration5 = [100, 200, 150, 400, 1, 670, 300, 140, 20, 600]
...
 
-    expect(events[0].returnValues.hash).toEqual(hashStr);
-    expect(events[0].returnValues.hashSubmitter).toEqual(addressRequestHashSubmitter);
-    expect(events[0].returnValues.feesParameters).toEqual(realSizeBytes32Hex);
+    expect(events[0].args.hash).toEqual(hashStr);
+    expect(events[0].args.hashSubmitter.toLowerCase()).toEqual(addressRequestHashSubmitter);
+    expect(events[0].args.feesParameters).toEqual(realSizeBytes32Hex);
   });
 
   // TODO since the migration to jest, this test fails.
```
yomarion, line 231:
```suggestion
```
Good one :)

packages/ethereum-storage/test/smartcontract-manager.test.ts
```diff
@@ -289,19 +273,15 @@ describe('SmartContractManager', () => {
...
-    smartContractManager.ethereumBlocks.getBlock = (_block: any): any => {
-      return {
-        number: 9,
-      };
-    };
+      });
+    jest
+      .spyOn(smartContractManager.ethereumBlocks.provider, 'getBlockNumber')
+      .mockImplementation(() => Promise.resolve(9));
+    // TODO ? smartContractManager.ethereumBlocks = new EthereumBlocks(mockEth, 1, 0, 0);
```
yomarion, line 284: reminder

yomarion (approved):
Not easy to review but everything looks good to me, nice move !

bertux (approved):
big good job :+1:
I wondered if renaming eth as provider would make it more confusing later because it's more generic word but it's also the word used by etherjs so it's good for now.

packages/ethereum-storage/test/ethereum-blocks.test.ts
```diff
@@ -103,34 +104,30 @@ const mockBlocksEthereum = [
   9807,
   9906,
 ];
+const provider = new providers.JsonRpcProvider('http://localhost:8545');
```
alexandre-abrioux, line 107: We could do this to not depend on any external service for unit tests
(but this would reintroduce some web3 dependencies 😆)
```ts
import ganache from "ganache";
const provider = new ethers.providers.Web3Provider(ganache.provider());
```

alexandre-abrioux (approved):
Amazing refacto 💯

packages/ethereum-storage/src/config.ts
```diff
@@ -61,15 +58,15 @@ export function getDefaultIpfs(): StorageTypes.IIpfsGatewayConnection {
  * @returns the url to connect to the network
  */
 export function getDefaultEthereumProvider(): string {
-  return config.ethereum.nodeUrlDefault[config.ethereum.default].url;
+  return config.ethereum.nodeUrlDefault.private.url;
```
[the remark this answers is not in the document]
benjlevesque: yes, this was only to fix the type, I reverted and used `as const` on the `default` network

KolevDarko (approved):

MantisClone:
Is this PR still relevant?

PR #469, opened 2021-03-30
feat: optionally disable events

benjlevesque:
when these events are not used, they can cause side effects, like leakage in tests

### DIFF
--- a/packages/request-client.js/src/api/request.ts
+++ b/packages/request-client.js/src/api/request.ts
@@ -60,6 +60,10 @@ export default class Request {
    * if true, skip the payment detection
    */
   private skipPaymentDetection = false;
+  /**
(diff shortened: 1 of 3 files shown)


packages/request-client.js/src/api/request.ts
```diff
@@ -60,6 +60,10 @@ export default class Request {
    * if true, skip the payment detection
    */
   private skipPaymentDetection = false;
+  /**
+   * if true, do not send events
```
yomarion, line 64:
```suggestion
   * if true, do not send blockchain confirmation events (on creation, approval, etc.)
```

yomarion (approved):
Good workaround, well done that you found the issue

PR #477, opened 2021-04-08
feat: mainnet conversion contract and aggregators

yomarion:
## Description of the changes
* Added the reference to the mainnet contracts for conversion
* Referenced the aggregators


benjlevesque (approved):
🚀

PR #483, opened 2021-04-13
chore: Token and Currency classes

yomarion:
Motivation: ease the use of the different helpers we have for currencies.

This is a first step, we should migrate all the similar utilities to class or object method, IMO.

* chore: Token an Currency classes design
* feat: fetch a currency from its address with `Currency.from(...)`

### DIFF
--- a/packages/currency/src/currency.ts
+++ b/packages/currency/src/currency.ts
@@ -1,0 +1,79 @@
+import { RequestLogicTypes } from '@requestnetwork/types';
+import { currencyToString } from '.';
+import { getSupportedERC20Currencies, getErc20Currency } from './erc20';
+import iso4217 from './iso4217';
+
+/**
+ * @class Currency implements ICurrency with helpers
+ * Represents a currency supported by the Request Logic, with minimum required
+ * information: value, type and network (optional).
+ */
+export class Currency implements RequestLogicTypes.ICurrency {
+  public value: string;
+  public type: RequestLogicTypes.CURRENCY;
+  public network?: string;
+  constructor(currency: RequestLogicTypes.ICurrency) {
+    ({ value: this.value, type: this.type, network: this.network } = currency);
+  }
+
+  /**
+   * Gets a supported currency from a symbol, symbol-network or address.
+   * Iterates over all the supported networks if needed
+   * @param symbolOrAddress e.g. 'DAI', 'FAU', 'FAU-rinkeby' or '0xFab46E002BbF0b4509813474841E0716E6730136'
+   * @returns an ICurrency object
+   */
+  static from(symbolOrAddress: string): Currency {
+    try {
+      const currencyFromSymbol = this.fromSymbol(symbolOrAddress);
+      return currencyFromSymbol;
+    } catch (e) {
+      const erc20Currencies = getSupportedERC20Currencies();
+      const currencyFromAddress = erc20Currencies.find((c) => c.value === symbolOrAddress);
+      if (!currencyFromAddress) {
+        throw new Error(`The currency ${symbolOrAddress} does not exist or is not supported`);
+      }
+      return new Currency(currencyFromAddress);
+    }
+  }
+
+  /**
+   * Get currency from its symbol and network.
+   * @param symbol
+   * @param network
+   * @returns RequestLogicTypes.ICurrency
+   */
+  static fromSymbol = (symbol: string, network?: string): Currency => {
+    // Check if it's a supported cryptocurrency
+    if (symbol === 'BTC') {
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.BTC,
+        value: 'BTC',
+      });
+    }
+    if (symbol === 'ETH') {
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.ETH,
+        value: 'ETH',
+      });
+    }
+
+    // Check if it's an ERC20 token and return it if found
+    const erc20Currency = getErc20Currency(symbol, network);
+    if (erc20Currency) {
+      return new Currency(erc20Currency);
+    }
+
+    // Check if it's one of ISO4217 currencies
+    if (iso4217.find((i) => i.code === symbol)) {
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.ISO4217,
+        value: symbol,
+      });
+    }
+    throw new Error(`The currency ${symbol} is not supported`);
+  };
+
+  public toString(): string {
+    return currencyToString(this);
+  }
+}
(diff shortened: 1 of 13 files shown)


packages/currency/src/currency.ts
```diff
@@ -0,0 +1,79 @@
...
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.BTC,
+        value: 'BTC',
+      });
+    }
+    if (symbol === 'ETH') {
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.ETH,
+        value: 'ETH',
+      });
```
benjlevesque, line 57: imo BTC and ETH should have a network

packages/currency/src/currency.ts
```diff
@@ -0,0 +1,79 @@
...
+      return new Currency({
+        type: RequestLogicTypes.CURRENCY.ISO4217,
+        value: symbol,
+      });
+    }
+    throw new Error(`The currency ${symbol} is not supported`);
+  };
+
+  public toString(): string {
+    return currencyToString(this);
```
benjlevesque, line 77: I would move the logic from `currencyToString` here, and make `currencyToString` use this method. Actually `currencyToString` should become deprecated

packages/currency/src/token.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @class Token extends Currency to reference the symbol
+ */
+export class Token extends Currency {
+  constructor(
+    public value: string,
+    public type: RequestLogicTypes.CURRENCY,
+    public symbol: string,
+    public network?: string,
```
benjlevesque, line 13: the parent class already has those fields 

```suggestion
    value: string,
    type: RequestLogicTypes.CURRENCY,
    public readonly symbol: string,
    network?: string,
```

packages/currency/src/currency.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * @class Currency implements ICurrency with helpers
+ * Represents a currency supported by the Request Logic, with minimum required
+ * information: value, type and network (optional).
+ */
+export class Currency implements RequestLogicTypes.ICurrency {
+  public value: string;
+  public type: RequestLogicTypes.CURRENCY;
+  public network?: string;
```
benjlevesque, line 14:
```suggestion
  public readonly value: string;
  public readonly type: RequestLogicTypes.CURRENCY;
  public readonly network?: string;
```

packages/currency/src/erc20/index.ts
```diff
@@ -8,12 +8,19 @@ import { supportedNetworks, supportedNetworksDetails, ERC20SymbolDetails } from
...
   symbol: string,
-  network = 'mainnet',
+  network?: string,
 ): RequestLogicTypes.ICurrency | undefined {
   // Check if it's on one of the other supported networks
-  if (network in supportedNetworks && supportedNetworks[network].has(symbol)) {
+  if (network && network in supportedNetworks && supportedNetworks[network].has(symbol)) {
     return supportedNetworks[network].get(symbol);
   }
+  for (network of Object.keys(supportedNetworks)) {
```
benjlevesque, line 17: if `network` is specified and not found in the previous condition, it should return

packages/currency/src/erc20/index.ts
```diff
@@ -8,12 +8,19 @@ import { supportedNetworks, supportedNetworksDetails, ERC20SymbolDetails } from
...
+    // console.log(network, symbol);
```
benjlevesque, line 18:
```suggestion
```

packages/currency/src/erc20/index.ts
```diff
@@ -8,12 +8,19 @@ import { supportedNetworks, supportedNetworksDetails, ERC20SymbolDetails } from
...
+    if (supportedNetworks[network].has(symbol)) {
+      // console.log(supportedNetworks[network].get(symbol));
```
benjlevesque, line 20:
```suggestion
```

packages/currency/src/token.ts
```diff
@@ -0,0 +1,56 @@
...
+  ) {
+    super({ value, type, network });
+  }
+
+  /**
+   * Gets a supported currency from a symbol, symbol-network or address.
+   * Iterates over all the supported networks if needed
+   * @param symbolOrAddress e.g. 'DAI', 'FAU', 'FAU-rinkeby' or '0xFab46E002BbF0b4509813474841E0716E6730136'
+   * @returns an ICurrency object
```
benjlevesque, line 22: copy paste? I don't correct it because i don't see the point of these `@returns` comments...

packages/currency/src/token.ts
```diff
@@ -0,0 +1,56 @@
...
+        currencyFromAddress.symbol.split('-')[1] || 'mainnet',
+      );
+    }
+  }
+
+  /**
+   * Same result as Currency.toString() but with a faster implementation because symbol is known
+   * @returns
+   */
+  toString = (): string => {
```
benjlevesque, line 52: consistency with other class methods
```suggestion
  toString(): string {
```

packages/currency/src/token.ts
```diff
@@ -0,0 +1,56 @@
...
+        RequestLogicTypes.CURRENCY.ERC20,
+        currencyFromAddress.symbol.split('-')[0],
+        currencyFromAddress.symbol.split('-')[1] || 'mainnet',
+      );
+    }
+  }
+
+  /**
+   * Same result as Currency.toString() but with a faster implementation because symbol is known
+   * @returns
```
benjlevesque, line 50:
```suggestion
```

benjlevesque (commented):
great change!

packages/currency/src/index.ts
```diff
@@ -150,6 +125,7 @@ export function getAllSupportedCurrencies(): {
  * @param currency
  *
  * @returns the hash of the currency
+ * @deprecated use new Currency().getHash() instead
```
benjlevesque, line 128: should this method call `new Currency(currency).getHash()` to avoid having 2 implementations and a risk of divergence?

benjlevesque (approved):

siibars (approved):

packages/currency/src/index.ts
yomarion: You're right, I forgot. Thanks for spotting !

packages/currency/src/currency.ts
```diff
@@ -43,17 +50,22 @@ export class Currency implements RequestLogicTypes.ICurrency {
    * @returns RequestLogicTypes.ICurrency
    */
   static fromSymbol = (symbol: string, network?: string): Currency => {
+    if (symbol === '') {
+      throw new Error(`Cannot guess currency from empty symbol.`);
+    }
     // Check if it's a supported cryptocurrency
-    if (symbol === 'BTC') {
+    if (symbol === 'BTC' && (!network || network === 'mainnet')) {
```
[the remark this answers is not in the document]
yomarion: Yes, as well as NEAR, BNB, etc.

PR #484, opened 2021-04-15
chore: currency.getDecimals() and fixes

yomarion:
chore: deprecate getDecimalsForCurrency, replaced with currency.getDecimals()
fix: Currency.from() for lower case addresses


packages/currency/test/currency.test.ts
```diff
@@ -204,27 +204,6 @@ describe('api/currency', () => {
     });
   });
 
-  describe('Currency.fromSymbol()', () => {
```
yomarion, line 207: Just moved down for consistency

benjlevesque (approved):

PR #490, opened 2021-04-23
fix: Token.from('FAU-rinkeby')

yomarion:

### DIFF
--- a/packages/currency/src/token.ts
+++ b/packages/currency/src/token.ts
@@ -22,11 +22,14 @@ export class Token extends Currency {
    */
   static from(symbolOrAddress: string): Token {
     try {
-      const currencyFromSymbol = this.fromSymbol(symbolOrAddress);
+      const currencyFromSymbol = this.fromSymbol(
+        symbolOrAddress.split('-')[0],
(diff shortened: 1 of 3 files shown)


vrolland (approved):

packages/currency/src/token.ts
```diff
@@ -22,11 +22,14 @@ export class Token extends Currency {
...
-      const currencyFromSymbol = this.fromSymbol(symbolOrAddress);
+      const currencyFromSymbol = this.fromSymbol(
+        symbolOrAddress.split('-')[0],
+        symbolOrAddress.split('-')[1],
+      );
       return new Token(
         currencyFromSymbol.value,
         currencyFromSymbol.type,
-        symbolOrAddress,
+        symbolOrAddress.split('-')[0],
```
benjlevesque, line 32: I don't  get this part

if I do Token.from("FAU-rinkeby") it will work, but with Token.from("0xabcd") it won't?

yomarion: There is a try catch, we first try to guess from a token code. If it fails, we try an address.

PR #492, opened 2021-04-27
refactor: smart-contracts artifacts

benjlevesque:
## Proposal of refactoring of the smart-contracts artifacts

The ERC20Proxy contract artifact was used as an example to simplify this PR. 
If accepted, all other contracts need to be migrated before this PR is merged

### Motivation
- code duplication
- no `Artifact` type, which makes it harder to re-use  artifacts in other packages (ie, in `payment-processor` , we need maps of artifacts depending the on PaymentNetwork)
- remove dynamic `requires` which are bad for bundling.
- better Developer Experience thanks to generated types 

### Structure
The new artifact replaces both the artifact as defined currenlty and the generated `*__factory`  types

The artifact metadata and ABI are now together in a single folder.
Artifacts metadata are now stored in a TS file (instead of JSON), ABI remains in JSON. 

### Usage
```ts
import { erc20ProxyArtifact } from "@request-network/smart-contracts"
erc20ProxyArtifact.getAddress("private") 
erc20ProxyArtifact.getAddress("private","0.1.0") 
await erc20ProxyArtifact.connect('private').transferFromWithReference(...)
await erc20ProxyArtifact.connect('private', Wallet.fromMnemonic(...)).transferFromWithReference(...)
await erc20ProxyArtifact.connect('private', Wallet.fromMnemonic(...), "0.1.0").transferFromWithReference(...)
```

Before this change, the equivalent was:
```ts
import { erc20ProxyArtifact } from "@request-network/smart-contracts"
import { ERC20Proxy__factory } from '@request-network/smart-contracts/types/factories/ERC20Proxy__factory';

ERC20Proxy__factory.connect(erc20ProxyArtifact.getAddress("private")).transferFromWithReference(...)

```

### Optional change
it is possible to get strong typing on the `network` and the `version` parameters, like this:
```ts
erc20ProxyArtifact.getAddress("private","0.1.0") // strong typing on `private` and `0.1.0`
```
however, this might be annoying for some calls like `currency.network` which is a string


packages/smart-contracts/src/lib/artifacts/ERC20Proxy/index.ts
```diff
@@ -0,0 +1,28 @@
+import { ContractArtifact } from '../../ContractArtifact';
```
benjlevesque, line 1: this is a typical contract artifact. All others will look like this

packages/smart-contracts/test/lib/artifact.ts
```diff
@@ -0,0 +1,49 @@
+import { BigNumber, providers } from 'ethers';
```
benjlevesque, line 1: it won't be necessary to have more tests, this ensures the utilities are working and the types are correct

vrolland (approved):

packages/smart-contracts/package.json
```diff
@@ -32,10 +32,9 @@
     "types"
   ],
   "scripts": {
```
yomarion, line 34: It would be nice to document the new scripts in `README.md`, even briefly

yomarion (approved):
Super nice change, awesome ! Well done.

PR #497, opened 2021-05-04
feat: defaultProvider

benjlevesque:


packages/payment-detection/src/provider.ts
```diff
@@ -1,21 +1,34 @@
...
 
-export const getDefaultProvider = (network?: string | ethers.providers.Network | undefined) => {
-  if (network === 'private') {
-    return new ethers.providers.JsonRpcProvider();
-  }
-  if (configuration.infura) {
-    return new ethers.providers.InfuraProvider(network, process.env.RN_INFURA_KEY);
+export const initPaymentDetectionProvider = (overrides: {
+  /** */
+  blockchainRpcs?: Record<string, string>;
```
vrolland, line 15: `BlockchainRpcs` **might** be confusing as it's only for Ethereum forks.

packages/payment-detection/src/provider.ts
```diff
@@ -1,21 +1,34 @@
...
+): providers.Provider => {
+  if (typeof network === 'string') {
+    const rpc = blockchainRpcs[network];
+    if (rpc) {
+      return new providers.JsonRpcProvider(rpc);
+    }
   }
 
-  return ethers.getDefaultProvider(network);
+  return ethers.getDefaultProvider(network, { ...defaultProviderOptions, ...options });
```
vrolland, line 34: Default of default can be hard to understand for new comers.
But I am personally comfortable with that.

packages/payment-detection/src/provider.ts
```diff
@@ -1,21 +1,34 @@
...
-let configuration = {
+let defaultProviderOptions: Record<string, string> = {
   // fallback to Ethers v4 default projectId
   infura: process.env.RN_INFURA_KEY || '7d0d81d0919f4f05b9ab6634be01ee73',
 };
 
-export const initPaymentDetectionProvider = (config: Partial<typeof configuration>) => {
-  configuration = { ...configuration, ...config };
+let blockchainRpcs: Record<string, string> = {
+  matic: 'https://rpc-mainnet.matic.network/',
```
vrolland, line 10: missing other network like xDai & private ?

vrolland (approved):

packages/payment-detection/src/provider.ts
benjlevesque: forgot to comment, thx

packages/payment-detection/src/provider.ts
benjlevesque: you mean passing a param to `getDefaultProvider`? it's the way `ethers` does it. 

This wrapper enables NOT passing the param every time you call `getNetworkProvider`, but instead sharing the config

packages/payment-detection/src/provider.ts
benjlevesque: private, done. 
xdai & others will be added when relevant. This is only used for payment networks FYI

yomarion (approved):
Nice

vrolland (approved):
LGTM :)

PR #506, opened 2021-05-07
chore: reuse artifact in ERC20 proxy detection

benjlevesque:


packages/payment-detection/src/erc20/proxy-contract.ts
```diff
@@ -196,24 +174,24 @@ export default class PaymentNetworkERC20ProxyContract implements PaymentTypes.IP
...
-    if (!PROXY_CONTRACT_ADDRESS_BY_VERSION_BY_NETWORK[paymentNetworkVersion]) {
+    let proxyContractAddress: string;
+    let proxyCreationBlockNumber: number;
+    try {
+      const info = erc20ProxyArtifact.getDeploymentInformation(network, paymentNetworkVersion);
+      proxyContractAddress = info.address;
+      proxyCreationBlockNumber = info.creationBlockNumber;
+    } catch (e) {
+      console.log(e.message);
+      if (e.message?.startsWith('No deployment for network')) {
```
vrolland, line 185: We should have a better management error. It can be tricky to maintain.

But it's a for another PR.

vrolland (approved):

packages/payment-detection/src/erc20/proxy-contract.ts
benjlevesque: indeed, but honestly this is mostly not to break any tests. not a huge deal if the behaviour changes here

yomarion (approved):

PR #509, opened 2021-05-11
feat: add tokens and refacto custom ERC20s

benjlevesque:


packages/currency/test/currency.test.ts
```diff
@@ -441,6 +452,24 @@ describe('api/currency', () => {
       expect(() => Currency.fromSymbol('XXXXXXX')).toThrow();
     });
 
+    it('supports a token that exists on two chains', () => {
```
yomarion, line 455: Nice !

PR #514, opened 2021-05-19
fix: invoice total types + simplify/fix some deps

benjlevesque:


packages/data-format/package.json
```diff
@@ -39,8 +39,8 @@
     "test:watch": "yarn test --watch"
   },
   "dependencies": {
-    "@ethersproject/bignumber": "5.0.15",
```
benjlevesque, line 42: I'll verify the impact of changing this dependency on the size of the package, but it seems using these packages cause conflicts sometimes

packages/data-format/src/format/rnf_invoice/types.ts
```diff
@@ -63,8 +63,3 @@ export interface Address {
   region?: string;
   'street-address'?: string;
 }
-
-export interface Notes {
-  type: string;
-  message: string;
-}
```
benjlevesque, line 70: mistake from previous PR, not part of the standard

packages/ethereum-storage/package.json
```diff
@@ -49,8 +49,8 @@
     "keyv": "4.0.3",
     "node-fetch": "2.6.1",
     "shelljs": "0.8.4",
-    "web3-eth": "1.3.4",
-    "web3-utils": "1.3.4",
+    "web3-eth": "1.3.6",
+    "web3-utils": "1.3.6",
```
benjlevesque, line 53: for coherence with other packages

yomarion (approved):

PR #527, opened 2021-06-24
fix: amount padding for conversion of crypto amounts

yomarion:
This PR fixes the bug that makes it impossible to pay a request denominated in crypto with another crypto with on-payment conversion.

### DIFF
--- a/packages/payment-processor/src/payment/utils.ts
+++ b/packages/payment-processor/src/payment/utils.ts
@@ -276,13 +276,14 @@ export const padAmountForChainlink = (amount: BigNumberish, currency: Currency) 
       decimalPadding = Math.max(chainlinkFiatDecimal - currency.getDecimals(), 0);
       break;
     }
+    case RequestLogicTypes.CURRENCY.ETH:
     case RequestLogicTypes.CURRENCY.ERC20: {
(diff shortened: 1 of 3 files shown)


packages/payment-processor/src/payment/utils.ts
```diff
@@ -264,3 +264,27 @@ export function getAmountToPay(
...
+      decimalPadding = Math.max(chainlinkFiatDecimal - currency.getDecimals(), 0);
+      break;
+    }
+    case RequestLogicTypes.CURRENCY.ERC20: {
+      decimalPadding = 0;
+      break;
+    }
+    default:
+      throw new Error(
+        'Unsupported request currency for conversion with Chainlink. The request currency has to be fiat or ERC20.',
```
vrolland: Maybe give a CTA like "contact us for missing currency" ?

vrolland (approved):

packages/payment-processor/src/payment/utils.ts
yomarion: Currencies are constraints by Request Invoicing, users won't see these errors.

benjlevesque (approved):

PR #528, opened 2021-06-24
feat: add third party to the declarative payment network

vrolland:
I have added `from` in the event to make the difference between declaration from the thirdpary and the payee/payer.


packages/advanced-logic/src/extensions/payment-network/declarative.ts
```diff
@@ -203,11 +205,14 @@ export default class DeclarativePaymentNetwork<
     actionSigner: IdentityTypes.IIdentity,
     timestamp: number,
   ): ExtensionTypes.IState {
-    if (!requestState.payer) {
-      throw Error(`The request must have a payer`);
+    if (!requestState.payer && !extensionState.values.thirdparty) {
+      throw Error(`The request must have a payer or a thirdparty`);
```
yomarion, line 209: I'm a bit surprised by this condition, a declarative request allows the third party to detect that the payment has been made, which is an action done by the issuer, not the payer. Am I missing something?

packages/advanced-logic/src/extensions/payment-network/declarative.ts
```diff
@@ -299,11 +309,14 @@ export default class DeclarativePaymentNetwork<
...
+    if (!requestState.payee && !extensionState.values.thirdparty) {
+      throw Error(`The request must have a payee or a thirdparty`);
     }
-    if (!Utils.identity.areEqual(actionSigner, requestState.payee)) {
-      throw Error(`The signer must be the payee`);
+    if (
+      !Utils.identity.areEqual(actionSigner, requestState.payee) &&
+      !Utils.identity.areEqual(actionSigner, extensionState.values.thirdparty)
+    ) {
+      throw Error(`The signer must be the payee or the thirdparty`);
```
yomarion, line 319: I think it's worth factorizing here

packages/advanced-logic/src/extensions/payment-network/declarative.ts
```diff
@@ -347,11 +361,14 @@ export default class DeclarativePaymentNetwork<
     actionSigner: IdentityTypes.IIdentity,
     timestamp: number,
   ): ExtensionTypes.IState {
-    if (!requestState.payer) {
-      throw Error(`The request must have a payer`);
+    if (!requestState.payer && !extensionState.values.thirdparty) {
+      throw Error(`The request must have a payer or a thirdparty`);
```
yomarion, line 365: Same remark as above

packages/advanced-logic/src/extensions/payment-network/declarative.ts
```diff
@@ -347,11 +361,14 @@ export default class DeclarativePaymentNetwork<
...
     }
-    if (!Utils.identity.areEqual(actionSigner, requestState.payer)) {
-      throw Error(`The signer must be the payer`);
+    if (
+      !Utils.identity.areEqual(actionSigner, requestState.payer) &&
+      !Utils.identity.areEqual(actionSigner, extensionState.values.thirdparty)
+    ) {
+      throw Error(`The signer must be the payer or the thirdparty`);
```
yomarion, line 371: I don't think the third-party should have payer privileges. Maybe there should be a payee third party and a payer third party (in a later version ?) but one actor can only delegate his own action to a third-party, in my opinion.

Let's take bank accounts, if a third-party appointed by the issuer can make sure the refunds has been sent, it cannot know on behalf of the payer that the money has been received.

(this comment applies for all the functions below where the third party can impersonate the payer)

yomarion (commented):
Nice ! I wonder if I miss something about payer impersonation or if just disagree, but we may have to talk a bit :)

packages/advanced-logic/src/extensions/payment-network/declarative.ts
benjlevesque: @yomarion this action is a "payment sent" action

packages/advanced-logic/src/extensions/payment-network/declarative.ts
benjlevesque: this is in line with my earlier comment about multiple thirdparties

packages/types/src/payment-types.ts
```diff
@@ -86,6 +87,7 @@ export interface IPaymentNetworkEvent<TEventParameters> {
   name: EVENTS_NAMES;
   parameters?: TEventParameters;
   timestamp?: number;
+  from?: IIdentity;
```
benjlevesque, line 90: this will be used only by declarative, right? 

I think adding a comment is required here

benjlevesque (commented):
looks good! We need to decide on @yomarion's comment about the thirdparty privileges.

packages/advanced-logic/src/extensions/payment-network/declarative.ts
yomarion: Yes, I know. I think there are flows, it's not because the issuer trusts the 3rd party that the payer has to trust it.

packages/advanced-logic/src/extensions/payment-network/declarative.ts
yomarion: OK, are we in line also to make the multiple third parties in another release ?

packages/types/src/payment-types.ts
vrolland: Only for declarative for now. But could be use somewhere else if needed later.

PR #531, opened 2021-06-28
docs: delegation instead of third party in declarative pn

vrolland:


packages/advanced-logic/specs/payment-network-any-declarative-0.2.0.md
```diff
@@ -347,6 +355,88 @@ The 'addPaymentInstruction' event:
 
 ---
 
+#### addPayeeDelegate
```
benjlevesque, line 358: the 2 methods could be merged in 1 `addDelegate`, which would either add a payee or payer delegate depending on who did the action

benjlevesque (approved):
Not sure if this should be specified, but the delegates should probably be in the Topics of the Request

benjlevesque (approved):

packages/advanced-logic/specs/payment-network-any-declarative-0.2.0.md
```diff
@@ -347,6 +355,50 @@ The 'addPaymentInstruction' event:
...
+- The signer is the `payee` and the extension property `payeeDelegate` is undefined
+- The signer is the `payer` and the extension property `payerDelegate` is undefined
+
+##### Warnings
+
+None.
+
+##### Results
+
+A extension state is updated with the following properties:
```
yomarion, line 383:
```suggestion
An extension state is updated with the following properties:
```

yomarion (commented):
Just to be sure, the file will actually be renamed 0.3.0, correct ?

vrolland:
> Not sure if this should be specified, but the delegates should probably be in the Topics of the Request

I think it should not be specified here, but you are right I should be in the code !

yomarion (approved):

PR #534, opened 2021-06-30
Fix/request network packages version upgrade

siibars:
## Link to Jira


vrolland (changes requested):
This PR doesn't make sense to me.
We must only refer to the current version of the package in the repository. 

I guess what we need here is a publish of the last changes, no ?

PR #539, opened 2021-07-05
feat: Native tokens & cUSD case fix

benjlevesque:
This PR introduces Native tokens, mostly of type `ETH` as they are EVM-compatible chains. 
This is not ideal since it doesn't allow for representation of tokens like NEAR


yomarion (approved):

packages/currency/src/erc20/index.ts
```diff
@@ -98,7 +61,7 @@ export function getSupportedERC20Tokens(): ERC20TokenDetails[] {
           network: networkName,
           decimals: token.decimals,
           name: token.name,
-          symbol: `${token.symbol}${networkName !== 'mainnet' ? `-${networkName}` : ''}`,
+          symbol: token.symbol,
```
benjlevesque, line 64: the place where it matters to have the format `symbol-network` is the request's `currency` field, where it uses the `toString()` method; this method still appends the network.

yomarion (approved):

vrolland (approved):

bertux (approved):

PR #542, opened 2021-07-08
native payment multichains

benjlevesque:
- fix: Celo token names
- lower case tokens
- introduce native tokens
- feat: Native tokens, cUSD case fix, misc currency refactoring
- revert unnecessary change
- RIN + native changes
- feat: native token payment for evm chains

PR #546, opened 2021-07-22
feat: FTM token in Currency package

yomarion:


packages/currency/src/currency.ts
```diff
@@ -90,8 +90,6 @@ export class Currency implements RequestLogicTypes.ICurrency {
     switch (this.type) {
       case RequestLogicTypes.CURRENCY.BTC:
       case RequestLogicTypes.CURRENCY.ETH:
-        symbol = this.type;
-        break;
       case RequestLogicTypes.CURRENCY.ISO4217:
```
yomarion, line 93: Validate this please

benjlevesque:
will include this in #542

PR #548, opened 2021-07-29
specs: interpretation for smart escrow v0

yomarion:
**No need to merge**


packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Description
+
+This is an additional interpretation for the payment network "ERC20 With Fee", allowing the ERC20 payments to be locked on an escrow contract.
+
+The payment is made beforehand to the smart escrow, which locks the funds. The payer can then unlock the funds for a certain target date, after which the request issuer can withdraw the funds. In case of collaboration conflict, the payer can also decide to freeze the payment, which triggers a one-year period freeze. After a freeze, only the payer can withdraw the funds.
+
+The escrow contract pulls ERC20 tokens from the user, there should be no manual transfer.
+
+The escrow contract relies on a proxy when funds are withdrawn. The proxy is responsible for sending `TransferWithReferenceAndFee` events and process the payment to the issuer.
```
benjlevesque, line 15: can the proxy be administrated?

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Network | Contract Address |
+| ------- | ---------------- |
+| Mainnet | TODO             |
+| Rinkeby | TODO             |
+| Private | TODO             |
+
+## Interpretation
+
+When a request is paid with escrow, the events `TransferWithReferenceAndFee` are interepreted the same way and from the same contract.
```
benjlevesque, line 40:
```suggestion
When a request is paid with escrow, the events `TransferWithReferenceAndFee` are interpreted the same way and from the same contract.
```

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+The payment is made beforehand to the smart escrow, which locks the funds. The payer can then unlock the funds for a certain target date, after which the request issuer can withdraw the funds. In case of collaboration conflict, the payer can also decide to freeze the payment, which triggers a one-year period freeze. After a freeze, only the payer can withdraw the funds.
+
+The escrow contract pulls ERC20 tokens from the user, there should be no manual transfer.
+
+The escrow contract relies on a proxy when funds are withdrawn. The proxy is responsible for sending `TransferWithReferenceAndFee` events and process the payment to the issuer.
+
+The contract should only emit payment events when funds are unlocked or frozen by the payer.
+
+## Contract
```
benjlevesque, line 19: this is WIP right?
it misses details on the function to withdraw.

I also think the payee should be able to reject a payment, which would have the consequence of sending back the funds to the payer, with no delay

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+The equivalent of the contract function `transferFromWithReferenceAndFee` in ERC20 proxy is `lockWithReference` and it takes the same 6 arguments:
```
benjlevesque, line 21: shouldn't there be a date from which the funds are available to the payee?

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+**TODO: These specifications are work in progress and might change a lot before they are fully supported by the library.**
+
+Prerequisite: Payment Network "ERC20 With Fee" (see [here](./payment-network-erc20-fee-proxy-contract-0.1.0.md)).
+
+## Description
+
+This is an additional interpretation for the payment network "ERC20 With Fee", allowing the ERC20 payments to be locked on an escrow contract.
+
+The payment is made beforehand to the smart escrow, which locks the funds. The payer can then unlock the funds for a certain target date, after which the request issuer can withdraw the funds. In case of collaboration conflict, the payer can also decide to freeze the payment, which triggers a one-year period freeze. After a freeze, only the payer can withdraw the funds.
```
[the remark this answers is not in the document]
Dadogg80: Hi, @benjlevesque. I believe the point with the "lockperiod" is to incentivice both parties to find a solution, before initiating the dispute/lockperiod. 
So, the period not administrated at the moment as it should only be used as a "last resort" option.

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
Dadogg80: That is a good point. should we priorotize this for the MVP1? or can we have it as a MVP2 feature?

Dadogg80 (approved):

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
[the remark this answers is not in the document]
vrolland: I agree with @benjlevesque .

But as a POC it is good enough to let it as hardcoded.

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
vrolland: Two way of thinking this:
- administrated make easier maintenance (avoid to redeploy new smartcontract when the proxy need to be modify)
- administrated mean less decentralization (people need to trust Request to not sneaky change the proxy to a malicious one)

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
yomarion: Good point for payment rejection !

For the function to withdraw, we used to only specify the functions that impact interpretation but here I agree that it's a critical function and it should be documented properly.

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
yomarion: I think the low frequency of proxy updates (especially: with the same function signature) makes it less relevant. We've never used it for Swap.

packages/advanced-logic/specs/payment-network-erc20-escrow-0.1.0.md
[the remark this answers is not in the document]
yomarion: I don't think we should hardcode or configure this, I think it should be set on funds on the call to unlock. (a parameter is missing in the function signature below)

PR #554, opened 2021-08-04
feat: get all addresses for an artifact

benjlevesque:


packages/smart-contracts/src/lib/ContractArtifact.ts
```diff
@@ -86,19 +99,36 @@ export class ContractArtifact<TContract extends Contract> {
...
     if (!info) {
-      throw Error(`No deployment for network: ${networkName}`);
+      throw Error(`No deployment for network: ${networkName}.`);
     }
     return info;
   }
+
+  /**
+   * Retrieve the deployment information from the artifact of the used version
+   * deployed into the specified network
```
vrolland, line 126: Add a small sentence to explain that this one return null instead of throw.

vrolland (approved):

packages/smart-contracts/src/lib/ContractArtifact.ts
benjlevesque: it's right below in `returns`

PR #556, opened 2021-08-04
fix: use proxy address based on pn version

benjlevesque:
An important fix: without this, the payment always occur on the most recent deployed contract. 

This was discovered with the new matic erc20FeeProxy. 

This PR also includes a refactoring to "prepare" the transaction parameters, to be re-used in apps that need the proxy contract address
Removed unnecessary signer from `encode` methods.


Dadogg80 (approved):

vrolland (approved):
Looks very nice!

bertux (approved):

yomarion (approved):

Dadogg80 (approved):

PR #557, opened 2021-08-06
fix: chainlink paths on Matic

benjlevesque:
trying to understand why the path cannot be found for the path USD => DAI-matic.

the changes to the toolbox are to enable it on matic, since matic RPC's don't support wide block range.


packages/toolbox/src/commands/chainlink/listAggregators.ts
```diff
@@ -10,5 +10,16 @@ export const builder = (): yargs.Argv<IOptions> =>
       type: 'string',
       desc: 'Network for which to list aggregators (mainnet, rinkeby, private)',
     },
+    web3Url: {
+      demand: false,
+      type: 'string',
+      desc: 'A Web3 RPC url. If not provided will use the default',
+    },
+    lastBlock: {
```
yomarion, line 18: NB: this trick is not enough for Matic, there are already >100'000 blocks since the contract got released. It means we cannot compute paths automatically, unless we only update inputs / outputs / aggregators by full batch. Acceptable for now, considering the low costs, but might lead to errors.

yomarion (commented):
The actual problem was not the code but on-chain, cf. this transaction: https://polygonscan.com/tx/0x65dcc7d038211fc8918a3ef7e3889186bfe4e144e824cf735fa1a52f2d849bd5

vrolland (approved):

PR #561, opened 2021-08-13
feat: detect payments on near

yomarion:

### DIFF
--- a/packages/payment-detection/src/eth/near-info-retriever.ts
+++ b/packages/payment-detection/src/eth/near-info-retriever.ts
@@ -1,38 +1,51 @@
 import { PaymentTypes } from '@requestnetwork/types';
-import ETHInfoRetriever from '../eth/info-retriever';
 import { Client } from 'pg';
 
 /**
  * Gets a list of transfer events for an address and payment reference
  */
-export default class NearInfoRetriever extends ETHInfoRetriever {
+export default class NearInfoRetriever {
   private contractName: ReturnType<typeof getContractName>;
   private connectionString: ReturnType<typeof getIndexerConnectionString>;
   /**
-   * @param address Address to check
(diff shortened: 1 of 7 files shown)


packages/payment-detection/src/eth/near-info-retriever.ts
```diff
@@ -0,0 +1,128 @@
+import { PaymentTypes } from '@requestnetwork/types';
+import ETHInfoRetriever from '../eth/info-retriever';
+import { Client } from 'pg';
+
+/**
+ * Gets a list of transfer events for an address and payment reference
+ */
+export default class NearInfoRetriever extends ETHInfoRetriever {
```
benjlevesque, line 8: I'm not sure it's useful to extend ETHInfoRetrieve, what do you reuse?

benjlevesque (approved):

packages/payment-detection/src/eth/near-info-retriever.ts
yomarion: OK I'll adapt. I initially thought there would be more common logic

kevindavee (approved):

packages/payment-detection/src/eth/near-info-retriever.ts
```diff
@@ -0,0 +1,125 @@
...
+      ),
+    }));
+    return events;
+  }
+
+  /**
+   * Documentation: https://github.com/near/near-indexer-for-explorer/blob/master/docs/near-indexer-for-explorer-db.png
+   */
+  private async getTransactionsFromNearIndexerDatabase(): Promise<NearIndexerTransaction[]> {
+    const query = `SELECT t.transaction_hash as "txHash",
```
vrolland, line 49: Look like an SQL fancy query !

PR #564, opened 2021-08-19
chore: improved how we detect aggregators on Matic

yomarion:
## Description of the changes
* Added `startBlock` to `chainlinkPath` util in the `toolbox`, to handle chains with more than 100'000 blocks since the contract deployment


packages/toolbox/README.md
```diff
@@ -71,6 +71,20 @@ To get only aggregators of one network:
 yarn chainlinkPath mainnet
 ```
 
+To get aggregators of a network with a lot of blocks, you have to gather most updates with updateAggregatorsList.
+
+Case 1: you only use updateAGgregatorsList, and update every pair each time, take this block of its last execution.
```
benjlevesque, line 76:
```suggestion
Case 1: you only use updateAggregatorsList, and update every pair each time, take this block of its last execution.
```

benjlevesque (approved):

PR #565, opened 2021-08-20
feat: process near payments

yomarion:
## Description of the changes
Process Near payments with the `near-api-js` library.

Breaking changes: `connector` is encapsulated within `connectOptions` for methods `hasSufficientFunds` and `isSolvent`.

### DIFF
--- a/packages/payment-processor/src/payment/utils-near.ts
+++ b/packages/payment-processor/src/payment/utils-near.ts
@@ -8,7 +8,6 @@ export async function isValidNearAddress(nearNetwork: Near, address: string) {
     await nearNetwork.connection.provider.query(`account/${address}`, '');
     return true;
   } catch (e) {
-    console.warn(e.stack);
     return false;
   }
 }
--- a/packages/payment-processor/src/payment/utils-near.ts
+++ b/packages/payment-processor/src/payment/utils-near.ts
@@ -1,0 +1,75 @@
+import { BigNumber, BigNumberish, ethers } from 'ethers';
+import { Contract } from 'near-api-js';
+import { Near, WalletConnection } from 'near-api-js';
+
+export async function isValidNearAddress(nearNetwork: Near, address: string) {
+  try {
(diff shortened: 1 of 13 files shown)


packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -0,0 +1,75 @@
...
+      GAS_LIMIT,
+      amount,
+    );
+    return;
+  } catch (e) {
+    throw new Error(`Could not pay Near request. Got ${e.message}`);
+  }
+}
+
+// TODO: factorize with payment-detection
```
yomarion, line 65: I'm on it, it's a minor change

packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -0,0 +1,68 @@
+import { BigNumber, BigNumberish, ethers } from 'ethers';
+import { Contract } from 'near-api-js';
+import { Near, WalletConnection } from 'near-api-js';
+import { Near as NearPaymentDetection } from '@requestnetwork/payment-detection';
+
+export async function isValidNearAddress(nearNetwork: Near, address: string) {
+  try {
+    await nearNetwork.connection.provider.query(`account/${address}`, '');
```
benjlevesque, line 8: that seems a very slow way to validate a near address?

packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -0,0 +1,68 @@
...
+    return true;
+  } catch (e) {
+    console.warn(e.stack);
```
benjlevesque, line 11: I don't think a validator should log, it's expected to throw

benjlevesque (approved):

packages/payment-processor/src/payment/utils-near.ts
yomarion: It's the one Near team pushed so...

PR #569, opened 2021-08-23
chore: update openzeppelin without interface changes

vrolland:


packages/smart-contracts/src/contracts/Bytes.sol
```diff
@@ -10,7 +10,7 @@ library Bytes {
     * @notice Extract a bytes32 from a bytes.
     * @param data bytes from where the bytes32 will be extract
     * @param offset position of the first byte of the bytes32
-    * @return address
+    * @return bs
```
benjlevesque, line 13: what does bs mean? perfect place to describe it

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.4.25 <0.7.0;
+pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/math/SafeMath.sol";
```
benjlevesque, line 4: Why this change? 
is math "safe" by default in newer solidity versions?

benjlevesque (approved):

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
Dadogg80: Hi. Yes, SafeMath is included from solidity version 0.8.4..

Dadogg80 (approved):
Looks good.

packages/smart-contracts/src/contracts/ERC20FeeProxy.sol
```diff
@@ -17,12 +17,16 @@ contract ERC20FeeProxy {
   );
 
   // Fallback function returns funds to the sender
-  function() external payable {
+  fallback() external payable {
     revert("not payable fallback");
   }
```
yomarion, line 22: Do you think we still need a fallback ? My understanding is that since Solidity 0.6 if `receive()` is implemented and we try to call a function that does not exist, it will fail, hence I don't see the need for a "manual" revert.

Cf. here: https://blog.soliditylang.org/2020/03/26/fallback-receive-split/

yomarion: (I won't replicate the comment to all the other fallback implementations in other contracts FYI so if you agree, think about those)

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -83,7 +84,9 @@ contract ERC20SwapToPay is Ownable {
 
     uint256 requestedTotalAmount = _amount + _feeAmount;
 
+    // solhint-disable-next-line
```
yomarion, line 87: Is this mandatory ?

packages/smart-contracts/src/contracts/StorageFeeCollector.sol
```diff
@@ -1,7 +1,7 @@
-pragma solidity ^0.5.0;
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/math/SafeMath.sol";
-import "@openzeppelin/contracts/access/roles/WhitelistAdminRole.sol";
+import "./legacy_openzeppelin/contracts/access/roles/WhitelistAdminRole.sol";
```
yomarion, line 4: Have you considered using [Roles](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/be5ed7364b93daccbb74a09e3f5ec1be6c458097/contracts/access/Roles.sol) ?

packages/smart-contracts/test/contracts/StorageFeeCollector.test.ts
```diff
@@ -167,9 +167,7 @@ describe('contract: StorageFeeCollector', () => {
       );
       const contentSize = BigNumber.from(1000);
 
-      await expect(storageFeeCollector.getFeesAmount(contentSize)).to.be.revertedWith(
-        'SafeMath: multiplication overflow',
-      );
+      await expect(storageFeeCollector.getFeesAmount(contentSize)).to.be.reverted;
```
yomarion, line 170: From my experience, this is risky, I found a few tests failing for the wrong reason previously. Can you instead replace with the expected error message ?

yomarion (changes requested):
Well done, that's a lot of work for such a tiny figure change 😅 Only one comment is blocking, the others are more discussions to be sure.

packages/smart-contracts/src/contracts/StorageFeeCollector.sol
vrolland: Yes, It was my previous PR. I dashback on it change the ABI's.

packages/smart-contracts/test/contracts/StorageFeeCollector.test.ts
vrolland: The new safeMath mechanism does not give any message.

packages/smart-contracts/src/contracts/ERC20FeeProxy.sol
vrolland: Good point! Thanks.

packages/smart-contracts/test/contracts/StorageFeeCollector.test.ts
yomarion: Maybe since safeMath is a default we can remove these low level tests ?

yomarion (approved):
Wonderful

PR #570, opened 2021-08-23
fix: handle NEAR-near legacy naming

yomarion:
## Description of the changes
- Tests NEAR token & currency methods
- Handle `NEAR-near`


vrolland (approved):

benjlevesque (approved):
nice

PR #572, opened 2021-08-24
feat: support new rnf_generic invoice type for other payments

AdamDowson:
## Description of the changes
Allow a new type `rnf_generic` to handle https://app.asana.com/0/1200745919651000/1200620401647325

### DIFF
--- a/packages/data-format/src/format/rnf_invoice/types.ts
+++ b/packages/data-format/src/format/rnf_invoice/types.ts
@@ -1,6 +1,6 @@
 export interface Invoice {
   meta: {
-    format: 'rnf_invoice';
+    format: 'rnf_invoice' | 'rnf_generic';
     version: string;
   };
   buyerInfo?: ActorInfo;
--- a/packages/data-format/src/format/index.ts
+++ b/packages/data-format/src/format/index.ts
@@ -1,6 +1,7 @@
 import * as rnf_invoice_0_0_1 from './rnf_invoice/rnf_invoice-0.0.1.json';
 import * as rnf_invoice_0_0_2 from './rnf_invoice/rnf_invoice-0.0.2.json';
 import * as rnf_invoice_0_0_3 from './rnf_invoice/rnf_invoice-0.0.3.json';
+import * as rnf_generic_0_0_1 from './rnf_invoice/rnf_generic-0.0.1.json';
 
 // Re-export the JSON files structured by format and version.
 // NB: A dynamic require (require(`${format}/${format}-${version}.json`)) would prevent tree-shaking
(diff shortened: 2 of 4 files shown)


packages/data-format/src/format/rnf_invoice/types.ts
```diff
@@ -1,6 +1,6 @@
 export interface Invoice {
   meta: {
-    format: 'rnf_invoice';
+    format: 'rnf_invoice' | 'rnf_generic';
```
yomarion, line 3: I would like more explanation about this, please. It looks strange for an interface `Invoice` to support both formats.

yomarion (commented):
Except one detail, good job

packages/data-format/src/format/rnf_invoice/types.ts
AdamDowson: The rnf_generic type is identical to the rnf_invoice type so we can reuse the same interface vs creating a new one.

yomarion: Completely identical? Why do we have a new format if it's the same?

AdamDowson: It's much easier to handle things in the frontend, everything in API-sync like emails + same with portal. Long term it's much cleaner this way.

yomarion: Easier for you but the library should be clear for everyone using / implementing the protocol.

What about extending a common interface (like `InvoiceData` ?) to `Invoice` and `Generic` (or, to be more precise `InvoiceBasedGeneric`) so it's clear that the only difference is their name and format ? The front-end could then rely on the common interface or on `Invoice | Generic` and it would not be more difficult.

benjlevesque: can't we have `rnf_generic` without a change to the protocol? 

This is an experiment, not something we want very rigorously in the protocol I feel. I even think it's fine skipping the validation for this one if it helps use move forwared

AdamDowson: > can't we have `rnf_generic` without a change to the protocol?
> 
> This is an experiment, not something we want very rigorously in the protocol I feel. I even think it's fine skipping the validation for this one if it helps use move forwared

We can bypass the validation if you think that's best?

vrolland (commented):
Should we close this PR ?

AdamDowson:
https://github.com/RequestNetwork/requestNetwork/pull/572#pullrequestreview-759484425 - yes done @vrolland

PR #576, opened 2021-09-03
feat: Near detection with autobahn

yomarion:
## Description of the changes
- Replaced `pg` dependency with `autobahn` for front-end compatilbility
- Implemented `NearNativeTokenPaymentDetector`
- Created abstraction for address-based and reference-based payment detection for a better factorization


packages/payment-detection/src/near-info-retriever.ts
```diff
@@ -1,33 +1,36 @@
...
    * @param toAddress Address to check
    * @param eventName Indicate if it is an address for payment or refund
    * @param network The id of network we want to check
-   * @param paymentReference The reference to identify the payment
    */
   constructor(
+    private paymentReference: string,
     private toAddress: string,
+    private proxyContractName: string,
+    private procedureName: string,
```
yomarion, line 21: The computation of these 2 parameters is moved to the `detector` class, to keep the same low-level behavior as existing retrievers.

packages/payment-detection/src/reference-based-detector.ts
```diff
@@ -0,0 +1,224 @@
...
+import getBalanceErrorObject from './balance-error';
+import PaymentReferenceCalculator from './payment-reference-calculator';
+
+import { BigNumber } from 'ethers';
+import { IReferenceBased } from 'types/src/extensions/pn-any-reference-based-types';
+
+/**
+ * Abstract class to extend to get the payment balance of reference based requests
+ */
+export default abstract class ReferenceBasedDetector<TPaymentEventParameters>
```
yomarion, line 12: This abstract class was created as the refactoring for code re-usability, based on input-data ETH payment "detector" (the classes that did not have specif names until now, and which are not the retriever). The logic is kept untouched compared to ETH, except for a few factorizations.

packages/payment-detection/src/reference-based-detector.ts
```diff
@@ -0,0 +1,224 @@
...
+        request.requestId,
+        paymentNetwork,
+      );
+      const refunds = await this.extractBalanceAndEvents(
+        paymentNetwork.values.refundAddress,
+        PaymentTypes.EVENTS_NAMES.REFUND,
+        request.currency.network,
+        request.requestId,
+        paymentNetwork,
+      );
```
yomarion, line 116: This has been factorized a bit but the logic is untouched (based on ETH payment detection)

packages/types/src/extensions/pn-any-address-based-types.ts
```diff
@@ -1,17 +1,18 @@
...
   createAddRefundAddressAction: (
     creationParameters: IAddRefundAddressParameters,
   ) => Extension.IAction;
   createCreationAction: (
-    creationParameters: ICreationParameters,
-  ) => Extension.IAction<ICreationParameters>;
-  isValidAddress: (address: string) => boolean;
+    creationParameters: TCreationParameters,
+  ) => Extension.IAction<TCreationParameters>;
+  supportedNetworks: string[];
```
yomarion, line 15: `isValidAddress()` has always been protected, it was a mistake to declare it in this interface (and not needed now with inheritance)

benjlevesque (approved):
lgtm!

PR #583, opened 2021-09-21
feat: ethereum fee proxy smart contract

vrolland:


packages/smart-contracts/src/contracts/EthereumFeeProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+  * @notice Performs an Ethereum transfer with a reference
+  * @param _to Transfer recipient
+  * @param _paymentReference Reference of the payment related
+  * @param _feeAmount The amount of the payment fee (part of the msg.value)
+  * @param _feeAddress The fee recipient
+  */
+  function transferFromWithReferenceAndFee(
```
benjlevesque, line 31: "from" is for an ERC20 `transferFrom` call, here we don't do that.

```suggestion
  function transferWithReferenceAndFee(
```

benjlevesque (approved):

alexandre-abrioux (approved):

PR #587, opened 2021-09-22
feat: payment processing for ethereum fee proxy

vrolland:


benjlevesque (approved):

benjlevesque (changes requested):

packages/payment-processor/src/payment/eth-fee-proxy.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export function encodePayEthFeeProxyRequest(
+  request: ClientTypes.IRequestData,
+  signerOrProvider: providers.Web3Provider | Signer = getProvider(),
+  feeAmountOverride?: BigNumberish,
+): string {
+  validateRequest(request, PaymentTypes.PAYMENT_NETWORK_ID.ETH_FEE_PROXY_CONTRACT);
+  const signer = getSigner(signerOrProvider);
+
+  const proxyAddress = ethereumFeeProxyArtifact.getAddress(request.currencyInfo.network!);
```
benjlevesque, line 60: same as above

packages/payment-processor/src/payment/eth-fee-proxy.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Return the EIP-681 format URL with the transaction to pay an ETH
+ * Warning: this EIP isn't widely used, be sure to test compatibility yourself.
+ *
+ * @param request
+ * @param amount optionally, the amount to pay. Defaults to remaining amount of the request.
+ * @param feeAmountOverride optionally, the fee amount to pay. Defaults to the fee amount of the request.
+ */
+export function _getEthFeeProxyPaymentUrl(
```
benjlevesque, line 75: not necessary, never been used (should be removed from other PNs too, it's not properly tested)

packages/payment-processor/src/payment/utils.ts
```diff
@@ -214,6 +216,30 @@ export function validateErc20FeeProxyRequest(
   }
 }
 
+/**
+ * Validates the amount and fee parameters for an Eth Fee Proxy based request.
+ * @param request to validate
+ * @param feeAmountOverride optionally, the custom fee amount
+ * @param paymentNetwork defaults to ETH Fee Proxy contract
+ */
+export function validateEthFeeProxyRequest(
```
benjlevesque, line 225: I feel this method should be in the eth-fee-proxy file

benjlevesque (approved):

packages/payment-processor/src/payment/utils.ts
vrolland: I made it the same way as for erc20. Again a Refacto mission is required here :face_in_clouds:

vrolland: That is a good question. I would be better for this file but it makes the general organization even more messy.

vrolland: I will go for your point of view anyway :)

benjlevesque: you can put in a new file if you prefer. but the utils should not have specificities of PNs

vrolland: I agree, the questions was should I do it now or in a specific PR refactoring everything.

vrolland: I move it to this file as you suggest

PR #590, opened 2021-09-23
fix: don't check ETH solvency for a Gnosis provider

yomarion:
## Description of the changes
* Since Gnosis is now a dedicated provider, we need to fix the check for ETH solvency need

### DIFF
--- a/packages/payment-processor/src/payment/index.ts
+++ b/packages/payment-processor/src/payment/index.ts
@@ -191,9 +191,11 @@ export async function isSolvent(
   }
   const provider = providerOptions.provider;
   const ethBalance = await provider.getBalance(fromAddress);
-  const needsGas = !['Safe Multisig WalletConnect', 'Gnosis Safe Multisig'].includes(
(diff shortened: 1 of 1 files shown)


packages/payment-processor/src/payment/index.ts
```diff
@@ -191,9 +191,11 @@ export async function isSolvent(
   }
   const provider = providerOptions.provider;
   const ethBalance = await provider.getBalance(fromAddress);
-  const needsGas = !['Safe Multisig WalletConnect', 'Gnosis Safe Multisig'].includes(
-    (provider as any)?.provider?.wc?._peerMeta?.name,
-  );
+  const needsGas =
+    !(provider as any)?.provider?.safe?.safeAddress &&
```
vrolland, line 195: Dirty but clear enough :)

vrolland (approved):

packages/payment-processor/src/payment/index.ts
bertux: Yes at least it's fixing the issue

PR #595, opened 2021-09-28
fix: extension version not based on data

benjlevesque:
⚠️  Please review carefully and challenge that the bug is indeed a bug!

## Context
The Payment Processor should use the contract matching the version of the PN.

## Previous issues
A bug fix in Payment Processor was done earlier https://github.com/RequestNetwork/requestNetwork/pull/556

## Description of the bug
This new bug, not related to the previous, is that the extension version is not based on the stored data, but defaults to the `CURRENT_VERSION` of the extension. 

## Description of the changes
- rely on the abstract extension for `id`, `type` and `version`
- throw if there was no `version` at creation
- tests: test each payment network individually, since regression could have important impacts


vrolland (approved):
As discussed IRL: Good fix, Well done!

alexandre-abrioux (approved):

yomarion (approved):
Well spotted, well fixed :)

PR #600, opened 2021-09-30
fix: remove -rinkeby from symbol

kevindavee:
## Description of the changes
having `-network` in the symbol name cause the `id` to have extra network suffix when calling `CurrencyManager`'s `.from()` method.
It was returning `ETH-rinkeby-rinkeby` for the `id`


benjlevesque (changes requested):
This could conflict with ETH on mainnet , the -rinkeby is there on purpose

PR #601, opened 2021-09-30
fix: allow id to be used for .from method

kevindavee:
## Description of the changes
Allow getting currency by its id. Example usage is for `ETH-rinkeby-rinkeby`

### DIFF
--- a/packages/currency/src/currencyManager.ts
+++ b/packages/currency/src/currencyManager.ts
@@ -36,27 +36,39 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
   }
 
   /**
-   * Gets a supported currency from a symbol, symbol-network or address.
+   * Gets a supported currency from a symbol, symbol-network, currency definition id or address.
    *
-   * @param symbolOrAddress e.g. 'DAI', 'FAU', 'FAU-rinkeby' or '0xFab46E002BbF0b4509813474841E0716E6730136'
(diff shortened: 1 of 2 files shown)


packages/currency/src/currencyManager.ts
```diff
@@ -36,27 +36,40 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
   }
 
   /**
-   * Gets a supported currency from a symbol, symbol-network or address.
+   * Gets a supported currency from a symbol, symbol-network, currency definition id or address.
```
vrolland, line 39: Why not having both: from id  & from symbol?

something like:  `from(idOrSymbolOrAddress: string | undefined)` 
Searching first from id then Symbol or address ?

kevindavee: Yup, we still search by the symbol. The parameter name is more generic because I think it's ugly to have 2 `or`.

kevindavee: The id is in the last order because I put the id in the second search but it breaks the test for conflicting currency.

vrolland: Oh sorry! I should have read the code first :sweat_smile:
I let @benjlevesque make the call for the naming change.

alexandre-abrioux (approved):

yomarion (approved):
Well done

PR #608, opened 2021-09-30
chore: update address validation

alexandre-abrioux:
## Description of the changes

1. Replace [wallet-address-validator](https://www.npmjs.com/package/wallet-address-validator) that hasn't been updated in 3 years by an up-to-date fork: [multicoin-address-validator](https://www.npmjs.com/package/multicoin-address-validator). It also has its own types already implemented.
2. Add a new public API to the `CurrencyManager`:
```js
  /**
   * Validates an address for a given currency.
   * Throws if the currency is not given or if the currency is an ISO4217 currency.
   * We allow undefined type for the "currency" argument to make it easier to chain with the
   * CurrencyManager.from() method but if the argument is not defined it will throw an error.
   */
  static validateAddress(address: string, currency: CurrencyInput): boolean
```
3. (Refacto) Use this new method in our code to centralize address validation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "scripts": {
     "build": "lerna run build",
     "clean": "lerna run clean",
-    "build:tsc": "tsc -b packages/advanced-logic packages/data-access packages/data-format packages/epk-decryption packages/epk-signature packages/ethereum-storage packages/integration-test packages/multi-format packages/payment-detection packages/prototype-estimator packages/request-client.js packages/request-logic packages/request-node packages/smart-contracts packages/toolbox packages/transaction-manager packages/types packages/usage-examples packages/utils packages/web3-signature packages/currency",
+    "build:tsc": "tsc -b packages/advanced-logic packages/currency packages/data-access packages/data-format packages/epk-decryption packages/epk-signature packages/ethereum-storage packages/integration-test packages/multi-format packages/payment-detection packages/prototype-estimator packages/request-client.js packages/request-logic packages/request-node packages/smart-contracts packages/toolbox packages/transaction-manager packages/types packages/usage-examples packages/utils packages/web3-signature",
     "lint": "eslint packages/**/src/**/*.ts --quiet",
     "lint-staged": "lint-staged",
     "lerna": "lerna",
--- a/packages/advanced-logic/src/extensions/payment-network/address-based.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/address-based.ts
@@ -1,6 +1,7 @@
+import { CurrencyManager } from '@requestnetwork/currency';
 import { ExtensionTypes, IdentityTypes, RequestLogicTypes } from '@requestnetwork/types';
-import AbstractExtension from '../abstract-extension';
 import Utils from '@requestnetwork/utils';
+import AbstractExtension from '../abstract-extension';
 
 /**
  * Core of the address based payment networks
@@ -15,7 +16,7 @@ export default abstract class AddressBasedPaymentNetwork<
     public extensionId: ExtensionTypes.ID,
     public currentVersion: string,
     public supportedNetworks: string[],
-    public supportedCurrencyType: string,
+    public supportedCurrencyType: RequestLogicTypes.CURRENCY,
   ) {
     super(ExtensionTypes.TYPE.PAYMENT_NETWORK, extensionId, currentVersion);
     this.actions = {
@@ -141,7 +142,30 @@ export default abstract class AddressBasedPaymentNetwork<
     };
   }
 
-  protected abstract isValidAddress(_address: string, _networkName?: string): boolean;
+  protected isValidAddress(address: string, networkName?: string): boolean {
+    const currencyManager: CurrencyManager = CurrencyManager.getDefault();
+    switch (this.supportedCurrencyType) {
+      case RequestLogicTypes.CURRENCY.BTC: {
+        if (networkName) {
+          const currency = currencyManager.from('BTC', networkName);
+          return CurrencyManager.validateAddress(address, currency);
+        }
+        return this.supportedNetworks.some((network) => {
+          const currency = currencyManager.from('BTC', network);
+          return CurrencyManager.validateAddress(address, currency);
+        });
+      }
+      case RequestLogicTypes.CURRENCY.ETH:
+      case RequestLogicTypes.CURRENCY.ERC20: {
+        const currency = currencyManager.from('ETH', 'mainnet');
+        return CurrencyManager.validateAddress(address, currency);
+      }
+      default:
+        throw new Error(
+          `Default implementation of isValidAddress() does not support currency type ${this.supportedCurrencyType}. Please override this method if needed.`,
+        );
+    }
+  }
 
   /**
    * Applies add payment address
--- a/packages/advanced-logic/src/extensions/payment-network/address-based.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/address-based.ts
@@ -142,18 +142,15 @@ export default abstract class AddressBasedPaymentNetwork<
     };
   }
 
-  protected isValidAddress(address: string, networkName?: string): boolean {
+  protected isValidAddressForNetwork(address: string, networkName: string): boolean {
     const currencyManager: CurrencyManager = CurrencyManager.getDefault();
     switch (this.supportedCurrencyType) {
       case RequestLogicTypes.CURRENCY.BTC: {
-        if (networkName) {
-          const currency = currencyManager.from('BTC', networkName);
-          return CurrencyManager.validateAddress(address, currency);
-        }
-        return this.supportedNetworks.some((network) => {
-          const currency = currencyManager.from('BTC', network);
-          return CurrencyManager.validateAddress(address, currency);
-        });
+        const currency = currencyManager.from(
+          networkName === 'testnet' ? 'BTC-testnet' : 'BTC',
+          networkName,
+        );
+        return CurrencyManager.validateAddress(address, currency);
       }
       case RequestLogicTypes.CURRENCY.ETH:
       case RequestLogicTypes.CURRENCY.ERC20: {
@@ -162,9 +159,18 @@ export default abstract class AddressBasedPaymentNetwork<
       }
       default:
         throw new Error(
-          `Default implementation of isValidAddress() does not support currency type ${this.supportedCurrencyType}. Please override this method if needed.`,
+          `Default implementation of isValidAddressForNetwork() does not support currency type ${this.supportedCurrencyType}. Please override this method if needed.`,
         );
     }
+  }
+
+  protected isValidAddress(address: string, networkName?: string): boolean {
+    if (networkName) {
+      return this.isValidAddressForNetwork(address, networkName);
+    }
+    return this.supportedNetworks.some((network) =>
+      this.isValidAddressForNetwork(address, network),
+    );
   }
 
   /**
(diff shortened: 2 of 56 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "build": "lerna run build",
     "clean": "lerna run clean",
-    "build:tsc": "tsc -b packages/advanced-logic packages/data-access packages/data-format packages/epk-decryption packages/epk-signature packages/ethereum-storage packages/integration-test packages/multi-format packages/payment-detection packages/prototype-estimator packages/request-client.js packages/request-logic packages/request-node packages/smart-contracts packages/toolbox packages/transaction-manager packages/types packages/usage-examples packages/utils packages/web3-signature packages/currency",
+    "build:tsc": "tsc -b packages/advanced-logic packages/currency packages/data-access packages/data-format packages/epk-decryption packages/epk-signature packages/ethereum-storage packages/integration-test packages/multi-format packages/payment-detection packages/prototype-estimator packages/request-client.js packages/request-logic packages/request-node packages/smart-contracts packages/toolbox packages/transaction-manager packages/types packages/usage-examples packages/utils packages/web3-signature",
```
alexandre-abrioux, line 19: Order packages by name (not related to the feature)

packages/advanced-logic/src/extensions/payment-network/address-based.ts
```diff
@@ -141,7 +142,36 @@ export default abstract class AddressBasedPaymentNetwork<
     };
   }
 
-  protected abstract isValidAddress(_address: string, _networkName?: string): boolean;
```
alexandre-abrioux, line 144: I added a default implementation of the method to reduce code duplication in the extended classes.

packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
```diff
@@ -9,7 +10,12 @@ export default class AnyToErc20ProxyPaymentNetwork extends Erc20FeeProxyPaymentN
     extensionId: ExtensionTypes.ID = ExtensionTypes.ID.PAYMENT_NETWORK_ANY_TO_ERC20_PROXY,
     currentVersion: string = CURRENT_VERSION,
   ) {
-    super(extensionId, currentVersion, [], RequestLogicTypes.CURRENCY.ERC20);
```
alexandre-abrioux, line 12: Here the supported networks is an empty array. I think it's an error but please correct me if I'm wrong. The default `isValidAddress()` implementation doesn't work without changing this.

packages/advanced-logic/src/extensions/payment-network/native-token.ts
```diff
@@ -38,4 +38,11 @@ export default abstract class NativeTokenPaymentNetwork extends ReferenceBasedPa
     }
     return super.createCreationAction(creationParameters);
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected isValidAddress(_address: string, _networkName?: string): boolean {
+    throw new Error(
+      `Default implementation of isValidAddress() does not support native tokens. Please override this method.`,
```
alexandre-abrioux, line 45: Security to prevent running the default implementation of `isValidAddress()` on native chains which may have other address format.

packages/advanced-logic/src/extensions/payment-network/bitcoin/testnet-address-based.ts
```diff
@@ -13,23 +11,10 @@ const BITCOIN_NETWORK = 'testnet';
  * Every bitcoin transaction that reaches these addresses will be interpreted as payment or refund.
  * Important: the addresses must be exclusive to the request
  */
-export default class BitcoinTestnetAddressBasedPaymentNetwork extends AddressBasedPaymentNetwork {
+export default class BitcoinTestnetAddressBasedPaymentNetwork extends BitcoinAddressBasedPaymentNetwork {
```
alexandre-abrioux, line 14: To make `codeclimate` happy, `BitcoinTestnetAddressBasedPaymentNetwork` now extends `BitcoinAddressBasedPaymentNetwork` :) Which isn't a bad idea IMO

packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
vrolland: Good move !

packages/advanced-logic/src/extensions/payment-network/bitcoin/testnet-address-based.ts
vrolland: looks reasonable indeed !

vrolland (approved):
Nice change, a lot of comments: very good PR :+1: 
Maybe just change the PR title for something more explicit, for example: `chore: update of the address validations`

packages/currency/package.json
```diff
@@ -44,6 +44,8 @@
     "@metamask/contract-metadata": "1.28.0",
     "@requestnetwork/types": "0.35.0",
     "@requestnetwork/utils": "0.35.0",
+    "@types/multicoin-address-validator": "0.5.0",
```
benjlevesque, line 47: maybe a detail but this should be a devDependency

packages/advanced-logic/src/extensions/payment-network/address-based.ts
```diff
@@ -141,7 +142,42 @@ export default abstract class AddressBasedPaymentNetwork<
...
+  protected isValidAddress(address: string, networkName?: string): boolean {
+    if (networkName) {
+      return this.isValidAddressForNetwork(address, networkName);
+    }
+    return this.supportedNetworks.some((network) =>
+      this.isValidAddressForNetwork(address, network),
+    );
```
benjlevesque, line 151: maybe outside the scope of this PR but I feel the network should be mandatory

packages/advanced-logic/src/extensions/payment-network/address-based.ts
```diff
@@ -141,7 +142,42 @@ export default abstract class AddressBasedPaymentNetwork<
...
+  }
+
+  protected isValidAddressForSymbolAndNetwork(
+    address: string,
+    symbol: string,
+    network: string,
+  ): boolean {
+    const currencyManager = CurrencyManager.getDefault();
+    const currency = currencyManager.from(symbol, network);
+    return CurrencyManager.validateAddress(address, currency);
```
benjlevesque, line 179: Could this method be called for a token that is not in the default Currency Manager, or only for ETH/BTC/NEAR.... ?
Using the default currency manager may be a problem here.

packages/currency/src/currency-utils.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Checks if a Near address is valid according to a currency network.
+ * Returns true if the currency network is not given and the address is correct for any network.
+ */
+export const isValidNearAddress = (address: string, network?: string): boolean => {
+  if (!network) {
+    return isValidNearAddress(address, 'aurora') || isValidNearAddress(address, 'aurora-testnet');
+  }
+  if (!address.match(/^(([a-z\d]+[-_])*[a-z\d]+\.)*([a-z\d]+[-_])*[a-z\d]+$/)) {
```
benjlevesque, line 9: this deserves a comment :)

packages/currency/src/currencyManager.ts
```diff
@@ -172,6 +174,43 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
...
+  /**
+   * Validates an address for a given currency.
+   * Throws if the currency is not given or if the currency is an ISO4217 currency.
+   * We allow undefined type for the "currency" argument to make it easier to chain with the
+   * CurrencyManager.from() method but if the argument is not defined it will throw an error.
+   */
+  static validateAddress(address: string, currency: CurrencyInput | undefined): boolean {
+    if (!currency) {
+      throw new Error('Could not validate an address of an undefined currency');
+    }
```
benjlevesque, line 186: maybe just don't accept undefined? I'm not  a huge fan of throwing in these cases, it kind of requires apps that use this helper to build a wrapper around this method

benjlevesque (commented):
nice!

packages/currency/package.json
alexandre-abrioux: oops! done

packages/advanced-logic/src/extensions/payment-network/address-based.ts
alexandre-abrioux: Yes I agree with you. It's a major refacto as there is a lot of usage of this method without the second argument in our code, but it should be doable as we should know the network in advance in all those cases.

packages/advanced-logic/src/extensions/payment-network/address-based.ts
alexandre-abrioux: It cannot as it is designed because it's a protected method of `AddressBasedPaymentNetwork` that is only used by the class and its descendants. It is only called with the `ETH`, `BTC`, and `NEAR` tokens.

I hesitated a lot between placing this method as a public helper in the `CurrencyManager` where it would just chain a `from` and a `validateAddress`
```ts
isValidAddressForSymbolAndNetwork(
    address: string,
    symbol: string,
    network: string,
): boolean {
    const currency = currencyManager.from(symbol, network);
    return this.validateAddress(address, currency);
}
```
; or as a protected method in the `AddressBasedPaymentNetwork`.

I choose to place it in the `AddressBasedPaymentNetwork` as it is only used here for now, in order to prevent adding another public API to `CurrencyManager` that we have to maintain. But this choice is debatable, tell me what you prefer, I can move it easily.

benjlevesque: I think it's OK 👍

alexandre-abrioux: PS. In any case (even if we move it to the `CurrencyManager`), the `AddressBasedPaymentNetwork` would still be using the default currency manager.

packages/currency/src/currency-utils.ts
alexandre-abrioux: This regex was moved from the `near-native.ts` implementation: https://github.com/RequestNetwork/requestNetwork/blob/d26ef8beea87e9ca268689495143222c6b8d034b/packages/advanced-logic/src/extensions/payment-network/near-native.ts#L40
; but I don't know where it comes from exactly. I didn't find much content on it on the web.
Maybe @yomarion can help on this?

packages/currency/src/currencyManager.ts
alexandre-abrioux: done is https://github.com/RequestNetwork/requestNetwork/pull/608/commits/3b350dd8b9c3045c1fd4694f2a4367e7b572dd42 , please tell me if it's OK

benjlevesque: yep!

benjlevesque (approved):

packages/advanced-logic/src/extensions/payment-network/bitcoin/mainnet-address-based.ts
```diff
@@ -13,22 +11,11 @@ const BITCOIN_NETWORK = 'mainnet';
...
-   *
-   * @param address address to check
-   * @returns true if address is valid
-   */
-  protected isValidAddress(address: string): boolean {
-    return walletAddressValidator.validate(address, 'bitcoin', 'prod');
+  public constructor(
+    public extensionId: ExtensionTypes.ID = ExtensionTypes.ID.PAYMENT_NETWORK_BITCOIN_ADDRESS_BASED,
+    public currentVersion: string = CURRENT_VERSION,
+    public supportedNetworks: string[] = [BITCOIN_NETWORK],
```
kevindavee, line 17: Why do we introduce parameters? Are those parameters injectable with other kinds of values?

kevindavee (approved):
LGTM! No additional comments :) Just need to addressed some existing comments

packages/currency/src/currency-utils.ts
alexandre-abrioux: See with @yomarion, added comments with NEAR specification link

PR #611, opened 2021-10-01
feat: yarn unlink:all

alexandre-abrioux:
## Description of the changes

cf. https://github.com/RequestNetwork/requestNetwork/pull/609 but does the opposite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "test": "lerna run test",
     "format": "prettier packages/**/src/**/*.ts -w",
     "format:check": "prettier packages/**/src/**/*.ts -c",
-    "link:all": "for d in packages/*; do pushd $d; yarn link; popd; done"
+    "link:all": "for d in packages/*; do pushd $d; yarn link; popd; done",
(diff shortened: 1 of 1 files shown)


benjlevesque (approved):
dumb question: when do you need to unlink? :)

alexandre-abrioux:
@benjlevesque Not needed normally :) It helped me quickly reset to a clean state for debugging.

I ran by mistake `yarn link @requestnetwork/currency` in this repo root folder. So naturally I ran `yarn unlink @requestnetwork/currency` thinking it would undo my mistake. But this broke `lerna` linking as I ended up with no `@requestnetwork/currency` symlink in the root `node_modules` anymore. I ran into compiling errors and I didn't understand what happened until I decided to reset everything. The solution for me was to remove the root `node_modules` and run `yarn` again to bootstrap `lerna` so that it could recreate all symlinks (running `yarn` or `yarn lerna bootstrap` without removing the `node_modules` folder in the first place does not recreate symlinks). So this new command didn't solve anything but it helped me debug.

PR #613, opened 2021-10-01
fix: getDataIdMeta heavy load

benjlevesque:
## Description of the changes
Currently, everytime we try to get metadata for a hash that wasn't confirmed, we get all events since initial blocks. This is a massive waste of resources because this data will anyway be updated when the transaction is confirmed. 

For nodes that did not initiate the request, the data will be available once the block is mined, as expected.

This PR proposes to no longer check onchain for ethereum metadata, but rather wait for the metadata to be added to the cache. Until then, the request is considered PENDING. 

TODO: the metadata is different on the node that created the requests than another passive node


vrolland (approved):
The root cause is the lack of indexation of the hash in `requestHashStorage`:
`event NewHash(**string hash**, address hashSubmitter, bytes feesParameters);`
If it was indexed we could have filter the events by hashes, and so, avoid so many calls.
We might want to update the `requestHashStorage` which is not a light change.

But I feel this change is not an issue for the so called "passive request nodes", as the metadata will be updated during a future synchronization of new DataIds.
It requires some tests to be sure there is no side effect, though.

Nice discover @benjlevesque :+1:

alexandre-abrioux (approved):
Nice findings!

packages/ethereum-storage/src/smart-contract-manager.ts
```diff
@@ -401,25 +401,6 @@ export default class SmartContractManager {
     });
   }
 
-  /**
-   * Get Ethereum metadata from the content hash
-   * @param contentHash Hash of the content to store, this hash should be used to retrieve the content
-   * @returns Promise resolved when transaction is confirmed on Ethereum
-   */
-  public async getMetaFromEthereum(contentHash: string): Promise<StorageTypes.IEthereumMetadata> {
```
benjlevesque, line 409: this method was the culprit of bad perfs. it became unnecessary

packages/ethereum-storage/test/ethereum-metadata-cache.test.ts
```diff
@@ -0,0 +1,63 @@
+import { StorageTypes } from '@requestnetwork/types';
```
benjlevesque, line 1: file renamed ! sorry 😬 

the main change is that I dropped 2 tests that were either useless with these changes, or redundant

packages/ethereum-storage/test/ethereum-metadata-cache.ts
```diff
@@ -1,130 +0,0 @@
...
-    );
-    await expect(ethereumMetadataCache.metadataCache.get(hashExample2)).resolves.toEqual(
-      ethereumMetadataExample2,
-    );
-    await expect(ethereumMetadataCache.metadataCache.get(hashExample3)).resolves.toEqual(
-      ethereumMetadataExample3,
-    );
-  });
-
-  it('allows to retrieve saved metadata', async () => {
```
benjlevesque, line 79: redundant with the previous one

packages/ethereum-storage/test/ethereum-metadata-cache.ts
```diff
@@ -1,130 +0,0 @@
...
-    const readReturn3 = await ethereumMetadataCache.getDataIdMeta(hashExample3);
-
-    expect(readReturn1).toEqual(ethereumMetadataExample1);
-    expect(readReturn2).toEqual(ethereumMetadataExample2);
-    expect(readReturn3).toEqual(ethereumMetadataExample3);
-
-    expect(spy).not.toHaveBeenCalled();
-  });
-
-  it('allows to save when trying to read new metadata', async () => {
```
benjlevesque, line 97: no longer the case : the metadata cache is dumb

packages/ethereum-storage/test/ethereum-storage.test.ts
```diff
@@ -373,15 +371,6 @@ describe('EthereumStorage', () => {
       );
     });
 
-    it('cannot read if ethereumMetadataCache.getDataIdMeta fail', async () => {
```
benjlevesque, line 376: test dropped, `getDataIdMeta` cannot fail

packages/integration-test/test/layers.test.ts
```diff
@@ -424,7 +424,8 @@ describe('Request system', () => {
     const request1CancelHash: RequestLogicTypes.ICancelParameters = {
       requestId: requestId1,
     };
-    await requestLogic.cancelRequest(request1CancelHash, payeeIdentity);
+    const cancel = await requestLogic.cancelRequest(request1CancelHash, payeeIdentity);
+    await new Promise((resolve) => cancel.on('confirmed', resolve));
```
benjlevesque, line 428: about this and all await on "confirmed": i'm almost sure this is because we no longer do a long query to get the metadata, so the request is no longer confirmed straight away

packages/ethereum-storage/test/smartcontract-manager.test.ts
```diff
@@ -427,29 +427,6 @@ describe('SmartContractManager', () => {
     expect(SmartContracts.requestHashSubmitterArtifact.getCreationBlockNumber('private')).toBe(1);
   });
 
-  it('allows to getMetaFromEthereum() a hash', async () => {
```
benjlevesque, line 430: getMetaFromEthereum was dropped

packages/ethereum-storage/test/ethereum-storage.test.ts
alexandre-abrioux: Actually it could fail if the provided store for `Keyv` is not an in-memory store, but a Redis store or something else. But I guess we could let any Keyv exception errors bubble up the stack and kill the NodeJS process.

Keyv storage should be able to reconnect itself to its backend (in case of connection issues with the backend) and it should be transparent for us. If an exception happens, then it is critical, and there is nothing that can be done on our side, I think.

packages/integration-test/test/layers.test.ts
alexandre-abrioux: Nice! We should replace all those too I think, to stabilize our tests. I'll open a PR later on for this.
```ts
    // wait a bit
    // eslint-disable-next-line no-magic-numbers
    await new Promise((r: any): any => setTimeout(r, 1000));
```

alexandre-abrioux (approved):

packages/integration-test/test/layers.test.ts
benjlevesque: I did a bunch already, but there's a lot to do. We should maybe introduce a util for that.

alexandre-abrioux: Yes I saw that, very nice! I agree for the util.

PR #592, opened 2021-09-24
specs: declarePayment action on main payment networks

yomarion:
## Description of the changes

Specify that the payee can `declarePayment` for main payment networks:
- ERC20FeeProxy
- AnyToERC20Proxy
- ETHInputData
- ETHFeeProxy

❓ Cross-check: We can keep the same payment network versions for each of them as only the payee can trigger this action.

PR #620, opened 2021-10-07
feat: address based pn inherits from declarative

kevindavee:
## Description of the changes
Address-based payment network can simply inherit declarative payment network. It unlocks all of the declare payment and refund actions right away.

Specs: #592


kevindavee:
Yeah, prefer to change it in another PR @vrolland

PR #621, opened 2021-10-07
feat: !WIP! simple batching POC 

vrolland:

### DIFF
--- a/packages/data-access/src/data-access.ts
+++ b/packages/data-access/src/data-access.ts
@@ -10,6 +10,8 @@ import IgnoredLocationIndex from './ignored-location';
 import IntervalTimer from './interval-timer';
 import TransactionIndex from './transaction-index';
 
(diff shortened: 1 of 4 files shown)


packages/data-access/src/data-access.ts
```diff
@@ -79,6 +81,8 @@ export default class DataAccess implements DataAccessTypes.IDataAccess {
    */
   private logger: LogTypes.ILogger;
 
+  private buffer: Buffer;
+
```
yomarion, line 85: (TODO for later) The consturctor should have additional arguments:
```suggestion
// Existing
synchronizationIntervalTime: DEFAULT_INTERVAL_TIME,
// New, optional, could be 0 if there is no max batch size
maxBatchSize: int,
// New, mandatory to give a period for batching
batchingIntervalTime: int,
```

yomarion (commented):
TODO before merging: the buffer should be on disk and not on memory.

PR #656, opened 2021-10-27
feat: Eth native payments detection from TheGraph

KolevDarko:
## Description of the changes
We want to detect payments from TheGraph instead of using web3 for all networks except Mainnet.
We don't have subgraph for mainnet yet.


alexandre-abrioux (approved):

benjlevesque (approved):
all good except it should be for all chains but mainnet (not only on rinkeby)

benjlevesque:
I actually have a doubt about this @KolevDarko , I think we might need to update the subgraph actually

packages/payment-detection/test/eth/fee-proxy-detector.test.ts
```diff
@@ -1,5 +1,9 @@
 import { AdvancedLogicTypes, PaymentTypes, RequestLogicTypes } from '@requestnetwork/types';
 import { EthFeeProxyPaymentDetector } from '../../src/eth/fee-proxy-detector';
+import TheGraphInfoRetriever from '../../src/erc20/thegraph-info-retriever';
+import { EVENTS_NAMES } from '@requestnetwork/types/dist/payment-types';
```
benjlevesque, line 4:
```suggestion
import { PaymentTypes } from '@requestnetwork/types';
```

and use `PaymentTypes.EVENT_NAMES`

packages/payment-detection/test/eth/fee-proxy-detector.test.ts
```diff
@@ -145,4 +152,71 @@ describe('api/eth/fee-proxy-contract', () => {
...
+    const transferEvents = await graphRetriever.getTransferEvents();
+    expect(transferEvents).toHaveLength(1);
+    expect(transferEvents[0].amount).toEqual("30000000000000");
+    expect(transferEvents[0].name).toEqual("payment");
+    expect(transferEvents[0].parameters?.to).toEqual(paymentData.to);
+    expect(transferEvents[0].parameters?.txHash).toEqual(paymentData.txHash);
+    expect(transferEvents[0].parameters?.block).toEqual(paymentData.block);
+  })
+
+  it('should get payment event from ethFeeConversionProxy via subgraph', async () => {
```
benjlevesque, line 189: is this the right file for this test?

benjlevesque (approved):

packages/payment-detection/test/eth/fee-proxy-detector.test.ts
KolevDarko: Omg I'm importing from dist even 🤦 . Fixed in latest commit

packages/payment-detection/src/eth/fee-proxy-detector.ts
```diff
@@ -61,14 +62,24 @@ export class EthFeeProxyPaymentDetector extends FeeReferenceBasedDetector<Paymen
...
 
-    const proxyInfoRetriever = new EthProxyInfoRetriever(
-      paymentReference,
-      proxyContractArtifact.address,
-      proxyContractArtifact.creationBlockNumber,
-      address,
-      eventName,
-      network,
-    );
+    const proxyInfoRetriever =
```
benjlevesque, line 65: shouldn't there also be a similar thing for any-to-eth?

packages/payment-detection/test/eth/fee-proxy-detector.test.ts
KolevDarko: You're right.
I put the tests in a new separate file specifically for thegraph info retriever called thegraph-info-retriever.test.ts

packages/payment-detection/src/eth/fee-proxy-detector.ts
KolevDarko: In all places where there could be a distinction between theGraph or web3 info retrievers we are using this [networkSupportsTheGraph](https://github.com/RequestNetwork/requestNetwork/blob/2a884beb1447b7edc02328670275dac1ca1a8563/packages/payment-detection/src/thegraph/index.ts#L33) function.
But in [any-to-eth](https://github.com/RequestNetwork/requestNetwork/blob/eth-payments-thegraph/packages/payment-detection/src/any/any-to-eth-proxy.ts) that same function is not used, so I'm confused how it should be handled there. 
Perhaps @yomarion could clarify this stuff for us.

benjlevesque (approved):
one comment pending from previous review, otherwise seems fine

KolevDarko:
The last sticking point here is this error in integration tests where ETH Fee proxy test is failing because it's trying to access a local subgraph:
console.error
    Eth fee proxy data error:  {
      code: 0,
      message: 'request to http://localhost:8000/subgraphs/name/requestnetwork/request-payments-private failed, reason: connect ECONNREFUSED 127.0.0.1:8000'
    }
I'm not sure the solution is to start a local subgraph because the Erc20 tests which use theGraph already do not require it.

PR #667, opened 2021-11-03
chore: ignore ipfs-manager.test.ts

kevindavee:
## Description of the changes
`ipfs-manager.test.ts` has been so flaky nowadays. Ignoring this first as it's blocking our deployments.
Asana ticket has been created to be investigated: https://app.asana.com/0/1200939559032347/1201300801042818/f


benjlevesque (changes requested):
I'd rather we don't ignore tests, espcially in core layers.
Maybe we can use `retry`  instead, or actually fix the test?

btw, if we ignore a test, better to do it in the test itself (with .skip), better for understanding of what's going on

PR #670, opened 2021-11-05
fix(request-client): make creation refresh synchronous

alexandre-abrioux:
## Issue

During creation of a request we do an `await request.refresh()` right before returning the request object. This creates a call to the request node that queries the pending request's data.

However the request is an EventEmitter, so `error` or `confirmed` events can be emitted while the pending request's data is being fetched. This causes consuming code to miss those events.

## Description of the changes

Two solutions:

1. Either we remove the `await request.refresh()` all together. It is not a big deal knowing that another refresh is done after the confirmation happens anyway here:
https://github.com/RequestNetwork/requestNetwork/blob/7efaec7add64e5723f50652851e3b4fb241103a2/packages/request-client.js/src/api/request.ts#L106-L113
However it means we would remove having a "pending" state in the request.

2. Or we could refresh the request by injecting the data that we already have = the data that was sent to the request node  = without querying the request node =  in a synchronous way.

I went for the second option here.


MantisClone:
@alexandre-abrioux Do you think this PR is still relevant?

alexandre-abrioux:
@MantisClone The topic is still relevant, but the PR is outdated. Closing for now.

PR #673, opened 2021-11-11
Feat/add fantom tokens

KolevDarko:
## Description of the changes
I added the chainlink fantom aggregators for the tokens [described here](https://app.asana.com/0/1190298648548641/1201172909956370/f) and updated the fantom.ts list of supported tokens

### DIFF
--- a/packages/currency/src/erc20/networks/fantom.ts
+++ b/packages/currency/src/erc20/networks/fantom.ts
@@ -6,4 +6,54 @@ export const supportedFantomTokens: TokenMap = {
     symbol: 'YEL',
     decimals: 18,
   },
+  '0x6a07A792ab2965C72a5B8088d3a069A7aC3a993B': {
+    name: 'AAVE Token',
+    symbol: 'AAVE',
+    decimals: 18,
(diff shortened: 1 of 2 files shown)


packages/currency/src/erc20/networks/fantom.ts
```diff
@@ -6,4 +6,54 @@ export const supportedFantomTokens: TokenMap = {
     symbol: 'YEL',
     decimals: 18,
   },
+  '0x6a07A792ab2965C72a5B8088d3a069A7aC3a993B': {
```
benjlevesque, line 9: I'm not sure it's necessary to add these tokens here? the currency API should be enough

benjlevesque (approved):

packages/currency/src/erc20/networks/fantom.ts
KolevDarko: Oh man, I was having issues because I added them in both places and it was breaking because of duplicates.
I'll add them to the currency manager only then

bertux (approved):

PR #676, opened 2021-11-15
refactor: payment detectors inheritance

benjlevesque:
- naming: `DeclarativePaymentDetector` instead of various names
- harmonize usage of  `PaymentTypes.PAYMENT_NETWORK_ID`.
- enforce passing `PaymentTypes.PAYMENT_NETWORK_ID` and the extension as constructor params to avoid forgetting overriding this property in a child class. **This is the main motivation of this PR**. For this to work properly, I sometimes introduced a `-Base` class to keep the current class constructor signature and have a better extensibility

### DIFF
--- a/packages/payment-detection/src/any-to-any-detector.ts
+++ b/packages/payment-detection/src/any-to-any-detector.ts
@@ -1,9 +1,4 @@
-import {
-  AdvancedLogicTypes,
-  ExtensionTypes,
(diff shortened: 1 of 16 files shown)


alexandre-abrioux (approved):

kevindavee (approved):
Super clean!

PR #679, opened 2021-11-19
chore: remove usage of node-fetch

benjlevesque:


kevindavee (approved):

alexandre-abrioux (approved):
@benjlevesque A suggestion: use the same `axios` version in all of our packages, so use `0.21.1` in `ethereum-storage`, or upgrade others to `0.24.0`. Apart from that all good! Well done 👍

PR #680, opened 2021-11-19
feat: adding fantom chainlink aggregators to advanced-logic > conversion-supported-currencies

KolevDarko:
Updated the chainlink aggregators for fantom previously but only in currency package, now adding them also in advanced-logic package.
Also updated the aggregators hashes in currency package to be all lowercase, I noticed the other ones are lowercase.

Update: Implemented the suggested change requests in the following way.
Now advanced-logic package gets the chainlink oracles mapping from currency package and then creates it's own mapping that it uses for any-to-eth and any-to-erc20. 

It uses the same logic from the toolbox package, except that toolbox gets it from the contract directly, but advanced-logic here uses the mapping from currency package as mentioned in previous paragraph.


KolevDarko:
Actually, I need to add fantom tests as well to this

benjlevesque (approved):

AdamDowson (approved):

packages/currency/src/chainlink-path-aggregators.ts
```diff
@@ -272,3 +273,71 @@ export function getPath(
   // Get the path
   return route.path(currencyFrom.hash.toLowerCase(), currencyTo.hash.toLowerCase());
 }
+
+/**
+ * These currencies are supported by Chainlink for conversion.
+ */
+export const supportedCurrencies: Record<string, Record<RequestLogicTypes.CURRENCY, string[]>> = {
```
yomarion, line 280:
```suggestion
export const currenciesWithConversionOracles: Record<string, Record<RequestLogicTypes.CURRENCY, string[]>> = {
```

yomarion (commented):
Are you sure it's a good idea to migrate the `supportedCurrencies` map to the `currency` package ? I have a big doubt, it looks related to one payment network only, so very specific to the advanced logic or payment processor.

benjlevesque (approved):

packages/currency/src/chainlink-path-aggregators.ts
KolevDarko: I see what you mean. In the meantime I managed to generate the supportedCurrencies map out of the aggregators map in currency package, using the logic from toolbox.
So I will leave the mapping in advanced-logic but generated it out of the aggregator list in /currency package.
The whole idea behind this small refactor was to stop keeping duplicates and have one source of truth.

packages/advanced-logic/src/extensions/payment-network/conversion-supported-currencies.ts
```diff
@@ -1,56 +1,69 @@
...
+    record[type].push(address);
+  }
+  record[type] = record[type].sort();
 };
+
+const generateCurrenciesWithOraclesFromAggregators = (allChainlinkPairs: Record<string, Record<string, Record<string, number>>>):
+  Record<string, Record<RequestLogicTypes.CURRENCY, string[]>> => {
+  const supportedCurrencies: Record<string, Record<RequestLogicTypes.CURRENCY, string[]>> = {};
+  const supportedNetworks = Object.keys(chainlinkCurrencyPairs);
+  // const supportedNetworks = ['private', 'rinkeby', 'mainnet', 'matic', 'fantom'];
```
yomarion, line 48: Commented code here FYI

packages/toolbox/src/chainlinkConversionPathTools.ts
```diff
@@ -154,7 +154,7 @@ const addSupportedCurrency = (
   record[type] = record[type].sort();
 };
 
-export const listAggregators = async (options?: IOptions): Promise<void> => {
+export const listAggregators = async (options?: IOptions): Promise<any> => {
```
yomarion, line 157: Ideally, can you type the return promise ?

yomarion (approved):

packages/toolbox/src/chainlinkConversionPathTools.ts
KolevDarko: I didn't even end up using this, so I'm returning it to void

PR #683, opened 2021-11-22
chore: enable typescript importHelpers

alexandre-abrioux:
## Description of the changes

This is a proposition (open discussion) to globally enable typescript's compiler [`importHelpers` option](https://www.typescriptlang.org/tsconfig#importHelpers) in order to reduce duplicated code in our library.

For instance a lot of our compiled files contain the `_awaiter` helper. This change will remove the creation of such helpers in each and every of our compiled files, and `require()` the centralized helpers of `tslib` instead when needed.

Only downside is that some packages that do not use (yet) any helper like `advanced-logic` now have an unused dependency. We could avoid adding this new dependency for those packages but, as the option is globally enabled, and for future proofing, the dependency has been added everywhere, ie. in case helpers are needed later on, to prevent forgetting adding it.

The only exception is the `types` package. The option here is specifically disabled as we will probably never need any helper in this package.


benjlevesque (approved):
sounds like a good idea!

bertux (approved):

PR #690, opened 2021-11-30
feat: standard deployments on celo

yomarion:

### DIFF
--- a/packages/smart-contracts/src/lib/artifacts/ERC20FeeProxy/index.ts
+++ b/packages/smart-contracts/src/lib/artifacts/ERC20FeeProxy/index.ts
@@ -102,6 +102,15 @@ export const erc20FeeProxyArtifact = new ContractArtifact<ERC20FeeProxy>(
         },
       },
     },
+    '0.2.0-next': {
+      abi: ABI_0_1_0,
+      deployment: {
+        celo: {
+          address: '0x0DfbEe143b42B41eFC5A6F87bFD1fFC78c2f0aC9',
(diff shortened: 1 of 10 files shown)


packages/smart-contracts/src/lib/artifacts/ERC20FeeProxy/index.ts
```diff
@@ -102,6 +102,15 @@ export const erc20FeeProxyArtifact = new ContractArtifact<ERC20FeeProxy>(
         },
       },
     },
+    '0.2.0-next': {
```
benjlevesque, line 105: what is this?

benjlevesque (approved):

packages/smart-contracts/src/lib/artifacts/ERC20FeeProxy/index.ts
yomarion: I did not want to change the version just for that but I also wanted to document the new address. Do you think it's a bad idea?

benjlevesque: hum maybe just add a comment...

yomarion: You mean instead of the entry or to explain the entry?

benjlevesque: both :D

PR #702, opened 2021-12-13
fix: skip validation for declarative actions

kevindavee:
## Description of the changes
When using `any-to-erc20` or `any-to-eth`, we will not be able to declare invoices if the network doesn't have chainlink oracle. We need to check action parameters only for creation action.


benjlevesque (commented):
I guess that works, but wouldn't it be better to change the name of the `network` field that we add for declarative payments? 

Imo, the original `network` is something very strict, whereas the `network` added for payment declaration is purely informative

yomarion (commented):
Can't we actually declare this array once only in `ReferenceBasedPaymentNetwork from './reference-based'`? If I'm not wrong, both networks inherit from it.

benjlevesque:
> Hmm. But how about the previous data and data that has been stored in IPFS? Are we able to alter it especially for the decentralized one?

sorry @kevindavee I missed this comment
I think this is a relatively new feature and it's OK to not "fix" those...

kevindavee:
Hmm, I see. Do you think it's okay @yomarion ? Any opinion about this?

@benjlevesque so if we go this way, then we need to migrate the existing data which already uses `network` to the new field name, right?

yomarion (approved):
Nice, well done

KolevDarko (approved):

PR #704, opened 2021-12-14
feat: add userId in data-format

kevindavee:
## Description of the changes
Add `userId` field for data-format validation


yomarion (changes requested):
I think that's wrong, this userId does not belong to the protocol layer.

kevindavee:
This PR is no longer relevant. I'll close

PR #706, opened 2021-12-14
chore: Dockerfile.dev

benjlevesque:
It can be very useful to be able to rapidly build the Request Node with the latest build from all packages. 

I found this to be the best config, as it takes a few seconds to build the whole monorepo (typescript only) and the request-node with [NCC](https://github.com/vercel/ncc)

### DIFF
--- a/packages/request-node/Dockerfile.dev
+++ b/packages/request-node/Dockerfile.dev
@@ -1,0 +1,16 @@
+FROM node:14-alpine
+
+## Warning! This Docker config is meant to be used for development and debugging, not in prod.
+## This is a simplified Dockerfile for a fast build
+## It expects that the monorepo is compiled (`yarn build:tsc` at the monorepo root)
+## And that the request-node is built with NCC: `npx @vercel/ncc build src/server.ts -o dist/ncc`
+
+WORKDIR /request-node
+COPY dist/ncc .
+
+# Port configuration
+ENV PORT 3000
+EXPOSE 3000
+
+# Run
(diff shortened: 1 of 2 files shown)


kevindavee (approved):

Dadogg80 (approved):

yomarion (approved):

alexandre-abrioux (approved):
Great 👍

Suggestion: we could use the `Dockerfile` at the root of the repository and add a new layer just for the request-node. This would remove any dependency on externally compiled files, reducing risks of unexpected error occurring due to Node.js version that may differ on host. It could even be renamed `Dockerfile.dev` as it looks like it is not used anymore since https://github.com/RequestNetwork/requestNetwork/commit/dad9e7d88544c9d226e8dd62454aba59e0e2cfc3

benjlevesque:
> Great 👍
> 
> Suggestion: we could use the `Dockerfile` at the root of the repository and add a new layer just for the request-node. This would remove any dependency on externally compiled files, reducing risks of unexpected error occurring due to Node.js version that may differ on host. It could even be renamed `Dockerfile.dev` as it looks like it is not used anymore since [dad9e7d](https://github.com/RequestNetwork/requestNetwork/commit/dad9e7d88544c9d226e8dd62454aba59e0e2cfc3)

Sure! I went for the simplest option, and doing that might be actually quite a bit longer, since it needs installing packages + does a full build instead of just TSC. 

I'll keep the simple version for now, just did this PR to not lose that config, but of course, can be improved later...

PR #707, opened 2021-12-14
refactor: make node extensible, extract ipfs storage logic

benjlevesque:
This is in preparation for a Node with TheGraph as it backend.

- enabled RequestNode to be inherited from, and changing the DataAccess
- moved IPFS logic out of EthereumStorage
- make the start/stop process of DataAccess more abstract


packages/data-access/src/data-access.ts
```diff
@@ -440,7 +466,9 @@ export default class DataAccess implements DataAccessTypes.IDataAccess {
    * Stop to synchronize with the storage automatically
    */
   public stopAutoSynchronization(): void {
-    this.synchronizationTimer.stop();
+    if (this.synchronizationTimer.isStarted) {
```
benjlevesque, line 469: not related but I noticed this crashed sometimes

packages/ethereum-storage/src/ethereum-storage.ts
```diff
@@ -2,12 +2,11 @@ import { LogTypes, StorageTypes } from '@requestnetwork/types';
 import Utils from '@requestnetwork/utils';
```
benjlevesque, line 2: all IPFS logic in this file has been moved to ipfs-storage

packages/ethereum-storage/test/ethereum-storage.test.ts
```diff
@@ -528,14 +528,6 @@ describe('EthereumStorage', () => {
       await expect(ethereumStorage.read('')).rejects.toThrowError('No id provided');
     });
 
-    it('append and read on an invalid ipfs gateway should throw an error', async () => {
```
benjlevesque, line 531: I don't understand why we have such a method, I removed it, it's only used in this test

packages/types/src/data-access-types.ts
```diff
@@ -3,6 +3,7 @@ import { EventEmitter } from 'events';
 /** Data Access Layer */
 export interface IDataAccess {
   initialize: () => Promise<void>;
+  close: () => Promise<void>;
```
benjlevesque, line 6: making the closing generic allows to avoid specific actions in the request node

packages/ethereum-storage/src/ethereum-storage.ts
yomarion: Clever, thanks

yomarion (approved):
Super clean, well done ! Note for later: find a better name for `EthereumStorage`, for example `EthereumStorageOracle` or `EthereumStorageIndexer`.

alexandre-abrioux (approved):
Nice decoupling!

PR #719, opened 2021-12-21
refactor: proper typing for storageMeta

benjlevesque:


packages/data-access/src/data-access.ts
```diff
@@ -607,7 +610,7 @@ export default class DataAccess implements DataAccessTypes.IDataAccess {
     const transactionsStorageLocation = Array(transactions.length).fill(location);
 
     // Gets the list of storage meta of the transactions found
-    const storageMeta = Array(transactions.length).fill(meta);
+    const storageMeta = transactions.map(() => meta);
```
benjlevesque, line 613: this allows better typing

packages/data-access/src/data-access.ts
```diff
@@ -305,7 +305,8 @@ export default class DataAccess implements DataAccessTypes.IDataAccess {
     return transactionsAndMetaPerBlocks.reduce(
       (accumulator: DataAccessTypes.IReturnGetTransactions, elem) => ({
         meta: {
-          storageMeta: accumulator.meta.storageMeta.concat(elem.storageMeta),
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
benjlevesque, line 308: the initial value sets the storageMeta, so I feel it's better to disable eslint than changing the logic

alexandre-abrioux (approved):

PR #720, opened 2021-12-23
chore: detect near mainnet payments with a subgraph

yomarion:

### DIFF
--- a/packages/payment-detection/src/near-info-retriever.ts
+++ b/packages/payment-detection/src/near-info-retriever.ts
@@ -1,12 +1,16 @@
 import { PaymentTypes } from '@requestnetwork/types';
+import { getTheGraphClient, TheGraphClient } from '.';
 
-const NEAR_WEB_SOCKET_URL = 'wss://near-explorer-wamp.onrender.com/ws';
+// FIXME#1: when Near subgraphes can retrieve a txHash, replace the custom IPaymentNetworkEvent with PaymentTypes.ETHPaymentNetworkEvent
+interface NearSubGraphPaymentEvent extends PaymentTypes.IETHPaymentEventParameters {
+  receiptId: string;
+}
 
 /**
  * Gets a list of transfer events for a set of Near payment details
  */
 export class NearInfoRetriever {
-  private nearWebSocketUrl: string;
+  private client: TheGraphClient<'near'>;
   /**
    * @param paymentReference The reference to identify the payment
    * @param toAddress Address to check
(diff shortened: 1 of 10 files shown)


packages/payment-detection/src/near-info-retriever.ts
```diff
@@ -17,124 +21,45 @@ export class NearInfoRetriever {
...
       name: this.eventName,
       parameters: {
-        block: transaction.block,
-        confirmations: transaction.confirmations,
-        txHash: transaction.txHash,
+        block: p.block,
+        confirmations: p.block,
+        // Cf. FIXME#1 above
+        // txHash: transaction.txHash,
+        receiptId: p.receiptId ?? 'FIXME String!',
```
yomarion, line 60: I wait for the current graph indexing to be finished before I push a version where the receiptId cannot be undefined.

packages/payment-detection/src/thegraph/index.ts
```diff
@@ -9,24 +10,29 @@ const BASE_URL = `https://api.thegraph.com`;
 /**
  * A GraphQL client to query Request's subgraph.
  */
-export type TheGraphClient = ReturnType<typeof getSdk>;
+export type TheGraphClient<
+  TGraphClientVariant extends 'near' | null = null
+> = TGraphClientVariant extends 'near' ? ReturnType<typeof getSdkNear> : ReturnType<typeof getSdk>;
```
yomarion, line 15: @benjlevesque , I'd like your opinion about this. It may look overkill but the queries interfaces and possibilities between EVM and Near payments are very different.

packages/payment-detection/src/thegraph/index.ts
```diff
@@ -9,24 +10,29 @@ const BASE_URL = `https://api.thegraph.com`;
 /**
  * A GraphQL client to query Request's subgraph.
  */
-export type TheGraphClient = ReturnType<typeof getSdk>;
+export type TheGraphClient<
```
yomarion, line 13: @benjlevesque , what do you think of this pattern?

packages/payment-detection/src/thegraph/index.ts
```diff
@@ -9,24 +10,29 @@ const BASE_URL = `https://api.thegraph.com`;
...
   const baseUrl = options?.baseUrl || network === 'private' ? 'http://localhost:8000' : BASE_URL;
   // Note: it is also possible to use the IPFS hash of the subgraph
   //  eg. /subgraphs/id/QmcCaSkefrmhe4xQj6Y6BBbHiFkbrn6UGDEBUWER7nt399
   //  which is a better security but would require an update of the
   //  library each time the subgraph is updated, which isn't ideal
   //  for early testing.
   const url = `${baseUrl}/subgraphs/name/requestnetwork/request-payments-${network}`;
-  return getSdk(new GraphQLClient(url, options));
+  if (network === 'near') {
+    return getSdkNear(new GraphQLClient(url, options)) as TheGraphClient<TGraphClientVariant>;
```
benjlevesque, line 33: Is the cast necessary ? Specifying the return type of the method doesn't work here ?

packages/payment-detection/src/near-info-retriever.ts
```diff
@@ -17,124 +21,45 @@ export class NearInfoRetriever {
...
     private proxyContractName: string,
-    private procedureName: string,
     private eventName: PaymentTypes.EVENTS_NAMES,
     private network: string,
   ) {
     if (this.network !== 'aurora' && this.network !== 'aurora-testnet') {
       throw new Error('Near input data info-retriever only works with Near mainnet and testnet');
     }
-    this.nearWebSocketUrl = NEAR_WEB_SOCKET_URL;
+    if (this.network !== 'aurora') {
```
benjlevesque, line 30: If we keep aurora as the name of the chain, why not naming the subgraph aurora?

benjlevesque (approved):
Great simplification
Shouldn't this PR also remove autobahn? 🤩

packages/payment-detection/src/near-info-retriever.ts
yomarion: Hm good point. Overall I'm super confused about which name to use where. And yes it should actually be the time to remove autobahn finally !!!

packages/payment-detection/src/thegraph/index.ts
yomarion: Pretty sure yes. Another option would be to not make the method generic, and instead have 2 types `TheGraphClient` and `TheGraphNearClient`, and two methods `getTheGraphClient` and `getTheGraphNearClient`.

I think it actually makes sense because the genericity of this method is not used and since the types are quite different, I don't see how it could factorize anything.

I'll remove it.

packages/payment-detection/test/near-native.test.ts
```diff
@@ -22,14 +22,14 @@ const mockAdvancedLogic: AdvancedLogicTypes.IAdvancedLogic = {
...
+const paymentAddress = 'gus.near';
 const request: any = {
-  requestId: '017a738821782329122ffb1b944dc2bbcecc56fdc8d95b050fe49a1fc04349a9c4',
+  requestId: '01cc915b9b2cde7f97be812679f6212d3d464965cc27d423d55e37962b06937a7f',
   currency: {
-    network: 'aurora-testnet',
+    network: 'aurora',
     type: RequestLogicTypes.CURRENCY.ETH,
-    value: 'NEAR-testnet',
+    value: 'NEAR',
```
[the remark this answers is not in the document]
yomarion: That's the token name

packages/payment-detection/src/thegraph/index.ts
bertux: That's why I've thought it was easier for me to have a separated file for superfluid subgraph methods which can be imported in Superfluid info retriever instead of the existing one.

packages/payment-detection/src/near-info-retriever.ts
yomarion: In the end, the chain should be renamed Near instead of Aurora, but I prefere making it a different topic 👌 @benjlevesque

PR #721, opened 2021-12-23
fix: tslib dependency in smart-contracts

yomarion:
## Description of the changes
- The devDependency was breaking some `smart-contracts` imports like in `payments-subgraph`


alexandre-abrioux (approved):
Thanks! That's a mistake from https://github.com/RequestNetwork/requestNetwork/pull/683, I imported `tslib` in the wrong place for this package.

PR #735, opened 2022-01-17
ci: adjust jest maxWorkers

alexandre-abrioux:
## Description of the changes

Adds `--maxWorkers=1` to test commands in CI to prevent Jest from spawning 31 workers on VMs with only 2vCPU, resulting in failing tests, like here: https://app.circleci.com/pipelines/github/RequestNetwork/requestNetwork/5348/workflows/c6e3c200-7b29-4afa-8756-6367e04310fa/jobs/86870/steps

`maxWorkers` should be set to amount of CPUs minus one for Jest's main thread, so `2-1=1` for CircleCI's medium Docker executor that has 2vCPU.

The reason it is spawning 31 workers is that CircleCI hosts have 32 (real) CPUs, making Jest think it is running on a 32 cores machines whereas it is running on a 2vCPU machine.

See: https://support.circleci.com/hc/en-us/articles/360005442714-Your-test-tools-are-smart-and-that-s-a-problem-Learn-about-when-optimization-goes-wrong-

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
           command: 'yarn workspace @requestnetwork/smart-contracts run deploy'
       - run:
           name: 'Test request-client.js'
-          command: 'yarn workspace @requestnetwork/request-client.js run test --ci --maxWorkers=2'
+          command: 'yarn workspace @requestnetwork/request-client.js run test --ci --maxWorkers=1'
       - persist_to_workspace:
           root: *working_directory
           paths:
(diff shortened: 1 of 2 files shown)


KolevDarko (approved):

benjlevesque (approved):

.circleci/config.yml
```diff
@@ -252,7 +252,7 @@ jobs:
           command: 'yarn workspace @requestnetwork/smart-contracts run deploy'
       - run:
           name: 'Test request-node'
-          command: 'yarn workspace @requestnetwork/request-node run test'
+          command: 'yarn workspace @requestnetwork/request-node run test --ci'
```
bertux, line 255: why it's not added here and in some other places ?

alexandre-abrioux: Because they already have the `--runInBand` option which is equivalent to running with 1 worker, and both options are not compatible with each other, see: https://github.com/RequestNetwork/requestNetwork/blob/2c967592c9e295cf1f8ae6d93ecc0ecf8d66dacd/packages/request-node/package.json#L34

bertux (approved):

PR #743, opened 2022-01-19
fix(contract): ETH conversion native hashes per chain

yomarion:
## Description of the changes

ChainlinkConversionPath v0.2.0
- Configure the native token hash in the constructor of Chainlink conversion path

ETHConversionProxy v0.2.0
- Whitelisted admins can edit the conversion path contract
- Whitelisted admins can edit the proxy contract

Live payment script:
- nonce 10: ChainlinkConversionPath v0.2.0
- nonce 11: ETHConversionProxy v0.2.0

### DIFF
--- a/packages/smart-contracts/scripts/3_deploy_chainlink_contract.ts
+++ b/packages/smart-contracts/scripts/3_deploy_chainlink_contract.ts
@@ -36,7 +36,7 @@ export default async function deploy(args: any, hre: HardhatRuntimeEnvironment) 
 
   const conversionPathInstance = await (
     await hre.ethers.getContractFactory('ChainlinkConversionPath', deployer)
-  ).deploy();
+  ).deploy(ETH_hash);
 
   // all these aggregators are for test purposes
   await conversionPathInstance.updateAggregatorsList(
--- a/packages/smart-contracts/scripts/deploy-payments.ts
+++ b/packages/smart-contracts/scripts/deploy-payments.ts
@@ -11,11 +11,13 @@ import {
 import { deploySwapConversion } from './erc20-swap-to-conversion';
 import { deployERC20ConversionProxy, deployETHConversionProxy } from './conversion-proxy';
 import { DeploymentResult, deployOne } from './deploy-one';
-import { uniswapV2RouterAddresses } from './utils';
+import { uniswapV2RouterAddresses, jumpToNonce } from './utils';
 import { Contract } from 'ethers';
 // eslint-disable-next-line
 // @ts-ignore Cannot find module
 import { ChainlinkConversionPath } from '../src/types/ChainlinkConversionPath';
+import { CurrencyManager } from '@requestnetwork/currency';
+import { RequestLogicTypes } from '@requestnetwork/types';
 
 /**
  * Script ensuring all payment contracts are deployed and usable on a live chain.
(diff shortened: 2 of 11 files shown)


packages/smart-contracts/scripts/3_deploy_chainlink_contract.ts
```diff
@@ -36,7 +36,7 @@ export default async function deploy(args: any, hre: HardhatRuntimeEnvironment)
 
   const conversionPathInstance = await (
     await hre.ethers.getContractFactory('ChainlinkConversionPath', deployer)
-  ).deploy();
+  ).deploy(ETH_hash);
```
yomarion, line 39: This is a local deployment only, in local we use the same hash as on mainnet.

packages/smart-contracts/src/contracts/EthConversionProxy.sol
```diff
@@ -120,4 +121,26 @@ contract EthConversionProxy is ReentrancyGuard {
         amountToPay = (_requestAmount * rate) / decimals;
         amountToPayInFees = (_feeAmount * rate) / decimals;
     }
+
+    /**
+     * @notice Update the conversion path contract used to fetch conversions
+     * @param _chainlinkConversionPathAddress address of the conversion path contract
+     */
+    function updateConversionPathAddress(address _chainlinkConversionPathAddress)
```
yomarion, line 129: @vrolland  ideally if you could have a look at these changes 🙏

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -1,226 +1,231 @@
...
-      emit AggregatorUpdated(_inputs[i], _outputs[i], _aggregators[i]);
+    uint256 constant PRECISION = 1e18;
+    uint256 constant NATIVE_TOKEN_DECIMALS = 18;
+    uint256 constant FIAT_DECIMALS = 8;
+    address public nativeTokenHash;
+
+    /**
+     * @param _nativeTokenHash hash of the native token
+     */
+    constructor(address _nativeTokenHash) {
```
yomarion, line 32: @vrolland  ideally if you could have a look at these changes 🙏

packages/smart-contracts/src/contracts/ChainlinkConversionPath.sol
```diff
@@ -1,226 +1,231 @@
...
-  {
-    require(_inputs.length == _outputs.length, "arrays must have the same length");
-    require(_inputs.length == _aggregators.length, "arrays must have the same length");
-
-    // For every conversions of the path
-    for (uint i; i < _inputs.length; i++) {
-      allAggregators[_inputs[i]][_outputs[i]] = _aggregators[i];
-      emit AggregatorUpdated(_inputs[i], _outputs[i], _aggregators[i]);
+    uint256 constant PRECISION = 1e18;
+    uint256 constant NATIVE_TOKEN_DECIMALS = 18;
```
alexandre-abrioux, line 25: Can this be different for another chain? I can't find the spec for this.

yomarion: It's not completely impossible but unlikely. If it happens we can anticipate and adapt the contract before we deploy it (either with a hardcoded value or add a constructor parameter)

alexandre-abrioux (approved):

packages/smart-contracts/scripts/deploy-payments.ts
```diff
@@ -215,27 +228,38 @@ export async function deployAllPaymentContracts(
...
-      nonceCondition: 7,
-    });
+    if (!args.simulate) {
+      throw new Error('Missing implementation new version of ERC20 Conversion Proxy');
+    }
 
-    // Batch 4
-    await runDeploymentBatch_4(chainlinkInstanceAddress, ethFeeProxyAddress);
+    // Batch 2
+    await runDeploymentBatch_2(chainlinkInstance, erc20FeeProxyAddress);
```
leoslr, line 262: Maybe I missed something, but it seems that when running :
```javascript
await runDeploymentBatch_2(chainlinkInstance, erc20FeeProxyAddress);
```
You should already have a nonce greater than 10 as it is the nonce condition to deploy the ChainlinkConversionPath contract. However there are several hardcoded nonce conditions lower than 10 in this function.

KolevDarko (approved):

packages/smart-contracts/scripts/deploy-payments.ts
yomarion: That's why I added the error thrown just above, line 258, I'll fix that later but now I want to solve the incident quickly.

vrolland (commented):
LGTM

PR #752, opened 2022-01-27
chore(payment-detection): clean Retrievers type. Document.

yomarion:
## Description of the changes
- Revamped README
- Renamed `IPaymentNetworkInfoRetriever` > `IPaymentRetriever`
- Renamed `IPaymentNetworkBaseInfoRetriever` > `IEventRetriver`
- Moved the definition of retrievers from `types` to `payment-detection`

### DIFF
--- a/packages/payment-detection/README.md
+++ b/packages/payment-detection/README.md
@@ -17,7 +17,8 @@ Based on information found in the payment network state, and included manual pay
 # Design: retrievers and detectors
 
 This library relies on two concepts:
-- **Retrievers** perform RPC calls or TheGraph and fetch relevant events. Balance-impacting events are fetched with `InfoRetrievers`, implementing the  `getTransferEvents()` method (cf. [IPaymentRetriever](./src/types.ts))
+
+- **Retrievers** perform RPC calls or TheGraph and fetch relevant events. Balance-impacting events are fetched with `InfoRetrievers`, implementing the `getTransferEvents()` method (cf. [IPaymentRetriever](./src/types.ts))
 - **Payment detectors** call retrievers and interpret events according to the payment network (cf. [Abstract PaymentDetectorBase](./src/payment-detector-base.ts))
 
 ## Subgraph-based payment retrievers
@@ -27,6 +28,7 @@ For TheGraph-based information retrieval, we implement a generic method `getTheG
 The automated type generation is configured within files `./codegen.yml` (for EVM chains) and `./codegen-near.yml` (for Near) and output in `./src/thegraph/generated`. It depends on the deployed subgraphes schema and on the queries.
 
 The code generation is included in the pre-build script and can be run manually:
+
 ```
 yarn codegen
 ```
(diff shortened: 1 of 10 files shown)


packages/payment-detection/src/any/retrievers/thegraph.ts
```diff
@@ -1,13 +1,14 @@
-import { PaymentTypes } from '@requestnetwork/types';
 import { CurrencyDefinition } from '@requestnetwork/currency';
+import { PaymentTypes } from '@requestnetwork/types';
 import { BigNumber, utils } from 'ethers';
+import { IPaymentRetriever } from 'payment-detection/src/types';
```
benjlevesque, line 4: I guess this is VS Code doing, but this kind of import won't work in JS. I think this works because it's only a type, but I recommend against it


```suggestion
import { IPaymentRetriever } from '../types';
```

packages/payment-detection/README.md
```diff
@@ -1,22 +1,34 @@
...
+Based on information found in the payment network state, and included manual payment declarations, the library will perform queries and feed the property `balance` of the request with:
+
+- `balance`: the detected amount paid on the request, in request currency
+- `events`: all the payments, payment declarations, refunds, and other balance-related events with the amount, timestamp etc...
+
+# Design: retrievers and detectors
+
+This library relies on two concepts:
+
+- **Retrievers** perform RPC calls or TheGraph and fetch relevant events. Balance-impacting events are fetched with `InfoRetrievers`, implementing the `getTransferEvents()` method (cf. [IPaymentRetriever](./src/types.ts))
```
benjlevesque, line 21:
```suggestion
- **Retrievers** perform RPC or TheGraph calls and fetch relevant events. Balance-impacting events are fetched with `InfoRetrievers`, implementing the `getTransferEvents()` method (cf. [IPaymentRetriever](./src/types.ts))
```

packages/payment-detection/README.md
```diff
@@ -1,22 +1,34 @@
...
+- **Payment detectors** call retrievers and interpret events according to the payment network (cf. [Abstract PaymentDetectorBase](./src/payment-detector-base.ts))
+
+## Subgraph-based payment retrievers
+
+For TheGraph-based information retrieval, we implement a generic method `getTheGraphClient()` in `./src/thegraph/index.ts`, based on the queries in `/src/thegraph/queries` and automated types.
```
benjlevesque, line 26: what do you mean by generic here? In TS generic usually refer to `function foo<T>(){...}`. 
I don't this this method is actually very "generic" anyway, as it's only exposing a little set of methods, generated based on the queries
```suggestion
For TheGraph-based information retrieval, a client can be retrieved using `getTheGraphClient()` in `./src/thegraph/index.ts`. It provides a strongly typed interface, generated based on the queries in `/src/thegraph/queries`.
```

benjlevesque (changes requested):
change imports, otherwise 👍

leoslr (approved):

benjlevesque (approved):

PR #755, opened 2022-02-01
Feat: Erc20swapToPay w/fees

Dadogg80:
## Feature: ERC20SwapToPay with swap fee.
Context: 
The purpose of this new feature is to be able to collect a small `swapFee` from every transaction executed by this smart contract. 

### Code addition to ERC20SwapToPay.sol

-Added an internal function to calculate the fee amount. 
`_calculateSwapFee(uint256 _amount) internal view returns (uint256 _swapFeeAmount)`.

-Added a public function to change the swap fee.
`function setSwapFee(uint256 _newSwapFee) public onlyWhitelisted`.

-Added a public function to change the swap fee address.
`function setFeeAddress(address _newFeeAddress) public onlyWhitelisted`.

**Note:** Only whitelisted admins can change the `currentSwapFee` and `swapFeeAddress`.


packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -118,24 +150,51 @@ contract ERC20SwapToPay is Ownable {
...
 
   function setRouter(address _newSwapRouterAddress) public onlyOwner {
     swapRouter = IUniswapV2Router02(_newSwapRouterAddress);
   }
+
+  /** 
+   * Admin function to edit the swap fee address.
+   * @param _newFeeAddress Address that receives the swap fees.
+   */
+  function setFeeAddress(address _newFeeAddress) public onlyWhitelisted {
```
benjlevesque, line 187: any reason why this is not using `onlyOwner`?

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -118,24 +150,51 @@ contract ERC20SwapToPay is Ownable {
...
   }
 
+  /**
+   * Internal function to calculate the amount of request fee to pay with a transaction.
+   * @param _amount The Wei amount to calculate the fee from.
+   * @dev The _amount has to be nominated in Wei.
+   * @dev returns the _feeAmount in Wei. 
+   */
+  function _calculateSwapFee(uint256 _amount) internal view returns (uint256 _swapFeeAmount) {
+    _swapFeeAmount = (_amount.mul(currentSwapFee)).div(1000);
```
benjlevesque, line 169: why 1000 🤔 ?

benjlevesque (changes requested):
nice! you need to add tests before merging 👍

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -82,15 +107,22 @@ contract ERC20SwapToPay is Ownable {
     IERC20 spentToken = IERC20(_path[0]);
     IERC20 requestedToken = IERC20(_path[_path.length-1]);
 
+    
     uint256 requestedTotalAmount = _amount + _feeAmount;
 
-    require(spentToken.allowance(msg.sender, address(this)) > _amountInMax,
+    // make sure the allowance includes the swapFee.
+    require(spentToken.allowance(msg.sender, address(this)) > (_amountInMax + _calculateSwapFee(requestedTotalAmount)),
```
KolevDarko, line 114: Because you imported SafeMath above, shouldn't it be .add instead of +
But I [think 0.8.0](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/) also has the overflow checks ? so SafeMath is not required ?

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
Dadogg80: `onlyWhitelisted` was a requirement in the spec I got. 
The benefit is that It allows multiple addresses to be whitelisted instead of `onlyOwner`.

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
Dadogg80: The reason was due to the `swapFee` being 0.5%.
In `setSwapFee(uint256 _newSwapFee)` the minimum input value would be one digit. 

For example: let's get a 0.5% swap fee of 1000.  
If we have the `_newSwapFee` set to `5`. 
then `1000.mul(5).div(1000)` should return `5` if I'm not mistaken?

benjlevesque: I thought it could be something like this, but then it should be very clearly documented in the contract
maybe even increase to 2 decimals, to have more control over the rate, like `1.25` ? (not sure about this)

Dadogg80: Yes, I did write it as a `@dev` comment and gave three examples. 
But I'm always open to suggestions to better documenting it.

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
Dadogg80: Yes, I used SafeMath.sol, it was a good habit I had, that turned into a bad habit after `0.8.0`, when `overflow` protection became standard from 0.8.0.
Do you want me to remove it? 
I haven't checked which requires most gas of `.add()` or `+`.
I think using SafeMath might be more expensive, so I´ll remove it. 
:-)

Dadogg80: Also, I didn't change any of the current code. so that´s why the +, etc is there. 

I wanted to change all `require(condition, string);` into `if () revert customError();` which would save some gas on error messages.

KolevDarko: I was just curious because I saw the import  but it wasn't used.
But about changing things and saving gas etc.., I'll leave that decision to Yoann or Benji, don't mind me 🙂

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
Dadogg80: fixed

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -21,18 +22,42 @@ interface IUniswapV2Router02 {
...
+    address _swapFeeAddress,
+    uint256 _currentSwapFee
+  ){
+    admin = _adminAddress;
     swapRouter = IUniswapV2Router02(_swapRouterAddress);
     paymentProxy = IERC20FeeProxy(_paymentProxyAddress);
+    swapFeeAddress = _swapFeeAddress;
+    currentSwapFee = _currentSwapFee;
+
+    super.addWhitelisted(address(admin));
```
yomarion, line 57:
```suggestion
    addWhitelisted(address(admin));
```

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -82,15 +107,33 @@ contract ERC20SwapToPay is Ownable {
...
+      msg.sender,
+      address(swapFeeAddress),
+      _calculateSwapFee(requestedTotalAmount)
+      ), 
+      "SwapFee failed!"
+    );
+    
     // Allow the router to spend all this contract's spentToken
-    if (spentToken.allowance(address(this),address(swapRouter)) < _amountInMax) {
+    if (spentToken.allowance(address(this), address(swapRouter)) < _amountInMax) {
```
yomarion, line 136: Shouldn't you acount for the fees here as you do in line 117?

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -21,18 +22,42 @@ interface IUniswapV2Router02 {
...
+  * @param _paymentProxyAddress Address to the paymentProxy.
+  * @param _swapFeeAddress The address to collect fees.
+  * @param _currentSwapFee The fee % to pay with every transaction, example: 50 = 0.5%, 100 = 1%.
+  */
+  constructor(
+    address _adminAddress,
+    address _swapRouterAddress,
+    address _paymentProxyAddress,
+    address _swapFeeAddress,
+    uint256 _currentSwapFee
```
yomarion, line 49:
```suggestion
    uint256 _swapFeePercent
```
+ other usage and comments

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -21,18 +22,42 @@ interface IUniswapV2Router02 {
...
+  address public swapFeeAddress;
+  address public admin;
+
+  /** 
+  * Constructor. 
+  * @param _adminAddress Address to the admin.
+  * @param _swapRouterAddress Address to the swapRouter.
+  * @param _paymentProxyAddress Address to the paymentProxy.
+  * @param _swapFeeAddress The address to collect fees.
+  * @param _currentSwapFee The fee % to pay with every transaction, example: 50 = 0.5%, 100 = 1%.
```
yomarion, line 42:
```suggestion
  * @param _swapFeePercent The fee % to pay with every transaction, with 2 decimals. Example: 50 = 0.5%.
```

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -82,15 +107,33 @@ contract ERC20SwapToPay is Ownable {
...
+    
     uint256 requestedTotalAmount = _amount + _feeAmount;
 
-    require(spentToken.allowance(msg.sender, address(this)) > _amountInMax,
-                                 "Not sufficient allowance for swap to pay.");
+    // make sure the allowance includes the swapFee.
+    require(
+      spentToken.allowance(
+        msg.sender, 
+        address(this)) > (_amountInMax + _calculateSwapFee(requestedTotalAmount)
```
yomarion, line 117: It's probably wise to call `_calculateSwapFee` only once and to use the constant where needed.

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -118,24 +161,52 @@ contract ERC20SwapToPay is Ownable {
...
+    swapFeeAddress = _newFeeAddress;
+  }
+
+  /** 
+   * Admin function to edit the swap fee in %.
+   * @param _newSwapFee  the value to set as new swap fee.
+   * @dev swapFee in %. Example: 1000 = 10% swap fee, 500 = 5% swap fee, 50 = 0.5% swap fee.
+   * @dev Example: 125 = 1.25% swap fee, 175 = 1.75% swap fee, 210 = 2.10% swap fee.
+   */
+  function setSwapFee(uint256 _newSwapFee) external onlyWhitelisted {
```
yomarion, line 208:
```suggestion
  function setSwapFeePercent(uint256 _newSwapFeePercent) external onlyWhitelisted {
```

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -118,24 +161,52 @@ contract ERC20SwapToPay is Ownable {
...
+   * @param _newFeeAddress Address that receives the swap fees.
+   */
+  function setFeeAddress(address _newFeeAddress) external onlyWhitelisted {
+    swapFeeAddress = _newFeeAddress;
+  }
+
+  /** 
+   * Admin function to edit the swap fee in %.
+   * @param _newSwapFee  the value to set as new swap fee.
+   * @dev swapFee in %. Example: 1000 = 10% swap fee, 500 = 5% swap fee, 50 = 0.5% swap fee.
```
yomarion, line 205:
```suggestion
   * @dev swapFee in % with 2 decimals. Example: 1000 = 10% swap fee.
```

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -118,6 +120,47 @@ describe('contract: SwapToPay', () => {
...
+        ethers.utils.keccak256(referenceExample),
+        '1',
+        ethers.utils.getAddress(builder),
+      );
+
+    const finalFeeBalance = await spentErc20.balanceOf(swapFeeAddress);
+    const finalBuilderBalance = await paymentNetworkErc20.balanceOf(builder);
+    const finalIssuerBalance = await paymentNetworkErc20.balanceOf(to);
+
+    expect(initialFeeBalance.toNumber()).to.equals(0);
```
yomarion, line 153: This kind of check should be made as early as possible (usually commented "sanity checks" to distinguish with actual tests)

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -118,6 +120,47 @@ describe('contract: SwapToPay', () => {
...
+        exchangeRateOrigin + 1000, // _uniswapDeadline. 100 -> 1000: Too low value may lead to error (network dependent)
+      ),
+    )
+      .to.emit(erc20FeeProxy, 'TransferWithReferenceAndFee')
+      .withArgs(
+        ethers.utils.getAddress(paymentNetworkErc20.address),
+        to,
+        '100',
+        ethers.utils.keccak256(referenceExample),
+        '1',
```
yomarion, line 145:
```suggestion
        '2',
```
I think it's wise to test with different expected amounts to be sure we never mix both kinds of fees.

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -118,6 +120,47 @@ describe('contract: SwapToPay', () => {
...
+        ethers.utils.getAddress(builder),
+      );
+
+    const finalFeeBalance = await spentErc20.balanceOf(swapFeeAddress);
+    const finalBuilderBalance = await paymentNetworkErc20.balanceOf(builder);
+    const finalIssuerBalance = await paymentNetworkErc20.balanceOf(to);
+
+    expect(initialFeeBalance.toNumber()).to.equals(0);
+    expect(finalFeeBalance.toNumber()).to.equals(1);
+    expect(finalBuilderBalance.toNumber()).to.equals(1);
```
yomarion, line 155:
```suggestion
    expect(finalBuilderBalance.toNumber()).to.equals(2);
```

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -118,6 +120,47 @@ describe('contract: SwapToPay', () => {
...
+        '1',
+        ethers.utils.getAddress(builder),
+      );
+
+    const finalFeeBalance = await spentErc20.balanceOf(swapFeeAddress);
+    const finalBuilderBalance = await paymentNetworkErc20.balanceOf(builder);
+    const finalIssuerBalance = await paymentNetworkErc20.balanceOf(to);
+
+    expect(initialFeeBalance.toNumber()).to.equals(0);
+    expect(finalFeeBalance.toNumber()).to.equals(1);
```
yomarion, line 154:
```suggestion
    expect(finalFeeBalance.toNumber()).to.equals(1); // Contract deployed with 1% fee locally
```

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -300,4 +343,14 @@ describe('contract: SwapToPay', () => {
...
     });
   });
+  describe('OnlyWhitelisted admin', () => {
+    it('Should revert when `setFeeAddress` is executed by a non admin', async () => {
+      expect(
+        testSwapToPay.connect(from).setFeeAddress('0x821aEa9a577a9b44299B9c15c88cf3087F3b5544'),
+      ).to.be.reverted;
+    });
+    it('Should revert when `setSwapFee` is executed by a non admin', async () => {
+      expect(testSwapToPay.connect(from).setSwapFee(50)).to.be.reverted;
```
yomarion, line 353: Please add revert messages to the tests so we are sure of why it fails.

packages/smart-contracts/test/contracts/ERC20SwapToPay.test.ts
```diff
@@ -300,4 +343,14 @@ describe('contract: SwapToPay', () => {
       expect(finalContractRequestBalance.toNumber()).to.equals(0);
     });
   });
+  describe('OnlyWhitelisted admin', () => {
```
yomarion, line 346:
```suggestion
  describe('Admin tasks', () => {
```

Also please test that we can change fee address and percentage

packages/smart-contracts/src/contracts/ERC20SwapToPay.sol
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity 0.8.4;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
+import "./legacy_openzeppelin/contracts/access/roles/WhitelistedRole.sol";
```
yomarion, line 6: Since you're using `solidity 0.8.4`, can't you use one of the non-legacy OZ contracts?

yomarion (approved):
Good job in progress

packages/smart-contracts/scripts/2_deploy-main-payments.ts
```diff
@@ -76,7 +76,13 @@ export default async function deploy(args: any, hre: HardhatRuntimeEnvironment):
     await testERC20Instance.transfer(FakeSwapRouterAddress, '10000000000000000000000000000');
     // SwapToPay
     const { address: ERC20SwapToPayAddress } = await deployOne(args, hre, 'ERC20SwapToPay', {
```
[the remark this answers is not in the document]
Dadogg80: done

PR #760, opened 2022-02-04
chore(request-node): eip-1559 for TheGraphDataAccess

alexandre-abrioux:
## Description of the changes

Gas limit can be too low for transactions submitted through `TheGraphDataAccess`. This adds an EIP-1559 gas price estimation for transactions submitted with this data access implementation.

PR #765, opened 2022-02-08
fix(eip1559): use hex number for eth_feehistory 'blockcount' parameter

alexandre-abrioux:
🚨 https://github.com/rainbow-me/fee-suggestions/pull/25 is needed before merging this PR.

## Description of the changes

Quicknode's RPC servers do not allow the use of an integer for the `blockCount` argument of `eth_feeHistory`.
Indeed, the doc specifies that it should be an hexadecimal number:
https://infura.io/docs/ethereum#operation/eth_feeHistory
Infura and Alchemy are more permissive in this regard.

This change will use the `hex` string instead of an integer to check if `eth_feeHistory` is supported on the RPC backend.

We need to wait for an update of the `rainbow-me/fee-suggestions` library before merging this because this library is using an integer and not a `hex` string in its RPC calls. If we merged this PR right away we would generate errors when running the request-node against Quicknode's backend.


KolevDarko (approved):

bertux:
So why not switch first to Alchemy RPC ?

bertux (approved):

alexandre-abrioux:
@bertux Infura & Alchemy do not support Gnosis Chain. But this PR is not urgent since the original issue was that we were not calling the `GasPriceDefiner` in `TheGraphStorage`, so we were using a gas price that was too low sometimes. But this was fixed in https://github.com/RequestNetwork/requestNetwork/pull/760.

alexandre-abrioux:
Marking this as Draft, PR is stale

alexandre-abrioux:
rebased to fix conflicts following refactor in https://github.com/RequestNetwork/requestNetwork/pull/931

benjlevesque (approved):

alexandre-abrioux:
https://github.com/rainbow-me/fee-suggestions/pull/25 has been merged and the lib has been updated by @leoslr in https://github.com/RequestNetwork/requestNetwork/pull/1072, merging now

PR #774, opened 2022-02-15
fix: price calculation

leoslr:
Fixes edge cases where we lost precision.

e.g. :
`item.unitPrice = 33.33`
`item.quantity = 1`
`item.tax.amount = 20`
`item.discount = 0`

The "true" result is `39.996`. This function returned `39.99`, it now returns `40.00`.

### DIFF
--- a/packages/data-format/src/format/rnf_invoice/utils.ts
+++ b/packages/data-format/src/format/rnf_invoice/utils.ts
@@ -9,7 +9,6 @@ export const getInvoiceTotal = (invoice: Invoice): BigNumber => {
 };
 
 export const getInvoiceLineTotal = (item: InvoiceItem): BigNumber => {
-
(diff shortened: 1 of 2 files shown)


kevindavee (approved):

alexandre-abrioux (changes requested):
- Please add a test case for this in `packages/data-format/test/rnf_invoice/utils.test.ts`.
- Do you know where this issue occurs exactly? I'm not sure this is the way to go at it, to me it looks like it's more of a display issue rather than a calculation issue.

leoslr:
@alexandre-abrioux Nice solution with the FixedNumber class, indeed `preciselyOne` is not needed anymore. It's weird though, we can instantiate a FixedNumber from a floating string but not a floating number.

I also added the test describing the edge case and it passes. (I checked again but indeed, it was not passing before).

PR #780, opened 2022-02-17
feat: ether to smart wallet payment

leoslr:
## Description of the changes

Enables to pay requests in ether when the recipient is a proxy smart contract, like smart wallet such as Gnosis Safe.

The `.transfer` is limited to 2300 gas, so when a payment recipient was a Gnosis Safe, we did not pass enough gas for it to execute it's fallback function. I changed it to `.call` which pass all the remaining transaction gas.

New tests cover for this specific case.

### DIFF
--- a/packages/smart-contracts/src/contracts/EthereumFeeProxy.sol
+++ b/packages/smart-contracts/src/contracts/EthereumFeeProxy.sol
@@ -1,78 +1,75 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
 
 /**
  * @title EthereumFeeProxy
  * @notice This contract performs an Ethereum transfer with a Fee sent to a third address and stores a reference
  */
-contract EthereumFeeProxy is ReentrancyGuard{
-  // Event to declare a transfer with a reference
-  event TransferWithReferenceAndFee(
-    address to,
-    uint256 amount,
-    bytes indexed paymentReference,
-    uint256 feeAmount,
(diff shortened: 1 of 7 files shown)


packages/smart-contracts/src/contracts/test/EtherFallbackPayment.sol
```diff
@@ -0,0 +1,15 @@
...
+contract EtherPaymentFallback {
+    event SafeReceived(address indexed sender, uint256 value);
+
+    /// @dev Fallback function accepts Ether transactions.
+    receive() external payable {
+        emit SafeReceived(msg.sender, msg.value);
+    }
+
+    constructor() {}
+}
```
leoslr, line 15: Gnosis Safe extends this contract to allow for ether payments, no need to import the whole Gnosis Safe Contract.
(Tests fail with our previous EthFeeProxy and EthProxy).

packages/smart-contracts/src/contracts/EthereumFeeProxy.sol
```diff
@@ -1,78 +1,75 @@
...
-    nonReentrant
-    public
-    payable
-  {
-    _to.transfer(_amount);
-    _feeAddress.transfer(_feeAmount);
-    // transfer the remaining ethers to the sender
-    payable(msg.sender).transfer(msg.value - _amount - _feeAmount);
+        // transfer the remaining ethers to the sender
+        (bool sendBackSuccess, ) = payable(msg.sender).call{
```
leoslr, line 68: Modified `.transfer` to `.call` here too. This way we cover the case where a gnosis safe is the one initiating a payment and we have to send ether back to it.

bertux (approved):
This seems a  good extension :-)

benjlevesque (approved):
very thourough, nice work

PR #784, opened 2022-02-21
refactor: payment-processor

leoslr:
## Description of the changes

Refactoring of the payment-processor package.

Before:
- For some payment methods we had signing, preparing and encoding functions, in others only signing and encoding.
- Same kind of issue for approval functions.
- In some files we used `Web3Provider `in other `Provider`

Now:
- All payment methods have signing, preparing and encoding functions.
- All kind of approvals have distinct `approvefNeeded`, `hasApproval` and `prepareApproval` functions 
- All files use `Provider` which allows to pass `ethers.providers.JsonRpcProvider` which can be useful if we only want to prepare a transaction and sign it externally.


benjlevesque (approved):
very nice refacto! well done

KolevDarko (approved):

alexandre-abrioux (approved):
👌

PR #786, opened 2022-02-24
feat: transaction encoding

leoslr:
## Description of the changes

It exports new methods from the `payment-processor` package:
- Encoding of the approval transaction for a request payment
- Check for knowing if an approval is needed for a request payment
- Encoding of an approval transaction if needed for a request payment
- Encoding of the payment transaction for a request payment
- Encoding of all necessary transactions (approval if needed + payment) for a request payment

New tests cover these methods

### DIFF
--- a/packages/payment-processor/src/payment/encoder-approval.ts
+++ b/packages/payment-processor/src/payment/encoder-approval.ts
@@ -14,38 +14,190 @@ import {
 } from './swap-conversion-erc20';
 import { getPaymentNetworkExtension } from './utils';
 
-export function encodeRequestApprovalIfNeeded(
+/**
+ * For a given request and user, encode an approval transaction if it is needed.
+ * @param request the request
+ * @param provider generic provider
+ * @param from the user who will pay the request
+ * @param options specific to the request payment (conversion, swapping, ...)
+ */
+export async function encodeRequestErc20ApprovalIfNeeded(
   request: ClientTypes.IRequestData,
   provider: providers.Provider,
   from: string,
   options?: IRequestPaymentOptions,
 ): Promise<IPreparedTransaction | void> {
   if (options && options.swap) {
-    return encodeRequestApprovalWithSwapIfNeeded(request, provider, from, options);
+    return encodeRequestErc20ApprovalWithSwapIfNeeded(request, provider, from, options);
   } else {
-    return encodeRequestApprovalWithoutSwapIfNeeded(request, provider, from, options);
+    return encodeRequestErc20ApprovalWithoutSwapIfNeeded(request, provider, from, options);
   }
 }
 
-export async function encodeRequestApprovalWithoutSwapIfNeeded(
+/**
+ * For a given request, encode an approval transaction.
+ * @param request the request
+ * @param provider generic provider
+ * @param options specific to the request payment (conversion, ...)
+ */
+export function encodeRequestErc20Approval(
+  request: ClientTypes.IRequestData,
+  provider: providers.Provider,
+  options?: IRequestPaymentOptions,
+): IPreparedTransaction | void {
+  if (options && options.swap) {
+    return encodeRequestErc20ApprovalWithSwap(request, provider, options);
+  } else {
+    return encodeRequestErc20ApprovalWithoutSwap(request, provider, options);
+  }
+}
+
+/**
+ * For a given request and user, encode an approval transaction if it is needed when swap is not used.
+ * @param request the request
+ * @param provider generic provider
+ * @param from user who will pay the request
+ * @param options specific to the request payment (conversion, swapping, ...)
+ */
+export async function encodeRequestErc20ApprovalWithoutSwapIfNeeded(
   request: ClientTypes.IRequestData,
   provider: providers.Provider,
   from: string,
   options?: IRequestPaymentOptions,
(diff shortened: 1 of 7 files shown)


packages/payment-processor/src/payment/encoder-payment.ts
```diff
@@ -0,0 +1,109 @@
...
+  options?: IRequestPaymentOptions,
+): Promise<IPreparedTransaction> {
+  if (options && options.swap) {
+    return encodeRequestPaymentWithSwap(request, provider, options);
+  } else {
+    return encodeRequestPaymentWithoutSwap(request, options);
+  }
+}
+
+export async function encodeRequestPaymentWithoutSwap(
```
benjlevesque, line 28: this method (and the next) should not be async

packages/payment-processor/src/payment/encoder-approval.ts
```diff
@@ -0,0 +1,145 @@
...
+  from: string,
+  options: IRequestPaymentOptions,
+): Promise<IPreparedTransaction | void> {
+  const paymentNetwork = getPaymentNetworkExtension(request)?.id;
+
+  switch (paymentNetwork) {
+    case ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT:
+      if (options && options.swap) {
+        if (
+          !(await hasApprovalErc20ForSwapToPay(
```
benjlevesque, line 94: I think it's better to keep the check for approval outside of this. Actually we need a generic top level method to approve if needed. But the encoding should not be concerned by it. It should remain a simple, synchronous, method

packages/payment-processor/src/payment/index.ts
```diff
@@ -116,6 +120,37 @@ export async function payRequest(
...
 
+/**
+ * Encode the transactions associated to a request
+ * @param request the request to pay.
+ * @param signerOrProvider the Web3 provider, or signer. Defaults to window.ethereum.
+ * @param from The address which will send the transaction.
+ * @param options encoding options
+ * @returns
+ */
+export async function encodeRequest(
```
benjlevesque, line 131: see my other comment regarding the responsability of the encoding function

name is too generic

```suggestion
export function encodeRequestPayment(
```

benjlevesque (commented):
super nice! The encoding should actually be used in the `payRequest` method as well.

packages/payment-processor/src/payment/encoder-approval.ts
leoslr: Now there are the following methods (global, with and without swa):
- Approval encoding
- Check for knowing if approval is needed
- Approval encoding if needed

packages/payment-processor/src/payment/index.ts
leoslr: I modified this one to `encodeRequestApprovalAndPayment`. The specific methods for approval or payment encoding can be found in `encoder-approval` and `encoder-payment`.

leoslr:
@benjlevesque I agree that we could use the encoding method inside the `payRequest` method - this way we could also export a global payment method for both 'classic' and swap payment.
However the methods arguments of the encoding method cover a larger scope than the one from the payment method, so changing that would introduce breaking changes for invoicing. 

Maybe it can be covered in a separate PR, what do you think ?

packages/payment-processor/src/payment/encoder-payment.ts
```diff
@@ -0,0 +1,141 @@
...
+      return {
+        ...prepareEthFeeProxyPaymentTransaction(request, amount, feeAmount),
+        ...overrides,
+      };
+    default:
+      throw new Error('Payment network not found');
+  }
+}
+
+export async function encodeRequestPaymentWithSwap(
```
benjlevesque, line 101: remove async?

packages/payment-processor/src/payment/encoder-payment.ts
```diff
@@ -0,0 +1,141 @@
...
+  options?: IRequestPaymentOptions,
+): Promise<IPreparedTransaction> {
+  if (options && options.swap) {
+    return encodeRequestPaymentWithSwap(request, provider, options);
+  } else {
+    return encodeRequestPaymentWithoutSwap(request, options);
+  }
+}
+
+export async function encodeRequestPaymentWithoutSwap(
```
benjlevesque, line 28: remove async?

packages/payment-processor/test/payment/encoder.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    expect(confirmedTx.status).toBe(1);
+    expect(tx.hash).not.toBeUndefined();
+
+    tx = await wallet.sendTransaction(encodedTransactions[1]);
+    confirmedTx = await tx.wait(1);
+    expect(confirmedTx.status).toBe(1);
+    expect(tx.hash).not.toBeUndefined();
+  });
+
+  it.only('Should handle ERC20 Swap and Conversion Proxy request', async () => {
```
benjlevesque, line 290:
```suggestion
  it('Should handle ERC20 Swap and Conversion Proxy request', async () => {
```

benjlevesque (approved):

KolevDarko (approved):

PR #802, opened 2022-03-14
feat: reference calculator command

KolevDarko:
Added toolbox command for calculating payment reference.
Also at the same time fixed a typo in an escrow event name in payment-types.

### DIFF
--- a/packages/types/src/payment-types.ts
+++ b/packages/types/src/payment-types.ts
@@ -84,8 +84,8 @@ export enum EVENTS_NAMES {
 export enum ESCROW_EVENTS_NAMES {
   PAID_ESCROW = 'paidEscrow',
   PAID_ISSUER = 'paidIssuer',
-  INITIATED_EMERGENCY_CLAIM = 'initiateEmergencyClaim',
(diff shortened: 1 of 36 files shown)


packages/types/src/payment-types.ts
```diff
@@ -84,7 +84,7 @@ export enum EVENTS_NAMES {
 export enum ESCROW_EVENTS_NAMES {
   PAID_ESCROW = 'paidEscrow',
   PAID_ISSUER = 'paidIssuer',
-  INITIATED_EMERGENCY_CLAIM = 'initializeEmergencyClaim',
+  INITIATED_EMERGENCY_CLAIM = 'initiateEmergencyClaim',
   REVERTED_EMERGENCY_CLAIM = 'revertEmergencyClaim',
```
kevindavee, line 88: The enum key is past tense but the value is present tense. Is it intended this way?

kevindavee (approved):

packages/types/src/payment-types.ts
KolevDarko: you got a point, I made all three escrow specific events to be in present tense. the others should stay paid imo because we use paid in other places

leoslr (approved):

PR #805, opened 2022-03-15
feat: Batch payments erc20 eth

olivier7delf:
## Description of the changes
Add BatchPayment smart contract and tests, to do batch payments with ERC20 tokens and Ethers, with multiple recipients addresses, and possibly multiple tokens in a batch.

Main functions:
- batchEthPaymentsWithReference
- batchERC20PaymentsWithReference
- batchERC20PaymentsMultiTokensWithReference

The owner of the contract manages:
- feeBatch: between 0 and 1000, initialized at 0. 
                    ex: batchFee = 10 is equivalent to 1.0% of batch fee
- ERC20FeeProxy address
- EthereumFeeProxy address

### DIFF
--- a/packages/smart-contracts/src/contracts/BatchPayments.sol
+++ b/packages/smart-contracts/src/contracts/BatchPayments.sol
@@ -1,0 +1,274 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./interfaces/ERC20FeeProxy.sol";
+import "./interfaces/EthereumFeeProxy.sol";
+import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
+
+/**
+ * @title BatchPayments
+ * @notice This contract makes multiple payments with references, in one transaction:
+ *          - on: ERC20 Payment Proxy and ETH Payment Proxy of the Request Network protocol
+ *          - to: multiple addresses
+ *          - fees: Proxy fees can be paid to the same address. 
+ *                  An additional batch fee is added, paid to the same address.
+ */
+contract BatchPayments is Ownable, ReentrancyGuard {
+    
+    // This event is declared in ERC20FeeProxy
+    event TransferWithReferenceAndFee(
+        address tokenAddress,
+        address to,
+        uint256 amount,
+        bytes indexed paymentReference,
+        uint256 feeAmount,
+        address feeAddress
+    );
+    
+    // @dev: Between 0 and 1000, i.e: batchFee = 10 represent 1% of fee
+    uint256 public batchFee;
+
+    IERC20FeeProxy public paymentErc20FeeProxy;
+    IEthereumFeeProxy public paymentEthereumFeeProxy;
+
+    /**
+     * @param _paymentErc20FeeProxy The address to the ERC20 payment proxy to use.
+     * @param _paymentEthereumFeeProxy The address to the Ethereum payment proxy to use.
+     * @param _owner Owner of the contract.
+     */
+    constructor(address _paymentErc20FeeProxy, address _paymentEthereumFeeProxy, address _owner) {
+        paymentErc20FeeProxy  = IERC20FeeProxy(_paymentErc20FeeProxy);
+        paymentEthereumFeeProxy = IEthereumFeeProxy(_paymentEthereumFeeProxy);
+        batchFee = 0;
+        transferOwnership(_owner);
+    }
+
+    // Needed because batchEthPaymentsWithReferenceAndFee requires that the contract has ethers
+    receive() external payable {}
+
+    /**
+    * @notice Send a batch of Ethereum payments w/fees with paymentReferences to multiple accounts.
+    *         The sum of _amounts and _feeAmounts must be <= to msg.value.
+    *         If one payment failed, the whole batch is reverted
+    * @param _recipients List of recipients accounts.
+    * @param _amounts List of amounts, corresponding to recipients[].
+    * @param _paymentReferences List of paymentRefs, corr. to the recipients[].
+    * @param _feeAmounts List of amounts of the payment fee, corr. to the recipients[].
+    * @param _feeAddress The fee recipient.
+    * @dev Uses EthereumFeeProxy.sol to pay an invoice and fees, with a payment reference.
+    *      Make sure: msg.value >= sum(_amouts)+sum(_feeAmounts)
+    */
+    function batchEthPaymentsWithReference(
+        address[] calldata _recipients, 
+        uint256[] calldata _amounts,
+        bytes[] calldata _paymentReferences,
+        uint256[] calldata _feeAmounts,
+        address payable _feeAddress 
+    ) external payable nonReentrant {
+        require(
+            _recipients.length == _amounts.length
+            && _recipients.length == _paymentReferences.length
+            && _recipients.length == _feeAmounts.length
+            , "the input arrays must have the same length"
+        );
+
+        // Sender transfers tokens to the contract
+        payable(address(this)).transfer(msg.value);
+        uint256 toReturn = msg.value;
+        uint256 sumBatchFeeAmount = 0;
+
+        // Contract pays the batch payment
+        for (uint256 i = 0; i < _recipients.length; i++) {
+            uint256 batchFeeAmount = (_amounts[i] * batchFee) / 1000;
+            require(toReturn >= _amounts[i] + _feeAmounts[i] + batchFeeAmount, "not enough funds");
+            sumBatchFeeAmount += batchFeeAmount;
+            toReturn -= (_amounts[i]+_feeAmounts[i] + batchFeeAmount);
+
+            paymentEthereumFeeProxy.transferWithReferenceAndFee{value: _amounts[i]+_feeAmounts[i]}(
+                payable(_recipients[i]), 
+                _paymentReferences[i],
+                _feeAmounts[i],
+                payable(_feeAddress)
+            );
+        }
+        _feeAddress.transfer(sumBatchFeeAmount);
+
+        // Transfer the remaining ethers to the sender
+        (bool sendBackSuccess, ) = payable(msg.sender).call{ value: toReturn }('');
+        require(sendBackSuccess, 'Could not send remaining funds to the payer');
+    }
+
+    /**
+     * @notice Send a batch of erc20 payments w/fees with paymentReferences to multiple accounts.
+     * @param _tokenAddress Token to transact with.
+     * @param _recipients List of recipients accounts.
+     * @param _amounts List of amounts, corresponding to recipients[].
+     * @param _paymentReferences List of paymentRefs, corr. to the recipients[] and .
+     * @param _feeAmounts List of amounts of the payment fee, corr. to the recipients[].
+     * @param _feeAddress The fee recipient.
+     * @dev Uses ERC20FeeProxy.sol to pay an invoice and fees, with a payment reference.
+     *      Make sure the contract has allowance to spend the sender token.
+     */
+    function batchERC20PaymentsWithReference(
+        address _tokenAddress, 
+        address[] calldata _recipients, 
+        uint256[] calldata _amounts,
+        bytes[] calldata _paymentReferences,
+        uint256[] calldata _feeAmounts,
+        address _feeAddress 
+    ) external {
+        require(
+            _recipients.length == _amounts.length
+            && _recipients.length == _paymentReferences.length
+            && _recipients.length == _feeAmounts.length
+            , "the input arrays must have the same length"
+        );
+        
+        uint256 batchAmount = 0;
+
+        for (uint256 i = 0; i < _recipients.length; i++) {
+             batchAmount += _amounts[i];
+           (bool status, ) = address(paymentErc20FeeProxy).delegatecall(
+            abi.encodeWithSignature(
+            "transferFromWithReferenceAndFee(address,address,uint256,bytes,uint256,address)",
+                _tokenAddress,
+                _recipients[i], 
+                _amounts[i],
+                _paymentReferences[i],
+                _feeAmounts[i],
+                _feeAddress
+                )
+            );
+            require(status, "transferFromWithReference failed");
+        }
+
+        require(safeTransferFrom(_tokenAddress, _feeAddress, (batchAmount * batchFee) / 1000),
+         "batch fee transferFrom() failed");
+    }
+
+
+    /**
+     * @notice Send a batch of erc20 payments on multiple tokens w/fees with paymentReferences to multiple accounts.
+     * @param _tokenAddresses List of tokens to transact with.
+     * @param _recipients List of recipients accounts.
+     * @param _amounts List of amounts, corresponding to recipients[].
+     * @param _paymentReferences List of paymentRefs, corr. to the recipients[].
+     * @param _feeAmounts List of amounts of the payment fee, corr. to the recipients[].
+     * @param _feeAddress The fee recipient.
+     * @dev Uses ERC20FeeProxy.sol to pay an invoice and fees, with a payment reference.
+     *      Make sure the contract has allowance to spend the sender token.
+     */
+    function batchERC20PaymentsMultiTokensWithReference(
+        address[] calldata _tokenAddresses, 
+        address[] calldata _recipients, 
+        uint256[] calldata _amounts,
+        bytes[] calldata _paymentReferences,
+        uint256[] calldata _feeAmounts,
+        address _feeAddress 
+    ) external {
+        require(
+            _tokenAddresses.length == _recipients.length
+            && _tokenAddresses.length == _amounts.length
+            && _tokenAddresses.length == _paymentReferences.length
+            && _tokenAddresses.length == _feeAmounts.length
+            , "the input arrays must have the same length"
+        );
+
+        address[] memory uniqueTokens = new address[](_tokenAddresses.length);
+        uint256[] memory amountByToken = new uint256[](_tokenAddresses.length);
+
+        // Pay the requests
+        for (uint256 i = 0; i < _recipients.length; i++) {
+            // Create a list of unique tokens used
+            for(uint j = 0; j < uniqueTokens.length; j++){
+                if(uniqueTokens[j] == _tokenAddresses[i]){
+                    amountByToken[j] += _amounts[i];
+                    break;
+                }
+                if(amountByToken[j] == 0){
+                    uniqueTokens[j] = _tokenAddresses[i];
+                    amountByToken[j] = _amounts[i];
+                    break;
+                }
+            }
+
+           (bool status, ) = address(paymentErc20FeeProxy).delegatecall(
+            abi.encodeWithSignature(
+            "transferFromWithReferenceAndFee(address,address,uint256,bytes,uint256,address)",
+                _tokenAddresses[i],
+                _recipients[i], 
+                _amounts[i],
+                _paymentReferences[i],
+                _feeAmounts[i],
+                _feeAddress
+                )
+            );
+            require(status, "transferFromWithReference failed");
+        }
+
+        // Pay batch fee
+        for(uint i = 0; i < uniqueTokens.length && amountByToken[i] > 0; i++){
+            require(safeTransferFrom(uniqueTokens[i], _feeAddress, (amountByToken[i] * batchFee) / 1000),
+             "batch fee transferFrom() failed");
+             amountByToken[i] = 0;
+        }
+    }
+
+    /**
+    * @notice Call transferFrom ERC20 function and validates the return data of a ERC20 contract call.
+    * @dev This is necessary because of non-standard ERC20 tokens that don't have a return value.
+    * @return result The return value of the ERC20 call, returning true for non-standard tokens
+    */
+    function safeTransferFrom(address _tokenAddress, address _to, uint256 _amount) internal returns (bool result) {
+    /* solium-disable security/no-inline-assembly */
+    // check if the address is a contract
+    assembly {
+        if iszero(extcodesize(_tokenAddress)) { revert(0, 0) }
+    }
+
+    // solium-disable-next-line security/no-low-level-calls
+    (bool success, ) = _tokenAddress.call(abi.encodeWithSignature(
+        "transferFrom(address,address,uint256)",
+        msg.sender,
+        _to,
+        _amount
+    ));
+
+    assembly {
+        switch returndatasize()
+        case 0 { // Not a standard erc20
+            result := 1
+        }
+        case 32 { // Standard erc20
+            returndatacopy(0, 0, 32)
+            result := mload(0)
+        }
+        default { // Anything else, should revert for safety
+            revert(0, 0)
+        }
+    }
+
+    require(success, "transferFrom() has been reverted");
+
+    /* solium-enable security/no-inline-assembly */
+    return result;
+  }
+
+    /*
+    * Admin functions to edit the proxies address
+    */
+
+    function setBatchFee(uint256 _batchFee) public onlyOwner {
+        batchFee = _batchFee;
+    }
+
+    function setPaymentErc20FeeProxy(address _paymentProxyAddress) public onlyOwner {
+        paymentErc20FeeProxy = IERC20FeeProxy(_paymentProxyAddress);
+    }
+
+    function setPaymentEthereumFeeProxy(address _paymentEthereumFeeProxyAddress) public onlyOwner {
+        paymentEthereumFeeProxy = IEthereumFeeProxy(_paymentEthereumFeeProxyAddress);
+    }
+}
--- a/packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
+++ b/packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
@@ -1,0 +1,558 @@
+import { ethers, network } from 'hardhat';
+import { BigNumber, Signer } from 'ethers';
+import { expect, use } from 'chai';
+import { solidity } from 'ethereum-waffle';
+import { TestERC20__factory, TestERC20, BatchPayments } from '../../src/types';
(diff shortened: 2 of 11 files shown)


packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+          },
+        ),
+    )
+      .to.emit(ethFeeProxy, 'TransferWithReferenceAndFee')
+      .withArgs(payeeOne, '20', ethers.utils.keccak256(referenceExample1), '1', feeAddress)
+      .to.emit(ethFeeProxy, 'TransferWithReferenceAndFee')
+      .withArgs(payeeTwo, '30', ethers.utils.keccak256(referenceExample2), '2', feeAddress);
+
+    afterEthBalance = await provider.getBalance(payeeOne);
+    expect(afterEthBalance).to.be.equal(beforeEthBalance.add(20));
```
benjlevesque, line 91: payeeTwo should also be verified.

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    expect(await provider.getBalance(batchAddress)).to.be.equal(0);
+  });
```
benjlevesque, line 93: the test says "and return the excess of ethers": this is not tested

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    afterEthBalance = await provider.getBalance(payeeOne);
+    expect(afterEthBalance).to.be.equal(beforeEthBalance.add(20));
+
+    afterEthBalance1 = await provider.getBalance(payeeTwo);
+    expect(afterEthBalance1).to.be.equal(beforeEthBalance1.add(30));
+  });
+
+  it('Gas evaluation: Should pay multiple Ethereum payments', async function () {
+    let listTxs = [1, 12]; // [1, 4, 8, 12, 30, 100]; //
```
benjlevesque, line 122: I cannot figure out what this is? could you rename or comment?

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    await tx.wait();
+
+    afterEthBalance = await provider.getBalance(payeeOne);
+    expect(afterEthBalance).to.be.equal(beforeEthBalance.add(20));
+
+    afterEthBalance1 = await provider.getBalance(payeeTwo);
+    expect(afterEthBalance1).to.be.equal(beforeEthBalance1.add(30));
+  });
+
+  it('Gas evaluation: Should pay multiple Ethereum payments', async function () {
```
benjlevesque, line 121: this test seems to be about gas, but there is nothing in it specific to gas 🤔 (except a log)

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    afterEthBalance = await provider.getBalance(payeeOne);
+    expect(afterEthBalance).to.be.equal(beforeEthBalance);
+
+    afterEthBalance1 = await provider.getBalance(payeeTwo);
+    expect(afterEthBalance1).to.be.equal(beforeEthBalance1);
+  });
+});
+
+// Allow to create easly BatchPayments input, especially for gas optimization. tokenAddress is unused for batch Ethereum
+const getInputsBatchPayments = function (
```
benjlevesque, line 188:
```suggestion
const getBatchPaymentsInputs = function (
```

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+            "transferFromWithReferenceAndFee(address,address,uint256,bytes,uint256,address)",
+                _tokenAddresses[i],
+                _recipients[i], 
+                _amounts[i],
+                _paymentReferences[i],
+                _feeAmounts[i],
+                _feeAddress
+                )
+            );
+            require(status, "transferFromWithReference failed");
```
benjlevesque, line 163: does this mean one failure will fail the whole batch transaction? will it revert successful ones?

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+        uint256[] calldata _amounts
+    ) external {
+        for (uint256 i = 0; i < _recipients.length; i++)
+            _token.transferFrom(msg.sender, _recipients[i], _amounts[i]);
+    }
+
+    
+    /**
+    * @notice Send a batch of Ethereum payments w/fees with paymentReferences to multiple accounts.
+    *         The sum of _amounts and _feeAMounts must be <= to msg.value.
```
benjlevesque, line 63:
```suggestion
    *         The sum of _amounts and _feeAmounts must be <= to msg.value.
```

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+    receive() external payable {}
+
+    /**
+     * @notice Send a batch of erc20 payments, to multiple accounts.
+     * @param _token The token to transact with.
+     * @param _recipients Takes a list of accounts.
+     * @param _amounts Takes a list of amounts, corresponding to recipients[].
+     * @dev Remember to give approval for this contract to spend user tokens.
+     */
+    function batchOrphanERC20Payments(
```
benjlevesque, line 51: why Orphan?

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,400 @@
...
+    await token.connect(owner).transfer(spender2Address, 160);
+    await token.connect(spender).approve(batchAddress, 1000);
+    await token.connect(spender2).approve(batchAddress, 170);
+
+    // 2nd token
+    await token1.connect(owner).transfer(spender2Address, 1000);
+    await token1.connect(spender2).approve(batchAddress, 1000);
+  });
+
+  it('Should execute a batch payments of a ERC20 to four accounts', async function () {
```
benjlevesque, line 80: why do we need this method if there is no payment reference?

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+     * @notice Send a batch of erc20 payments w/fees with paymentReferences to multiple accounts.
+     * @param _tokenAddress Token to transact with.
+     * @param _recipients List of recipients accounts.
+     * @param _amounts List of amounts, corresponding to recipients[].
+     * @param _paymentReferences List of paymentRefs, corr. to the recipients[] and .
+     * @param _feeAmounts List of amounts of the payment fee, corr. to the recipients[].
+     * @param _feeAddress The fee recipient.
+     * @dev Uses ERC20FeeProxy.sol to pay an invoice and fees, with a payment reference.
+     */
+    function batchERC20PaymentsWithReferenceAndFee(
```
benjlevesque, line 109: what's the reason for this method? does it reduce the gas usage?

benjlevesque (commented):
huge! I have a ton of questions and comments because i'm not super familiar with the topic, my bad if they aren't relevant

packages/smart-contracts/package.json
```diff
@@ -47,6 +47,7 @@
     "lint:check": "yarn run lint:lib:check && yarn run lint:sol:check",
     "ganache": "ganache-cli -l 90000000 -p 8545 -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\"",
     "deploy": "yarn hardhat deploy-local-env --network private",
+    "deploy-batch-payments": "yarn hardhat deploy-batch-payments --network rinkeby",
```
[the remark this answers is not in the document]
olivier7delf: That's right, I can delete it.
-> I added it because I used it a lot to test on rinkeby, and it would be useful when I will work again on the contract, e.g: add batch conversion. And also when we'll update contracts for gas optim.

packages/smart-contracts/src/contracts/BatchPayments.sol
olivier7delf: I didn't challenged the used of this legacy function, thought it was to offer the choice for other dev.
@yomarion ?

benjlevesque: there is no legacy here, this is a new contract. if you don't know why we need it, I say remove it ;)

packages/smart-contracts/src/lib/artifacts/ERC20FeeProxy/index.ts
```diff
@@ -18,8 +18,8 @@ export const erc20FeeProxyArtifact = new ContractArtifact<ERC20FeeProxy>(
           creationBlockNumber: 10774767,
         },
         rinkeby: {
-          address: '0xda46309973bFfDdD5a10cE12c44d2EE266f45A44',
-          creationBlockNumber: 7118080,
+          address: '0xf0AaAE5d39Df507608709B3f5Ed3373dD22E24e6',
```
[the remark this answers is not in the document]
olivier7delf: No you are right, I am fixing all arteficat issues

packages/smart-contracts/src/contracts/interfaces/ERC20FeeProxy.sol
```diff
@@ -3,13 +3,13 @@ pragma solidity ^0.8.0;
...
 interface IERC20FeeProxy {
   event TransferWithReferenceAndFee(
-    address tokenAddress,
-    address to,
-    uint256 amount,
-    bytes indexed paymentReference,
-    uint256 feeAmount,
-    address feeAddress
-  );
+        address _tokenAddress,
```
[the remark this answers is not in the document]
olivier7delf: corrected

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * @param _paymentErc20FeeProxy The address to the ERC20 payment proxy to use.
+     * @param _paymentEthereumFeeProxy The address to the Ethereum payment proxy to use.
+     */
+    constructor(address _paymentErc20FeeProxy, address _paymentEthereumFeeProxy) {
+        paymentErc20FeeProxy  = IERC20FeeProxy(_paymentErc20FeeProxy);
+        paymentEthereumFeeProxy = IEthereumFeeProxy(_paymentEthereumFeeProxy);
+    }
+
+    // Needed because batchEthereumPaymentsWithReferenceAndFee requires that the contract has ethers
```
yomarion, line 41: TODO for me to double-check

packages/smart-contracts/src/contracts/BatchPayments.sol
yomarion: Oh yes, you can remove this

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+    *         The sum of _amounts and _feeAMounts must be <= to msg.value.
+    *         If one payment failed, the whole batch is reverted
+    * @param _recipients List of recipients accounts.
+    * @param _amounts List of amounts, corresponding to recipients[].
+    * @param _paymentReferences List of paymentRefs, corr. to the recipients[].
+    * @param _feeAmounts List of amounts of the payment fee, corr. to the recipients[].
+    * @param _feeAddress The fee recipient.
+    * @dev Uses EthereumFeeProxy.sol to pay an invoice and fees, with a payment reference.
+     */
+    function batchEthereumPaymentsWithReferenceAndFee(
```
yomarion, line 72:
```suggestion
    function batchEthPaymentsWithReferences(
```
New naming convention, "WithFee" is by default.

packages/smart-contracts/src/contracts/BatchPayments.sol
yomarion: We need to be sure of that choice.

What would be good for him about the "All or nothing" ?

One advantage of this approach is the gas estimation should fail if one of the payments is doomed to fail, which is better for UX in case of a bug.

One drawback is that if the failure is not predictable (eg the ETH balance got drained in a previous tx from the same block), the user may find it handy to have paid "as much as possible"

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,178 @@
...
+        address[] calldata _recipients, 
+        uint256[] calldata _amounts,
+        bytes[] calldata _paymentReferences,
+        uint256[] calldata _feeAmounts,
+        address _feeAddress 
+    ) external payable nonReentrant {
+        // EOA transfer token on the contract
```
yomarion, line 79: Not only EOA, right?

packages/smart-contracts/src/contracts/interfaces/ERC20FeeProxy.sol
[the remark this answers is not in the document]
yomarion: We don't prepend with `_` in events

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+          [payeeOne, payeeTwo],
+          [20, 30],
+          [referenceExample1, referenceExample2],
+          [1, 2],
+          feeAddress,
+          {
+            value: totalAmout,
+          },
+        ),
+    ).to.be.reverted;
```
yomarion, line 177: Please precise revert messages. I saw in the contract we were not requiring that value >= fees + amounts, which might make sense but will it be clear for users / debug ?

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    afterEthBalance = await provider.getBalance(payeeOne);
+    expect(afterEthBalance).to.be.equal(beforeEthBalance);
+
+    afterEthBalance1 = await provider.getBalance(payeeTwo);
+    expect(afterEthBalance1).to.be.equal(beforeEthBalance1);
```
yomarion, line 183: I would also test that the wallet's ETH balance is 0, and it might apply to other tests BTW

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      }
+
+      afterEthBalance = await provider.getBalance(payeeTwo);
+      expect(afterEthBalance).to.be.equal(beforeEthBalance.add(amount * nbTxs));
+    }
+  });
+
+  it('Should revert and any payment is achieved because funds are not sufficient', async function () {
+    beforeEthBalance = await provider.getBalance(payeeOne);
+    beforeEthBalance1 = await provider.getBalance(payeeTwo);
```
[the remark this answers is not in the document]
yomarion: Furthermore, choose between "1" + "2" and "One" + "Two" please

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    afterERC20Balance = await token.balanceOf(payeeTwo);
+    expect(afterERC20Balance).to.be.equal(beforeERC20Balance.add(amount * 2));
+
+    afterERC20Balance1 = await token1.balanceOf(payeeTwo);
+    expect(afterERC20Balance1).to.be.equal(beforeERC20Balance1.add(amount * 2));
+  });
+
+  it('GAS EVALUATION: Should pay multiple ERC20 payments', async function () {
+    let listTxs = [1, 12]; // [1, 4, 8, 12, 30, 100];
```
yomarion, line 215: Commented code FYI

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,400 @@
...
+        console.log(`nbTxs= ${nbTxs}, gas consumption: `, (await receipt).gasUsed.toString());
+      }
+
+      afterERC20Balance = await token.balanceOf(payeeThree);
+      expect(afterERC20Balance).to.be.equal(beforeERC20Balance.add(amount * nbTxs));
+    }
+  });
+
+  it('GAS EVALUATION: Should pay multiple ERC20 payments on multiple tokens', async function () {
+    let listTxs = [1, 12]; // [1, 4, 8, 12, 30, 100];
```
yomarion, line 256: Commented code FYI

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,400 @@
...
+    await tx.wait();
+
+    afterERC20Balance = await token.balanceOf(payeeTwo);
+    expect(afterERC20Balance).to.be.equal(beforeERC20Balance.add(amount * 2));
+
+    afterERC20Balance1 = await token1.balanceOf(payeeTwo);
+    expect(afterERC20Balance1).to.be.equal(beforeERC20Balance1.add(amount * 2));
+  });
+
+  it('GAS EVALUATION: Should pay multiple ERC20 payments', async function () {
```
yomarion, line 214: For the sake of a readable code, please nest the gas evaluation related unit tests within a describe section

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,400 @@
...
+          [20, 30, 40, 50],
+        ),
+    )
+      .to.emit(token, 'Transfer')
+      .to.emit(token, 'Transfer')
+      .to.emit(token, 'Transfer')
+      .to.emit(token, 'Transfer');
+
+    afterERC20Balance = await token.connect(owner).balanceOf(payeeThree);
+    expect(afterERC20Balance).to.be.equal(beforeERC20Balance.add(40));
```
yomarion, line 176: This applies for many of these tests but we should test all the balances.

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
```diff
@@ -0,0 +1,210 @@
...
+import { batchPaymentsArtifact } from '../../src/lib';
+
+import { ethereumFeeProxyArtifact } from '../../src/lib/';
+import { HttpNetworkConfig } from 'hardhat/types';
+
+use(solidity);
+
+const logInfos = false;
+
+describe('contract: BatchPayments: Ethereum', () => {
```
yomarion, line 15: What should the behaviour be if the arrays given to the function are not of the same length? Shouldn't we test that to control the error messages and ease maintenance? (same applies for Batch ERC20 batch tests + contract)

yomarion (commented):
Well done, it looks solid and will bring a lot of value. It's a big topic so I had a lot of comments, we can discuss that. We should support additional fee per batch starting on the first version.

yomarion (commented):
Well done, it looks solid and will bring a lot of value. It's a big topic so I had a lot of comments, we can discuss that. We should support additional fee per batch starting on the first version.

packages/smart-contracts/src/contracts/BatchPayments.sol
olivier7delf: I agree, and I need to take a decision early because it can modify quite a lot the code:

"nothing": 
pros ->
- if X txs failed, customer has "just" to send again the amount these X txs, even without batch
- if there is a failed, gas already used is not wasted
cons ->
- right now, batchEthereum revert all, and will need an important change to avoid it. (because it cannot use delegatecall with msg.valueSpecificForEachTx, so I have to use an interface of feeProxy, -> the transaction reverted will revert the batch directly. I only know one way to use a delegatecall but it will use a lot of gas)

"all":
pros ->
- "gas estimation should fail if one of the payments is doomed to fail" Yoann
cons:
- "if the failure is not predictable" Yoann, a lot of customer's gas can be wasted

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
olivier7delf: I xdeleted the loop...

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
olivier7delf: I have added a notice to the function at @dev to take care about the sum...
I have also added a check in the loop.

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
olivier7delf: deleted because no more orphan payment in the contract.

packages/smart-contracts/src/contracts/BatchPayments.sol
olivier7delf: Kept it for gas economy.

packages/smart-contracts/src/contracts/BatchPayments.sol
olivier7delf: done

packages/smart-contracts/src/contracts/BatchPayments.sol
olivier7delf: updated

packages/smart-contracts/test/contracts/BatchEthereumPayments.test.ts
olivier7delf: added.

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,273 @@
...
+import "./interfaces/ERC20FeeProxy.sol";
+import "./interfaces/EthereumFeeProxy.sol";
+import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
+
+/**
+ * @title BatchPayments
+ * @notice This contract pays multiple payment requests in one transaction, with batch fee:
+ *          - on: ERC20 and Ethereum
+ *              - regarding ERC20: possibly on multiple tokens in a batch
+ *          - to: multiple addresses
```
yomarion, line 15:
```suggestion
 * @notice This contract makes multiple payments with references, in one transaction:
 *          - on: ERC20 Payment Proxy and ETH Payment Proxy of the Request Network protocol
 *          - to: multiple addresses
 *          - fees: Proxy fees can be paid to the same address. An additional batch fee is added, paid to the same address.
```

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,273 @@
...
+ */
+contract BatchPayments is Ownable, ReentrancyGuard {
+    
+    // Event to declare a transfer with a reference
```
yomarion, line 19:
```suggestion
    // This event is declared in ERC20FeeProxy
```

packages/smart-contracts/src/contracts/BatchPayments.sol
```diff
@@ -0,0 +1,273 @@
...
+        address payable _feeAddress 
+    ) external payable nonReentrant {
+        require(
+            _recipients.length == _amounts.length
+            && _recipients.length == _paymentReferences.length
+            && _recipients.length == _feeAmounts.length
+            , "the input arrays must have the same length"
+        );
+
+        // Sender transfer token on the contract
```
yomarion, line 76:
```suggestion
        // Sender transfers tokens to the contract
```

packages/smart-contracts/test/contracts/BatchEthPayments.test.ts
```diff
@@ -0,0 +1,321 @@
...
+      console.log(`nbTxs= ${nbTxs}, gas consumption: `, receipt.gasUsed.toString());
+    }
+
+    afterEthBalance2 = await provider.getBalance(payee2);
+    expect(afterEthBalance2).to.be.equal(beforeEthBalance2.add(amount * nbTxs));
+
+    expect(await provider.getBalance(batchAddress)).to.be.equal(0);
+  });
+
+  it('Should revert batch if not enough funds', async function () {
```
yomarion, line 148: It's wise to nest the former test cases under a describe "happy path" for example, and starting from this test case, nest under a describe "expected reverts".

packages/smart-contracts/test/contracts/BatchEthPayments.test.ts
```diff
@@ -0,0 +1,321 @@
...
+          [referenceExample1],
+          [1, 2],
+          feeAddress,
+        ),
+    ).revertedWith('the input arrays must have the same length');
+
+    expect(await provider.getBalance(batchAddress)).to.be.equal(0);
+  });
+
+  it('Should allow only the owner to update batchFee', async function () {
```
yomarion, line 260: Admin tasks should also be nested in a specific `describe`

packages/smart-contracts/test/contracts/BatchErc20Payments.test.ts
```diff
@@ -0,0 +1,520 @@
...
+
+    // 3nd token
+    await token3.connect(owner).transfer(spender3Address, 100);
+    await token3.connect(spender3).approve(batchAddress, 100);
+
+    // set batch fee at 100 (=10%) for the purpose of the tests.
+    await batch.connect(owner).setBatchFee(100);
+  });
+
+  it('Should pay 3 ERC20 payments with paymentRef and pay batch fee', async function () {
```
yomarion, line 83: Same nesting feedback as for the other test file

packages/smart-contracts/src/contracts/BatchPayments.sol
yomarion: Or better actually if you declare the event the same way you did for ETH Proxy

yomarion (approved):
Nice

bertux (approved):

PR #809, opened 2022-03-16
feat: exporting calcEscrowState function from payment-detection

KolevDarko:
Function for parsing through escrow events and computing the final escrow state

### DIFF
--- a/packages/payment-detection/src/utils.ts
+++ b/packages/payment-detection/src/utils.ts
@@ -106,7 +106,7 @@ export const hashReference = (paymentReference: string): string => {
  * @param escrowEvents Balance of the request being updated
  * @returns
  */
-export const calcEscrowState = (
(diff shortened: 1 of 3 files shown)


packages/payment-detection/src/utils.ts
```diff
@@ -100,3 +100,29 @@ export const makeGetDeploymentInformation = <
 export const hashReference = (paymentReference: string): string => {
   return keccak256(`0x${paymentReference}`);
 };
+
+/**
+ * Returns escrow status based on array of escrow events
+ * @param escrowEvents Balance of the request being updated
+ * @returns
+ */
+export const calcEscrowState = (
```
benjlevesque, line 109: avoid abbreviations for an exported function

benjlevesque (approved):

PR #827, opened 2022-04-07
fix: removed escrow graph info retriever test

KolevDarko:
Removed escrow thegraph info retriever test because this one was depending on rinkeby onchain data and easily breaks when the data changes or the contract address changes.


leoslr (approved):

bertux (approved):
Please do not forget to add it back in a mocked way before production release.

PR #832, opened 2022-04-18
feat: improve currency typing

benjlevesque:
This change allows 2 things:
- using more specific types, like `ERC20Currency`:
```ts
const method = (c: ERC20Currency) => c;
if (curr.type === 'ERC20') {
  method(curr);
}
```

- simplify the use of `network`:
if(curr.type === 'ISO4217'){
  // type: undefined
  curr.network
} else{
  // type: string
  curr.network
}


packages/currency/src/types.ts
```diff
@@ -43,22 +48,28 @@ export type ERC777Currency = {
...
-  | ({ type: RequestLogicTypes.CURRENCY.ERC20 } & ERC20Currency)
-  | ({ type: RequestLogicTypes.CURRENCY.ERC777 } & ERC777Currency);
+export type CurrencyInput = NativeCurrency | ISO4217Currency | ERC20Currency | ERC777Currency;
 
 /**
  * The description of Currency, its core properties and some computed properties.
  * `meta` enables applications to add any metadata they need to a Currency
  */
-export type CurrencyDefinition<TMeta = unknown> = CurrencyInput & {
+export type FromInput<T, TMeta> = T & {
```
yomarion, line 57: I don't really understand the naming `FromInput`, what input is it? Isn't the comment outdated? Is the export still required?

packages/currency/src/types.ts
```diff
@@ -1,29 +1,33 @@
 import { RequestLogicTypes } from '@requestnetwork/types';
 
+/** Native Currency types */
+export type NativeCurrencyType = RequestLogicTypes.CURRENCY.BTC | RequestLogicTypes.CURRENCY.ETH;
```
yomarion, line 4: Not 100% sure but: isn't it a good time to remove either the distinction between the 2 or at least to replace the ETH part with a more generic `NATIVE_TOKEN` ? Now that we have network, it makes sense, IMO. We will never have a CURRENCY.ETH on `network: bitcoin` or reversely and I'm not sure it ever helps to know that `BTC-bitcoin` is very close to `BTC-bitcoin-testnet`

yomarion (commented):
Very clean move but I wonder about the different values `NativeCurrencyType`, shouldn't we merge them?

PR #835, opened 2022-04-22
fix: near address validation

leoslr:
Validate address for both of these format:
- `account.near` 
- 32 bytes address

Previously it only validated the `account.near` format


KolevDarko (approved):

bertux (approved):
good fix

alexandre-abrioux (changes requested):
The regex should not be changed IMO as this is the regex found in the specs.
I think you want to change the switch...case and make the prefixes  (`.near `or `.testnet`) optional, like so:
```ts
switch (network) {
    case 'aurora':
      return !!address.match(/[^.]+(\.near)?$/);
    case 'aurora-testnet':
      return !!address.match(/[^.]+(\.testnet)?$/);
    default:
      throw new Error(`Cannot validate NEAR address for network ${network}`);
  }
```
Add tests for this to be sure 😊

alexandre-abrioux (approved):
👌

PR #836, opened 2022-04-28
fix-near-solvency

leoslr:
NATIVE_TOKEN was not registered as a supported network, leading the solvency check to fail when using near.

### DIFF
--- a/packages/payment-processor/src/payment/index.ts
+++ b/packages/payment-processor/src/payment/index.ts
@@ -27,6 +27,7 @@ export const supportedNetworks = [
   ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_PROXY_CONTRACT,
   ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT,
   ExtensionTypes.ID.PAYMENT_NETWORK_ETH_INPUT_DATA,
+  ExtensionTypes.ID.PAYMENT_NETWORK_NATIVE_TOKEN,
 ];
 
 export interface IConversionPaymentSettings {


alexandre-abrioux (approved):

packages/payment-processor/src/payment/index.ts
```diff
@@ -27,6 +27,7 @@ export const supportedNetworks = [
   ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_PROXY_CONTRACT,
   ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT,
   ExtensionTypes.ID.PAYMENT_NETWORK_ETH_INPUT_DATA,
+  ExtensionTypes.ID.PAYMENT_NETWORK_NATIVE_TOKEN,
```
alexandre-abrioux, line 30: Should we remove `PAYMENT_NETWORK_ETH_INPUT_DATA` ?
```suggestion
  ExtensionTypes.ID.PAYMENT_NETWORK_NATIVE_TOKEN,
```

leoslr:
@alexandre-abrioux I think it can stay for now. I was just trying to solve an issue for near, But I'am not sure why we have this variable here in the first place as it is only used in this file and lacks some of the payment network. I'll look more into later

alexandre-abrioux (approved):

olivier7delf (approved):

packages/payment-processor/src/payment/index.ts
olivier7delf: Invoicing use also ETH_INPUT_DATA, it might require more refacto, no?
(in payment.tsx)

PR #837, opened 2022-04-29
chore(request-node): properly catch IPFS timeout exceptions

alexandre-abrioux:
## Description of the changes

This PR makes sure the request-node catches IPFS timeout exceptions. It also removes support for NodeJS < 16.

## Issue

The ipfs-manager module uses the low-level NodeJS HTTP API to talk to IPFS.

The timeout setting is set via a call to [request.setTimeout](https://nodejs.org/api/http.html#requestsettimeouttimeout-callback):
https://github.com/RequestNetwork/requestNetwork/blob/265421b4cdf2fda82285ae87ce0af78f306e77b3/packages/ethereum-storage/src/ipfs-manager.ts#L105-L118

Depending on the NodeJS version, the timeout has to be set only once the socket is open. So we should either:

- listen for the socket "open" event, and then set the timeout
- or deprecate old NodeJS versions and set it directly via the new "timeout" option at request creation
- or, even better, stop using the low level API and rewrite the module with axios (what I did in the end)


packages/ethereum-storage/src/ipfs-manager.ts
```diff
@@ -109,16 +109,14 @@ export default class IpfsManager {
...
         port: this.ipfsConnection.port,
+        timeout: this.ipfsConnection.timeout > 0 ? this.ipfsConnection.timeout : undefined,
       });
+
       // Sending the request
       addForm.pipe(addRequest);
 
-      if (this.ipfsConnection.timeout && this.ipfsConnection.timeout > 0) {
-        addRequest.setTimeout(this.ipfsConnection.timeout);
-      }
```
alexandre-abrioux, line 118: This is the fix. The timeout can only be set once the socket is open. So 2 solutions:
- either listen for the socket "open" event, and then set the timeout
- or set it directly via the new "timeout" option, directly at request creation (what I did here)

packages/ethereum-storage/test/ipfs-manager.test.ts
```diff
@@ -379,4 +379,17 @@ describe('Ipfs manager', () => {
...
+    const assertionsNb = 10;
+    expect.assertions(assertionsNb);
+    let assertions = [];
+    for (let i = 0; i < assertionsNb; i++) {
+      assertions.push(
+        expect(ipfsManager.add('test')).rejects.toThrowError('Ipfs add request timeout'),
+      );
+    }
+    await Promise.all(assertions);
+  });
```
alexandre-abrioux, line 394: This test was failing before (at least once in the loop of 10 assertions). It is green now.

packages/request-node/test/thegraph/persistTransaction.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      assertions.push(
+        request(server)
+          .post('/persistTransaction')
+          .send({ channelId, topics, transactionData })
+          .set('Accept', 'application/json')
+          .expect(StatusCodes.INTERNAL_SERVER_ERROR),
+      );
+    }
+    await Promise.all(assertions);
+  });
```
alexandre-abrioux, line 66: This test was failing before the change and is now green.

packages/ethereum-storage/test/ipfs-manager.test.ts
```diff
@@ -52,40 +45,43 @@ const contentLengthOnIpfs2 = 38;
...
 
   it('allows to pin one file ipfs', async () => {
+    await ipfsManager.add(content);
     const pinnedHash = await ipfsManager.pin([hash]);
     expect(hash).toBe(pinnedHash[0]);
   });
 
   it('allows to pin multiple files to ipfs', async () => {
+    await ipfsManager.add(content);
+    await ipfsManager.add(content2);
```
alexandre-abrioux, line 84: Those added lines are only to make tests independent of each other (to be able to run those tests alone)

packages/request-node/test/thegraph/persistTransaction.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    const assertions = [];
+    for (let i = 0; i < assertionsNb; i++) {
+      assertions.push(
+        request(server)
+          .post('/persistTransaction')
+          .send({ channelId, topics, transactionData })
+          .set('Accept', 'application/json')
+          .expect(StatusCodes.INTERNAL_SERVER_ERROR),
+      );
+    }
```
alexandre-abrioux, line 58: This test was failing before the change and is now green.

packages/ethereum-storage/src/ipfs-manager.ts
```diff
@@ -298,230 +129,58 @@ export default class IpfsManager {
...
-            return reject(Error(`Ipfs pin request error: ${e}`));
-          }
-        });
-
-      if (timeout || (this.ipfsConnection.timeout && this.ipfsConnection.timeout > 0)) {
-        getRequest.setTimeout(timeout || this.ipfsConnection.timeout);
+  public async pin(hashes: string[], timeout?: number): Promise<string[]> {
+    try {
+      const response = await this.axiosInstance.get(
+        `${this.IPFS_API_PIN}?arg=${hashes.join('&arg=')}`,
```
benjlevesque, line 135: suggestion: use axios' `params` field instead of `?arg=`

you might need to override the paramsSerializer in the axiosInstance creation

```
// `paramsSerializer` is an optional function in charge of serializing `params`
// (e.g. https://www.npmjs.com/package/qs, http://api.jquery.com/jquery.param/)
  paramsSerializer: function (params) {
    return Qs.stringify(params, {arrayFormat: 'brackets'})
  },
 ```

File: packages/ethereum-storage/src/ipfs-manager.ts [hunk 7]
```diff
@@ -192,104 +95,32 @@ export default class IpfsManager {
...
-          }
-        });
-
-      if (this.ipfsConnection.timeout && this.ipfsConnection.timeout > 0) {
-        getRequest.setTimeout(this.ipfsConnection.timeout);
+    try {
+      if (maxSize !== Number.POSITIVE_INFINITY) {
+        const maxSizeBase64 = ((4 * maxSize) / 3 + 3) & ~3;
+        const jsonMinSize = 500;
+        maxSize = Math.max(maxSizeBase64, jsonMinSize);
```
benjlevesque, line 106: this needs to be documented I think, not trivial 😅

benjlevesque (approved):
great refacto!

yomarion (approved):
Well done

packages/ethereum-storage/src/ipfs-manager.ts
alexandre-abrioux: done in https://github.com/RequestNetwork/requestNetwork/pull/837/commits/26f995a9be2391027a940208ef6b0a1091a08cd2

packages/ethereum-storage/src/ipfs-manager.ts
alexandre-abrioux: done in https://github.com/RequestNetwork/requestNetwork/pull/837/commits/26f995a9be2391027a940208ef6b0a1091a08cd2

alexandre-abrioux:
@yomarion or @benjlevesque Could you change the CI expected test in the repository settings: please rename the expected `test-graph-node` test to `test-integration-graph-node`. It was renamed like this since we now have new tests for TheGraphRequestNode in the `request-node` package, which were confusing with the integration tests name.

PR #843, opened 2022-05-05
feat: reject approved request

KolevDarko:
## Description of the changes
By requests from users we are now allowing users to be able to cancel (as payer) or void (as payee) an already accepted request.


packages/request-logic/src/config.ts
```diff
@@ -1,6 +1,6 @@
 export default {
   specificationVersion: {
-    current: '2.0.3',
+    current: '2.0.4',
```
yomarion, line 3:
```suggestion
    current: '2.1.0',
```
😉

packages/request-logic/test/unit/actions/cancel.test.ts
```diff
@@ -109,7 +109,7 @@ describe('actions/cancel', () => {
         timestamp: 2,
       });
     });
-    it('cannot cancel by payer with state === accepted', async () => {
+    it('can cancel by payer with state === accepted', async () => {
```
yomarion, line 112: Can you keep the original test with a request v2.0.3 and keep the test that it still fail?

yomarion (changes requested):
It might not be very simple, but we actually need to implement the retrocompatibility for 2.0.3 that should not accept the cancellation of an approved request.

packages/request-logic/test/unit/actions/cancel.test.ts
KolevDarko: done

packages/request-client.js/test/etherscan-mock.ts
```diff
@@ -41,7 +41,7 @@ export default class EtherscanProviderMock {
             to: '0xc12F17Da12cd01a9CDBB216949BA0b41A6Ffc4EB',
             value: BigNumber.from('0x02dd231b00'),
             nonce: 205,
-            data: '0xefce79375b2db9f7',
+            data: 'a0ebc16890a6483f',
```
alexandre-abrioux, line 44: Should it start by `0x` or did you change this on purpose?

packages/request-logic/specs/request-logic-specification.md
```diff
@@ -354,7 +354,7 @@ This action is valid, if:
 
 - the request **has a payer**
 - the Role of the action **signer is the payer**
-- the request state is `created`
+- the request state is **NOT** `canceled`
```
alexandre-abrioux, line 357: Do we have access to the balance? Can we check that it hasn't been paid?

alexandre-abrioux (approved):

packages/request-logic/src/actions/cancel.ts
```diff
@@ -60,18 +61,21 @@ function applyActionToRequest(
...
+      if (request.state !== RequestLogicTypes.STATE.CREATED) {
+        throw new Error('A payer cancel need to be done on a request with the state created');
+      }
     }
     requestCopied.state = RequestLogicTypes.STATE.CANCELED;
     return requestCopied;
   }
 
   if (signerRole === RequestLogicTypes.ROLE.PAYEE) {
-    if (request.state === RequestLogicTypes.STATE.CANCELED) {
```
benjlevesque, line 72: does this mean we were not throwing if a PAYER was cancelling twice? weird...

benjlevesque (approved):

packages/request-logic/src/actions/cancel.ts
KolevDarko: Yes looks like it. But this can't happen from Invoicing because the cancel/reject action doesn't show if it's already canceled/rejected once.

PR #844, opened 2022-05-13
feat: escrow transaction builder

KolevDarko:
The transaction builder now generates an approval transaction for the escrow contract, based on the options provided.
The escrow is only enabled for ERC20 fee proxy payment network.


packages/payment-processor/src/payment/encoder-approval.ts
```diff
@@ -218,6 +218,12 @@ export async function isRequestErc20ApprovalWithoutSwapNeeded(
       }
       break;
     case ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT:
+      if (options?.isEscrow) {
+        if (!(await hasEscrowErc20Approval(request, from))) {
+          return true;
+        }
+        return false;
```
leoslr, line 225:
```suggestion
        return !(await hasEscrowErc20Approval(request, from));
```

leoslr (approved):

packages/payment-detection/src/erc20/fee-proxy-contract.ts
```diff
@@ -19,6 +19,11 @@ const PROXY_CONTRACT_ADDRESS_MAP = {
   ['0.2.0']: '0.2.0',
 };
 
+const ESCROW_CONTRACT_ADDRESS_MAP = {
+  ['0.1.0']: '0.1.0',
+  ['0.2.0']: '0.1.0',
```
yomarion, line 24: Do we really need to support versions that have not been used by public?

packages/payment-processor/src/payment/encoder-payment.ts
```diff
@@ -41,10 +43,21 @@ export function encodeRequestPaymentWithoutSwap(
         ...overrides,
       };
     case ExtensionTypes.ID.PAYMENT_NETWORK_ERC20_FEE_PROXY_CONTRACT:
-      return {
-        ...prepareErc20FeeProxyPaymentTransaction(request, amount, feeAmount),
-        ...overrides,
-      };
+      if (options?.isEscrow) {
```
yomarion, line 46: I'm not comfortable with that choice. I think escrow is, for now, a minor option, and later probably a separate RN, so I think it's not worth it re-using the same `encodeRequestPayment` method, but instead make a specific one like `encodeRequestPaymentWithEscrow`.

packages/payment-processor/src/payment/erc20.ts
```diff
@@ -159,20 +174,23 @@ export function prepareApproveErc20(
  * the request in its payment currency. Can be used with a Multisig contract.
  * @param request the request to pay
  * @param signerOrProvider the Web3 provider, or signer. Defaults to window.ethereum.
+ * @param isEscrow wether to approve the payment proxy or escrow contract
  */
 export function encodeApproveErc20(
   request: ClientTypes.IRequestData,
   signerOrProvider: providers.Provider | Signer = getProvider(),
+  isEscrow = false,
```
yomarion, line 182: Same here, not comfortable with that choice. I think escrow is, for now, a minor option, and later probably a separate RN, so I think it's not worth it re-using the same method.

packages/payment-processor/test/payment/encoder-approval.test.ts
```diff
@@ -493,3 +493,38 @@ describe('Approval encoder handles Eth Requests', () => {
...
+    );
+    const escrowProxyAddress = getEscrowProxyAddress(validRequestERC20FeeProxy);
+    expect(approvalTransaction).toEqual({
+      data: erc20ApprovalData(escrowProxyAddress),
+      to: erc20ContractAddress,
+      value: 0,
+    });
+  });
+
+  it('Should not return anything', async () => {
```
yomarion, line 513:
```suggestion
  it('Should not return a transaction if the token is already approved', async () => {
```

yomarion (commented):
I'm not sure about the interface override `isEscrow`, can we discuss that?

PR #848, opened 2022-05-24
feat: currency type guards

leoslr:
This allows to easily determine the type of a currency but also correctly type an array of currencies when filtered.


kevindavee (approved):

benjlevesque (approved):
😍 😍 😍

PR #849, opened 2022-05-24
fix: escrow ci deployment

KolevDarko:
## Description of the changes
Was trying to add the Escrow contract to the request-contracts Docker image.
First I created new escrow deployment script that uses the old way for deployment in scripts/deploy-payments.ts
Then it turned out that test-deploy-all should be used, in there I made another fix, removed the hardcoded Erc20FeeProxy and I'm getting it as an argument.


leoslr (approved):
I'm not sure that it's a good idea to have an escrow deployment script in `deploy-payments.ts` as we should deploy it through the create2 scripts.

yomarion (commented):
Seems good, but I agree with @leoslr , we should use xDeployer

PR #852, opened 2022-05-30
fix(data-format): remove invoice item currency constraints

alexandre-abrioux:
There is no need for constraints on invoice items' currency string. The currency name can be longer than 10 characters, for instance `ETH-rinkeby` (11 characters). I chose to update past format versions to fix the bug everywhere it was present. I consider this operation as safe as it is a constraint removal. Let me know if you would prefer only updating the latest version, or even creating a whole new version instead.


kevindavee (approved):

packages/data-format/src/format/rnf_invoice/rnf_invoice-0.0.1.json
```diff
@@ -130,9 +130,7 @@
             "type": "number"
           },
           "currency": {
-            "type": "string",
-            "minLength": 3,
-            "maxLength": 3
+            "type": "string"
```
yomarion, line 133: I'm up for keeping the minimum length constraint, to explicitely prevent "€" for instance.

yomarion (approved):

benjlevesque (approved):

PR #877, opened 2022-07-04
feat(smart-contracts): batch conversion

olivier7delf:
## Add the feature batch conversion ERC20
- can pay a batch of conversion ERC20 payments
- can pay in the same time conversion and non_conversion ERC20 payments: like gnosisSafe
- Include batch eth payment, even if it is not gonna be used directly because it requires modifications (cf Asana)

Archi: https://miro.com/app/board/uXjVOmOyMjg=/ , please unzoom to see also ERC20 + ETH archi
- Proposition 1) BatchConversionPayment inherits from BatchPayment(Public), the one developed here.
- Proposition 2 (aborted) BatchConversionPayment call BatchPayment when needed. But it implies some difficulties with approvals and transfer of tokens between proxies: no developed.)


packages/smart-contracts/package.json
```diff
@@ -35,6 +35,8 @@
     "build:lib": "tsc -b tsconfig.build.json && cp src/types/*.d.ts dist/src/types && cp -r dist/src/types types",
     "build:sol": "yarn hardhat compile",
     "build": "yarn build:sol && yarn build:lib",
+    "build:copy": "cp build/src/contracts/BatchConversionPayments.sol/BatchConversionPayments.json src/lib/artifacts/BatchConversionPayments",
```
olivier7delf, line 38: I am gonna clean it after it has been reviewed

olivier7delf, line 38: I will clean this file once it is reviewed

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,409 @@
...
+  ) public payable {
+    uint256 contractBalance = address(this).balance;
+    // amountAndFeeToPay in native token (as ETH), is updated at each payment
+    uint256 amountAndFeeToPay;
+
+    // rPaths stores _path, rate, and decimals only once by path
+    Path[] memory rPaths = new Path[](requestsInfo.length);
+    for (uint256 i = 0; i < requestsInfo.length; i++) {
+      RequestInfo memory rI = requestsInfo[i];
+      for (uint64 k = 0; k < requestsInfo.length; k++) {
```
leoslr, line 291: uint64 instead of uint256 ?

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,409 @@
...
+    conversionPaymentProxy = IERC20ConversionProxy(_paymentErc20ConversionFeeProxy);
+  }
+
+  function setEthConversionPaymentProxy(address _paymentEthConversionFeeProxy) public onlyOwner {
+    conversionPaymentEthProxy = IEthConversionProxy(_paymentEthConversionFeeProxy);
+  }
+
+  /**
+   * @notice Update the conversion path contract used to fetch conversions
+   * @param _chainlinkConversionPathAddress address of the conversion path contract
```
leoslr, line 404: Any reason why all the admin functions are public except setConversionPathAddress that is external ?

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,409 @@
...
+        requestedToken.safeTransfer(msg.sender, excessAmount);
+      }
+
+      // Payer pays batch fees amount
+      require(
+        safeTransferFrom(
+          uTokens[k].tokenAddress,
+          _feeAddress,
+          ((((uTokens[k].amountAndFee - excessAmount) * 10000) / (10000 + basicFee)) *
+            batchConversionFee) / 10000
```
leoslr, line 264: Just to be sure: we are not using `uTokens[k].batchFeeAmount` that we've computed earlier because it's an estimation, and insead we are computing the fees on the effective amount paid ?

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,409 @@
...
+
+    conversionPaymentProxy = IERC20ConversionProxy(_paymentErc20ConversionFeeProxy);
+    conversionPaymentEthProxy = IEthConversionProxy(_paymentEthConversionFeeProxy);
+    chainlinkConversionPath = ChainlinkConversionPath(_chainlinkConversionPathAddress);
+    transferOwnership(_owner);
+
+    basicFee = 0;
+    batchFee = 0;
+    batchConversionFee = 0;
+  }
```
leoslr, line 113: Can you explain the difference between these 3 kinds of fees please ? 
Batch fees are not the same for ERC20 payments and ERC20WithConversion payments ?

leoslr (commented):
I have made a first review for the main contract `BatchConversionPayments` and let a few comments.


I don't know if has already been discussed in the design of the smart contract, but shouldn't we pass the proxies addresses as parameters when calling the batch functions ?

The fact that they're variables within the contract means that if we update one of our proxies, all the previously created invoices for that proxy won't be compatible with batch.

On invoicing side it would implicate that we filter out invoices that were created with the previous payment network version, else our users payment would not be detected.

If we do so, it might confuse our users, especially if we advertise that all invoices are payable through batch.

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,409 @@
...
+    // For each token: check allowance, transfer funds on the contract and approve the paymentProxy to spend if needed
+    for (uint256 k = 0; k < uTokens.length && uTokens[k].amountAndFee > 0; k++) {
+      requestedToken = IERC20(uTokens[k].tokenAddress);
+      uTokens[k].batchFeeAmount = (uTokens[k].amountAndFee * batchConversionFee) / 10000;
+      // Check proxy's allowance from user, and user's funds to pay approximated amounts.
+      require(
+        requestedToken.allowance(msg.sender, address(this)) >= uTokens[k].amountAndFee,
+        'Not sufficient allowance for batch to pay'
+      );
+      require(requestedToken.balanceOf(msg.sender) >= uTokens[k].amountAndFee, 'not enough funds');
```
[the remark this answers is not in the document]
olivier7delf: It helps to find the error if a developer put enough tokens, but forget batchFees.
But if people doesn't need it, I will erase it

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
olivier7delf: right thx

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
olivier7delf: I did this choice to keep flexibilty.
basicFee are the one e

olivier7delf: I have just added: 
// Between 0 and 10000,fees applied for basic invoice, 0.1% at Request Finance
  uint256 public basicFee;

I did this choice to keep flexibilty.
basicFee are set for years at .1%.
We don't know yet if we are gonna apply the same rate for batch "basic" and batch "conversion". The second one required more expertise, which could justify a higher fee in the future (i.e: batch = 0.2% and batchConv = 0.3%).

But if we are sure to keep it this way, I can change it.
(I am cautious, because redeploy these contracts would be very painful for customers, they would have to "approve" again the proxies)

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
olivier7delf: exactly, it is only for require checking.
I have added:
 * @dev amountAndFee is an approximation of the amount and the fee to be paid, in order to get enough tokens.
   *                   The excess is sent back to the payer
   *      batchFeeAmount is an approximation for the same reason of amountAndFee

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
olivier7delf: Thx !

olivier7delf:
Thanks for the review!

Regarding the proxies, I agree it can be an issue if they are updated.
But it would increase the cost  (and the complexity) to add them to function args.

- Does it arrive often?
- How bad is it if they cannot use it?

IMO, it should be very exceptional, a solution could be to deploy a second batchConv proxy. This one would keep the old proxies, then, if a user really wants to batch, he can do it. (even if he has to approve again the tokens)
Is this solution enough?

leoslr:
@olivier7delf it should not happen very often, but it most likely will at some point, and when that happens there will most likely be issues.
Even if we redeploy a second batch contract, the user won't be able to batch pay invoices created for different versions of the same payment network.

I'am not sure it would increase complexity, for example we could:
- Add a `mapping(address => uint256)`. 
- For each Proxy update the whitelist and map the address to the number associated to the proxy kind.
- Instead of passing the number associated to the proxy kind as we're doing now, pass the proxy address.

If we do that we know that the addressed passed are trusted proxies that we've added, and which kind they are.
Maybe it's something we can do in a V2, so it's not a blocker for now but IMO we need to anticipate this else it will become a pain point in the future.

olivier7delf:
It has been discussed with Yoann that in order to get batch conversion ASAP, we keep the current version. If we need to update the proxies, we will deploy another batchConv contract (This means we could need to do 2 batches instead of 1 in this specific case)
The next generation of batch contracts will get this possibility to make proxies addresses evolve.

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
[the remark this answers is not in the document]
olivier7delf: I have deleted this part

packages/smart-contracts/scripts/test-deploy-batch-conversion-deployment.ts
```diff
@@ -0,0 +1,70 @@
...
+  args: any,
+  hre: HardhatRuntimeEnvironment,
+): Promise<any> {
+  try {
+    console.log('start BatchConversionPayments');
+    const _ERC20FeeProxyAddress = '0x75c35C980C0d37ef46DF04d31A140b65503c0eEd';
+    const _EthereumFeeProxyAddress = '0x3d49d1eF2adE060a33c6E6Aa213513A7EE9a6241';
+    const _chainlinkConversionPath = '0x4e71920b7330515faf5EA0c690f1aD06a85fB60c';
+    const _paymentErc20ConversionFeeProxy = '0xdE5491f774F0Cb009ABcEA7326342E105dbb1B2E';
+    const _paymentEthConversionFeeProxy = '0x98d9f9e8DEbd4A632682ba207670d2a5ACD3c489';
```
leoslr, line 20: It would be better to get these addresses from the previous deployment instead of hardocing them.
If there is an update in the previous script that modify the deployment order it will break the tests.

leoslr (approved):

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,692 @@
...
+            },
+          ];
+        };
+      }
+      if (suiteName === 'batchERC20ConversionPaymentsMultiTokens') {
+        batchConvFunction =
+          testBatchConversionProxy.connect(_signer).batchERC20ConversionPaymentsMultiTokens;
+        argTemplate = (requestInfos: RequestInfo[]) => {
+          return requestInfos;
+        };
```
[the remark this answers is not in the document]
olivier7delf: argTemplate depend also of suiteName, that's it is duplicate but the content of the function is different.
"- GAS" was a remainder, I delete it.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,692 @@
...
+   * It can also be used for batch IF batchFee == batchConvFee
+   * The 3rd arg is used to calcul batch fees
+   */
+  const calculERC20Balances = (
+    _conversionToPay_result: BigNumber,
+    _conversionFees_result: BigNumber,
+    _conversionsToPay_results: BigNumber[],
+  ) => {
+    fromDiffBalanceExpected = fromDiffBalanceExpected
+      .add(_conversionToPay_result)
```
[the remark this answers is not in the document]
olivier7delf: Sure, I am on it

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,364 @@
...
+
+  /**
+   * @notice Fees applied for basic invoice, 0.1% at Request Finance
+   * @param _basicFee Between 0 and 10000, e.i: basicFee = 10 represent 0.10% of fees
+   *         Update it cautiously.
+   *         e.i: Only if the Request Finance 'basicFee' has evolve, which should be exceptional
+   */
+  function setBasicFee(uint256 _basicFee) external onlyOwner {
+    basicFee = _basicFee;
+  }
```
yomarion, line 331: According to our oral discussion, this should not be needed anymore.

yomarion (commented):
CR in progress, I have mostly reviewed the contract. Impressive work, the contract is big and complex but solid. Please mind the english mistakes in naming and comments.

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,364 @@
...
+   * @dev amountAndFee is an approximation of the amount and the fee to be paid, in order to get enough tokens.
+   *                   The excess is sent back to the payer
+   *      batchFeeAmount is an approximation for the same reason of amountAndFee
+   */
+  function batchERC20ConversionPaymentsMultiTokens(
+    RequestInfo[] calldata requestsInfo,
+    address _feeAddress
+  ) public {
+    // Aggregate _maxToSpend by token
+    Token[] memory uTokens = new Token[](requestsInfo.length);
```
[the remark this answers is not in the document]
olivier7delf: // a list of unique tokens, with the sum of maxToSpend by token

packages/smart-contracts/scripts/test-deploy-batch-conversion-deployment.ts
```diff
@@ -0,0 +1,70 @@
...
+          _chainlinkConversionPath,
+          await (await hre.ethers.getSigners())[0].getAddress(),
+        ],
+      },
+    );
+
+    // Initialize batch conversion fee, useful to others packages.
+    const [owner] = await hre.ethers.getSigners();
+    const batchConversion = batchConversionPaymentsArtifact.connect(hre.network.name, owner);
+    await batchConversion.connect(owner).setBasicFee(10);
```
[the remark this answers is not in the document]
olivier7delf: the function exist, but after our talk, I am gonna erase the function and it

packages/smart-contracts/src/contracts/BatchConversionPayments.sol
```diff
@@ -0,0 +1,364 @@
...
+    chainlinkConversionPath = ChainlinkConversionPath(_chainlinkConversionPathAddress);
+    transferOwnership(_owner);
+
+    basicFee = 0;
+    batchFee = 0;
+    batchConversionFee = 0;
+  }
+
+  // batch Eth requires batch contract to receive funds from ethFeeProxy with a value = 0
+  //            and also from paymentEthConversionProxy with a value > 0
```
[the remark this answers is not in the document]
olivier7delf: done

yomarion (commented):
Good job on the contracts, they are super neat, but please take some time to refactor the tests, they are hard to read and validate. Tests should be very straightforward, we don't want them to be potentially buggy.

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+   * @param isBatchRouter allows to use the function "batchERC20ConversionPaymentsMultiTokens"
+   *                     through the batchRouter or directly
+   */
+  const ERC20ConversionTestSuite = (isBatchRouter: boolean) => {
+    before(() => {
+      setBatchConvFunction(isBatchRouter, signer);
+    });
+
+    describe(isBatchRouter ? 'Through batchRouter' : 'Without batchRouter', () => {
+      describe('batchERC20ConversionPaymentsMultiTokens with DAI', async () => {
```
yomarion, line 438: I think the function definition + calls can be eased with a simple loop through 2 values, and does not require the `before` section

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+  let fromDiffBalanceExpected: BigNumber;
+  let toDiffBalanceExpected: BigNumber;
+  let feeDiffBalanceExpected: BigNumber;
+
+  // variables needed for chainlink and
+  let path: string[];
+  type ConvToPay = [BigNumber, BigNumber] & {
+    result: BigNumber;
+    oldestRateTimestamp: BigNumber;
+  };
```
yomarion, line 74: I don't understand this type and it looks suspicious that we need a type definition for tests. Wouldn't it be enough with a fixed rate for example? It's not like we have to really test the conversion in these tests.

yomarion (commented):
I've tried to provide a few tips to simplify the tests but I understand there are plenty of similar test cases to run with a panel of variations. You have to find an intermediate approach, there are too many nested functions for a test.

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+      });
+
+      it('Not enough allowance', async function () {
+        // xSigner connect to the batch function
+        setBatchConvFunction(isBatchRouter, xSigner);
+        await expect(batchConvFunction(argTemplate([convDetail]), feeAddress)).to.be.revertedWith(
+          'Insufficient allowance for batch to pay',
+        );
+        // reset: signer connect to the batch function
+        setBatchConvFunction(isBatchRouter, signer);
```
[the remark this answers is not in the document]
olivier7delf: done

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+        await expect(batchConvFunction(argTemplate([convDetail]), feeAddress)).to.be.revertedWith(
+          'not enough funds, including fees',
+        );
+
+        // reset:
+        //   - decrease xSigner allowance
+        //   - connect with signer account
+        await testERC20.connect(xSigner).approve(testBatchConversionProxy.address, '0');
+        testERC20.connect(signer);
+        setBatchConvFunction(isBatchRouter, signer);
```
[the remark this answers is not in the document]
olivier7delf: I have added setBatchConvFunction to the afterEach.
But not the approve revoked line 513, because it is only used here

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+        from,
+        feeAddress,
+        amount * (batchFee / 10_000), // batch fee amount = 200000 * .5%
+      );
+
+    [fromDiffBalanceExpected, toDiffBalanceExpected, feeDiffBalanceExpected] =
+      expectedERC20Balances([BigNumber.from(amount)], [BigNumber.from(feeAmount)], batchFee, false);
+  };
+  /** Make sure the existing ERC20 functions from the parent contract BatchPaymentPublic.sol are still working */
+  describe('Test BatchErc20Payments functions', () => {
```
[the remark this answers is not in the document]
olivier7delf: I get the feeling that 2 nested loops are not decreasing the complexity

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,611 @@
...
+  };
+
+  /**
+   * @notice it does the set up of the main variables, it selects and pays with
+   *         the desired batch EC20 batch function (no conversion), and it checks the ERC20 balances
+   * @param isBatchRouter allows to use a function through the batchRouter or not
+   * @param erc20Function selects the batch function name tested: "batchERC20PaymentsWithReference"
+   *                      or "batchERC20PaymentsMultiTokensWithReference"
+   */
+  const batchERC20Payments = async (isBatchRouter: boolean, erc20Function: string) => {
```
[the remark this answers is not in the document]
olivier7delf: added

packages/smart-contracts/test/contracts/ERC20BatchConversionPayments.test.ts
olivier7delf: Done for the loop.
Regarding before, we need it for every functions, I don't see another way really cleaner

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+    });
+
+    describe(useBatchRouter ? 'Through batchRouter' : 'Without batchRouter', () => {
+      describe('batchERC20ConversionPaymentsMultiTokens with DAI', async () => {
+        it('allows to transfer DAI tokens for USD payment', async () => {
+          await onePaymentBatchConv(path);
+        });
+        it('allows to transfer DAI tokens for EUR payment', async () => {
+          path = [EUR_hash, USD_hash, DAI_address];
+          await onePaymentBatchConv(path);
```
yomarion, line 497:
```suggestion
          await onePaymentBatchConv([EUR_hash, USD_hash, DAI_address]);
```

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+        });
+        it('allows to transfer 2 transactions DAI tokens for USD payment', async function () {
+          await manyPaymentsBatchConv(path, 1);
+        });
+        it('allows to transfer DAI tokens for EUR payment', async () => {
+          path = [EUR_hash, USD_hash, DAI_address];
+          await onePaymentBatchConv(path);
```
yomarion, line 504:
```suggestion
          await onePaymentBatchConv([EUR_hash, USD_hash, DAI_address]);
```

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+        });
+        it('allows to transfer 2 transactions DAI tokens for USD and EUR payments', async function () {
+          const path2 = [EUR_hash, USD_hash, DAI_address];
+          await manyPaymentsBatchConv(path2, 1);
+        });
```
yomarion, line 509:
```suggestion
          await manyPaymentsBatchConv([EUR_hash, USD_hash, DAI_address], 1);
        });
```

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+        it('allows to transfer two kinds of tokens for USD', async function () {
+          const path2 = [USD_hash, FAU_address];
+          await manyPaymentsBatchConv(path2, 1);
```
yomarion, line 512:
```suggestion
          await manyPaymentsBatchConv([USD_hash, FAU_address], 1);
```

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+  let fromOldBalance: BigNumber;
+  let toOldBalance: BigNumber;
+  let feeOldBalance: BigNumber;
+
+  let fromDiffBalanceExpected: BigNumber;
+  let toDiffBalanceExpected: BigNumber;
+  let feeDiffBalanceExpected: BigNumber;
+
+  // variables needed for chainlink and conversion payments
+  let path: string[];
```
yomarion, line 70: It does not look needed to have this as a variable. I suggested a few replacements here and there in that direction, but it should lead to removing this variable.

yomarion: I took this one as an example but you should avoid using variables like this, it complexifies reading and maintenance a lot.

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+    describe(useBatchRouter ? 'Through batchRouter' : 'Without batchRouter', () => {
+      describe('batchERC20ConversionPaymentsMultiTokens with DAI', async () => {
+        it('allows to transfer DAI tokens for USD payment', async () => {
+          await onePaymentBatchConv(path);
+        });
+        it('allows to transfer DAI tokens for EUR payment', async () => {
+          path = [EUR_hash, USD_hash, DAI_address];
+          await onePaymentBatchConv(path);
+        });
+        it('allows to transfer 2 transactions DAI tokens for USD payment', async function () {
```
yomarion, line 499: I don't understand the title of this unit test and the absence of assertion looks weird.

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+    [fromDiffBalanceExpected, toDiffBalanceExpected, feeDiffBalanceExpected] =
+      expectedERC20Balances([conversionToPay], [conversionFees], batchConvFee);
+  };
+
+  /**
+   * @notice generate nTimes 2 convDetails, do an ERC20 conv batch payment with theses 2*nTimes payments
+   *         and calculate the balances
+   * @param path2 to update the second convDetail
+   */
+  const manyPaymentsBatchConv = async (path2: string[], nTimes: number) => {
```
yomarion, line 272: This method is only called with `nTimes = 1`, and it's quite hard to understand it. It's worth a simplification.

Maybe you can simplify the global structure of these tests by removing the dependency on the aggregator. Define a few fixed conversion rates (e.g 1.1, 2.2 and 3.3), set the oracles and you don't need a method to compute expected amounts anymore.

yomarion (commented):
(review in progress)

packages/smart-contracts/test/contracts/BatchConversionErc20Payments.test.ts
```diff
@@ -0,0 +1,586 @@
...
+
+  /**
+   * @notice it contains all the tests related to the ERC20 batch payment, and its context required
+   * @param useBatchRouter allows to use the function "batchERC20ConversionPaymentsMultiTokens"
+   *                     through the batchRouter or directly
+   */
+  for (const useBatchRouter of [true, false]) {
+    before(async () => {
+      [from, to, feeAddress] = (await ethers.getSigners()).map((s) => s.address);
+      [signer1, signer4, signer4, signer4] = await ethers.getSigners();
```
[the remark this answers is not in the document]
olivier7delf: " handled the unhappy paths like this one separately" do you mean with another test file ?

packages/smart-contracts/scripts/test-deploy-batch-conversion-deployment.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const FAU_USD_RATE = 201; // 2.01
+
+// Deploys, set up the contracts
+export async function deployBatchConversionPayment(
+  args: any,
+  hre: HardhatRuntimeEnvironment,
+): Promise<any> {
+  try {
+    console.log('start BatchConversionPayments');
```
yomarion, line 25:
```suggestion
    console.log('deploy BatchConversionPayments');
```
? Or maybe this was for debugging.

yomarion (commented):
Almost there! Well done, it looks cleaner and cleaner.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,866 @@
...
+        feeAddress,
+      );
+    } else {
+      await batchConversionProxy
+        .connect(fromSigner)
+        .batchMultiERC20ConversionPayments(
+          [fauConvDetail, daiConvDetail, daiConvDetail],
+          feeAddress,
+        );
+    }
```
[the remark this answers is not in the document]
olivier7delf: top ca

packages/types/src/payment-types.ts
```diff
@@ -342,10 +342,19 @@ export interface CryptoDetails {
   feeAmounts: Array<string>;
 }
 
+/** Each paymentNetworkId is linked with a batch function */
+export enum BatchPaymentNetworkId {
+  batchMultiERC20ConversionPayments,
```
yomarion, line 347: The enum values should be upper-case

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -829,37 +803,37 @@ describe('contract: BatchConversionPayments', async () => {
...
+      const initialFeeETHBalance = await provider.getBalance(feeAddress);
+      const initialFromETHBalance = await provider.getBalance(await fromSigner.getAddress());
       tx = await batchConversionProxy.batchEthPayments(
         [to],
         ['1000'],
         [referenceExample],
         ['1'],
         feeAddress,
-        { value: 1000 + 1 + 42 },
+        { value: 1000 + 1 + 42 }, // +42 in excess
```
yomarion, line 828:
```suggestion
        { value: 1000 + 1 + 42 }, // +42 in excess should not impact the final amount exchanged
```

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -781,16 +757,15 @@ describe('contract: BatchConversionPayments', async () => {
       wrongConvDetail.maxRateTimespan = '1';
       await expect(
         batchConversionProxy.batchEthConversionPayments([wrongConvDetail], feeAddress, {
-          value: 1000 + 1 + 42,
+          value: 1000 + 1 + 42, // +42 in excess
```
yomarion, line 760: Is the `+ 42` required for this specific unit test? It's not obvious why

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -386,56 +382,55 @@ describe('contract: BatchConversionPayments', async () => {
...
-        beforeETHBalanceFrom,
-        beforeETHBalanceTo,
-        beforeETHBalanceFee,
+        initialFromETHBalance,
+        initialToETHBalance,
+        initialFeeETHBalance,
       );
     });
 
     it('make 1 ETH payment with 1-step conversion', async function () {
```
yomarion, line 398:
```suggestion
    it('make 1 ETH payment with 1-step conversion', async () => {
```
This is the formalism we use generally

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
yomarion: (same comment on other lines)

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -456,22 +451,22 @@ describe('contract: BatchConversionPayments', async () => {
       tx = await batchConversionProxy.batchRouter(
         [
           {
-            paymentNetworkId: 0,
+            paymentNetworkId: PaymentTypes.BatchPaymentNetworkId.batchMultiERC20ConversionPayments,
```
yomarion, line 454: Import `BatchPaymentNetworkId` to shorten these references

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        .add(BigNumber.from(fee).mul(conversionRate).mul(nPayment));
+    fee;
+    const expectedFromDAIBalanceDiff = expectedToDAIBalanceDiff
+      .add(expectedFeeDAIBalanceDiff)
+      .mul(-1);
+    return [
+      expectedFromDAIBalanceDiff.div('1000000'), // divide by 1 million because we used millionDai
+      expectedToDAIBalanceDiff.div('1000000'),
+      expectedFeeDAIBalanceDiff.div('1000000'),
+    ];
```
yomarion, line 190:
```suggestion
    const precision = '1000000'; // Balance the non-support for decimals
    const EUR_USD = '120'; // TODO: move definition close to FAU_USD_RATE
    const DAI_USD = '101'; // TODO: move definition close to FAU_USD_RATE
    const conversionRate =
      path === 'EUR_DAI'
        ? BigNumber.from(daiDecimals).mul(precision).mul(EUR_USD).div(DAI_USD)
        : BigNumber.from(daiDecimals).mul(precision).mul(100).div(FAU_USD_RATE);
    const expectedToDAIBalanceDiff = BigNumber.from(amount).mul(conversionRate).mul(nPayment);
    const expectedFeeDAIBalanceDiff =
      // fee added by the batch
      expectedToDAIBalanceDiff
        .add(BigNumber.from(fee).mul(conversionRate).mul(nPayment))
        .mul(BATCH_CONV_FEE)
        .div(BATCH_DENOMINATOR)
        // fee within the invoice: .1% of the amount,
        .add(BigNumber.from(fee).mul(conversionRate).mul(nPayment));
    fee;
    const expectedFromDAIBalanceDiff = expectedToDAIBalanceDiff
      .add(expectedFeeDAIBalanceDiff)
      .mul(-1);
    return [
      expectedFromDAIBalanceDiff.div(precision),
      expectedToDAIBalanceDiff.div(precision),
      expectedFeeDAIBalanceDiff.div(precision),
    ];
```
This is a quick suggestion to replace required comments with explicit variable names.
But it looks very weird that we need to multiply by `daiDecimals` at the beginning, while we never devide by `daiDecimals`, everything should be denominated with the same level of precision, shouldn't it?

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -0,0 +1,840 @@
...
+  const expectETHBalanceDiffs = async (
+    ethAmount: BigNumber,
+    ethFeeAmount: BigNumber,
+    feeApplied = BATCH_CONV_FEE,
+    initialFromETHBalance: BigNumber,
+    initialToETHBalance: BigNumber,
+    initialFeeETHBalance: BigNumber,
+  ) => {
+    const receipt = await tx.wait();
+    const gasUsed = receipt.gasUsed.mul(2 * 10 ** 10);
```
yomarion, line 244: Why `2 * 10 ** 10` ?

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -386,56 +382,55 @@ describe('contract: BatchConversionPayments', async () => {
...
-            paymentNetworkId: 4,
+            paymentNetworkId: PaymentTypes.BatchPaymentNetworkId.batchEthConversionPayments,
             conversionDetails: [ethConvDetail],
             cryptoDetails: emptyCryptoDetails,
           },
         ],
         feeAddress,
         {
-          value: (1000 + 1 + 42) * USD_ETH_RATE,
+          value: (1000 + 1 + 42) * USD_ETH_RATE, // +42 in excess
```
yomarion, line 413: Just FYI, in the other test, it's 42 ETH in excess, here it's 42 USD, but it tests the same thing so these tests should probably be consistent with each other.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
olivier7delf: It is necessary because of the batch fees ( I could have 1001 * 10% -> about  +11 )

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
olivier7delf: The conversion aggregator multiplies by the daiDecimals.
If I delete daiDecimals, then I have to add it to the "amount" when I call this function.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
olivier7delf: 42 was a proxy for the batch fees amount: 1001 *10% which is about  +11 (I updated this)
Regarding this, we have to multiply the USD into ETH.
this will failed: (1000 + 1) * USD_ETH_RATE + 42

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
olivier7delf: 10^10: When forking a remote network, if the remote network uses EIP-1559, the first local block will use the right baseFeePerGas according to the EIP, otherwise "**10_000_000_000**" is used. from: https://hardhat.org/hardhat-network/docs/reference

olivier7delf: it is the gasPrice, it is updated

yomarion (approved):
🚀

PR #883, opened 2022-07-08
fix: payment encoder for ERC777-stream payment network

bertux:
## Description of the changes
Adding support for ERC777-stream payment network in payment encoder.


bertux:
Not needed until we can pay 2 invoices with the same SuperToken to the same address.
Will need to see it with @leoslr

bertux:
waiting for review from @leoslr to see if Signer can be used instead of Provider.

KolevDarko (approved):

packages/payment-processor/src/payment/encoder-payment.ts
```diff
@@ -10,25 +10,27 @@ import { prepareSwapToPayAnyToErc20Request } from './swap-any-to-erc20';
...
   }
 }
 
-export function encodeRequestPaymentWithoutSwap(
+export async function encodeRequestPaymentWithoutSwap(
   request: ClientTypes.IRequestData,
+  provider: providers.Provider,
   options?: IRequestPaymentOptions,
-): IPreparedTransaction {
+): Promise<IPreparedTransaction> {
```
yomarion, line 33: NOK for this breaking change. I understand where it comes from but I think it's not worth it. You should instead, as a first step at least, add an equivalent of `encodeRequestPaymentWithoutSwap`, async and specific to ERC777, and handle this case separately in your back-end.

yomarion (changes requested):
We cannot change one function signature just for one integration; it's too unsure and quite impactful. If we want to unify the logic, we should question the asynchronous aspect of building a transaction; in the end, is the SDK suited for what we do?

leoslr (approved):
Good, maybe just add comments for the superfluid functions.

packages/payment-processor/src/payment/encoder-payment.ts
bertux: OK I'm preparing it to handle it separately in the API backend.

bertux: Replaced by #896

PR #894, opened 2022-08-02
fix(payment-processor): add reset approvals to run twice tests

olivier7delf:
Tests cannot be run twice without restart Ganache, and then, deploy again smart contracts.
Solution: reset approvals for the 2 tests, using a new function in utils.

Also, update README regarding tests, preventing people being working on smart-contracts before to get weird error because of the Network.

Modifications impact only on tests.

### DIFF
--- a/packages/payment-processor/src/payment/utils.ts
+++ b/packages/payment-processor/src/payment/utils.ts
@@ -1,4 +1,4 @@
-import { ethers, Signer, providers, BigNumber, BigNumberish } from 'ethers';
+import { ethers, Signer, providers, BigNumber, BigNumberish, ContractTransaction } from 'ethers';
 
 import { PaymentReferenceCalculator, getDefaultProvider } from '@requestnetwork/payment-detection';
 import {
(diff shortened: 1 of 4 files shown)


KolevDarko (approved):

packages/payment-processor/src/payment/utils.ts
```diff
@@ -345,3 +346,27 @@ export function comparePnTypeAndVersion(
     pn?.version === getPaymentNetworkExtension(request)?.version
   );
 }
+
+/**
+ * set approval to 0 of a token from a user (as wallet address provider)
```
yomarion, line 351:
```suggestion
 * Revoke ERC20 approval of a token for a given `spenderAddress`
```

packages/payment-processor/src/payment/utils.ts
```diff
@@ -345,3 +346,27 @@ export function comparePnTypeAndVersion(
...
+ */
+export async function resetApproveErc20Utils(
```
yomarion, line 353: Would be best to rename the function `revokeErc20Approval`.

`revoke` is the term used in MM when setting an approval to 0.
`Erc20Approval` is more natural.
`Utils` adds complexity to the name, IMO.

yomarion (approved):
A nice step forward, thanks! I suggest a few style improvements.

PR #897, opened 2022-08-08
feat: add extension data at request creation

benjlevesque:
This enables creating a request with payee/payer delegates, instead of adding them after creation


KolevDarko (approved):

packages/types/src/request-logic-types.ts
```diff
@@ -287,6 +287,4 @@ export enum ROLE {
   PAYEE = 'payee',
   PAYER = 'payer',
   THIRD_PARTY = 'third-party',
-  PAYEE_DELEGATE = 'payeeDelegate',
-  PAYER_DELEGATE = 'payerDelegate',
```
leoslr, line 291: I understand the global change, but I don't get why you're deleting these roles, if you could explain please.

leoslr (approved):

packages/types/src/request-logic-types.ts
benjlevesque: yeah sorry. it's just that it's nevery used, so cleaning this up

alexandre-abrioux (approved):

packages/advanced-logic/src/extensions/abstract-extension.ts
```diff
@@ -67,7 +67,7 @@ export abstract class AbstractExtension<TCreationParameters> {
     }
 
     // if the action is not "create", the state must have been created before
-    if (!requestState.extensions[extensionAction.id]) {
+    if (!extensionsState[extensionAction.id]) {
```
benjlevesque, line 70: `requestState.extensions` is set only after all extensions have been applied

packages/request-client.js/src/api/request-network.ts
```diff
@@ -405,10 +405,6 @@ export default class RequestNetwork {
     const contentData = parameters.contentData;
     const topics = parameters.topics?.slice() || [];
 
-    if (requestParameters.extensionsData) {
```
benjlevesque, line 408: ⚠️ the protocol now supports adding extensions data at creation

packages/types/src/request-logic-types.ts
bertux: @benjlevesque that's what some builder was considering to use, so do you think it could be added back ?

benjlevesque: I just removed the enum which was useless. the feature is still active

PR #900, opened 2022-08-12
feat: erc777 superfluid get balance of subsequent requests

KolevDarko:
# Calculating the balance of superfluid requests
## How the balance is calculated
1. If calculating the balance for the first request, during the first month - all the balance is attributed to that request.
2. If calculating the balance of subsequent request, we get the full balance for the stream and then subtract the number of  previous request * expectedAmount. Whatever is left is the balance of this request.
 We can easily get the number of previous requests because each request has recurrenceNumber that represents exactly that.

3. Because we will always be creating subsequent requests until the stream is stopped, there can be no overpayment for this PN.


packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,65 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
     super(PaymentTypes.PAYMENT_NETWORK_ID.ERC777_STREAM, advancedLogic.extensions.erc777Stream);
   }
 
+  protected isSubsequentRequest(request: RequestLogicTypes.IRequest): boolean {
+    return !!request.extensions[ExtensionTypes.ID.PAYMENT_NETWORK_ERC777_STREAM].values
+      .masterRequestId;
```
benjlevesque, line 27: could this be named `originalRequestId`?

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from '@requestnetwork/types';
 import { SuperFluidInfoRetriever } from './superfluid-retriever';
 import { ReferenceBasedDetector } from '../reference-based-detector';
+import { PaymentReferenceCalculator } from '..';
```
benjlevesque, line 9: avoid this kind of imports, as it can create inter-dependant imports.

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,65 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+
   /**
-   * Extracts the balance and events of an address
+   * This returns the specific values we store for the ERC777 extensions.
+   * The extra fields specific for this extension are:
+   * expectedStartDate: The expected start date of streaming
+   * expectedFlowRate: extensionAction.parameters.expectedFlowRate,
+   * previousRequestId: The previous requestId in the streaming series
+   * masterRequestId: The first requestId of the streaming series
+   * recurrenceNumber: The position of this request in the streaming series, 0 indexed.
```
benjlevesque, line 37: why not have a type?

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,65 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+   * @param request The request we are processing
+   * @returns Object
+   */
+  protected getSubsequentValues(request: RequestLogicTypes.IRequest): Record<string, any> {
```
benjlevesque, line 41: it seems we do this in every detector without a method

btw, you can probably do `request.extensions[this.extensionId].values`

packages/payment-detection/src/erc777/superfluid-detector.ts
benjlevesque, line 37: recurrenceNumber => occurrenceNumber?

packages/advanced-logic/test/utils/payment-network/erc777/stream-create-data-generator.ts
```diff
@@ -32,17 +32,7 @@ export const actionCreationFull = {
...
-export const actionCreationFullSubsequent = {
-  action: RequestLogicTypes.ACTION_NAME.CREATE,
-  id: ExtensionTypes.ID.PAYMENT_NETWORK_ERC777_STREAM,
-  parameters: {
-    masterRequestId,
-    previousRequestId,
-    recurrenceNumber,
-  },
-  version,
+  version: '0.1.0',
```
alexandre-abrioux, line 35: It looks like you're not using this code above, you may be able to clean it:
https://github.com/RequestNetwork/requestNetwork/blob/eb6d39d0db93ec53c9d61d7e2dd16b3da44ce831/packages/advanced-logic/test/utils/payment-network/erc777/stream-create-data-generator.ts#L18-L20

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,65 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+        const subrequestValues = this.getSubsequentValues(request);
+        expectedPreviousBalance =
+          subrequestValues.recurrenceNumber * parseFloat(request.expectedAmount.toString());
+      }
+      const remainingBalance = parseFloat(totalBalance.balance) - expectedPreviousBalance;
+      let requestBalance = 0;
+      if (remainingBalance > request.expectedAmount) {
+        requestBalance = parseFloat(request.expectedAmount.toString());
+      } else {
+        requestBalance = remainingBalance > 0 ? remainingBalance : 0;
```
alexandre-abrioux, line 68: Shouldn't you make all those computations with `BigNumber`?

packages/payment-detection/src/erc777/superfluid-detector.ts
KolevDarko: Ah yes, Leo added this later, I'll include it from types

packages/payment-detection/src/erc777/superfluid-detector.ts
KolevDarko: yes close, it works with this.paymentNetworkId

packages/payment-detection/src/erc777/superfluid-detector.ts
KolevDarko: Done. I keep forgetting about BigNumber

packages/payment-detection/src/erc777/superfluid-detector.ts
KolevDarko: I'm not sure about this. Recurrence number implies recurring requests which is what we have here. 
Occurrence means multiple identical events.
Recurrence means one originating event happened and then follows a list of recurring events. In our case only the subsequent requests have recurrence number which makes sense to me.

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -56,4 +114,26 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+    const { paymentAddress, salt } = this.getPaymentExtension(request).values;
+    this.checkRequiredParameter(paymentAddress, 'paymentAddress');
+    this.checkRequiredParameter(salt, 'salt');
+    if (this.isSubsequentRequest(request)) {
+      const originalRequestId =
+        request.extensions[ExtensionTypes.ID.PAYMENT_NETWORK_ERC777_STREAM].values
+          .originalRequestId;
+      return PaymentReferenceCalculator.calculate(originalRequestId, salt, paymentAddress);
+    }
+    return PaymentReferenceCalculator.calculate(request.requestId, salt, paymentAddress);
```
benjlevesque, line 137:
```suggestion
    const requestId = this.isSubsequentRequest(request) ? this.getSubsequentValues(request).originalRequestId : request.requestId;
    return PaymentReferenceCalculator.calculate(requestId, salt, paymentAddress);
```

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,72 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+      return totalBalance;
+    }
+    if (totalBalance.balance) {
+      const zeroBN = BigNumber.from(0);
+      let expectedPreviousBalance = zeroBN;
+      if (this.isSubsequentRequest(request)) {
+        const subrequestValues = this.getSubsequentValues(request);
+        expectedPreviousBalance = BigNumber.from(subrequestValues.recurrenceNumber).mul(
+          request.expectedAmount,
+        );
```
benjlevesque, line 67: Computing the balance on the recurrence number seems quite fragile, but I don't have a strong knowledge of superfluid. 

For instance, what ensures that this number is "valid" (ie, continuous)?

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,72 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+      const expectedAmount = BigNumber.from(request.expectedAmount);
+      if (remainingBalance.gte(expectedAmount)) {
+        requestBalance = expectedAmount;
+      } else {
+        requestBalance = remainingBalance.gt(zeroBN) ? remainingBalance : zeroBN;
+      }
+      totalBalance.balance = requestBalance.toString();
+    }
+    return totalBalance;
+  }
```
benjlevesque, line 80: this whole code block is hard to read. I think you can reduce conditions: 

```ts
const recurrenceNumber = this.getSubsequentValues(request).recurrenceNumber || 0;
// This request's balance is the total streamed amount, less all previous requests balance.
const balance =  BigNumber.from(totalBalance.balance).sub(BigNumber.from(request.expectedAmount).mul(recurrenceNumber);
// Balance is stricty between 0 and expectedAmount
return balance.lt(0) ? constants.Zero : balance.gt(expectedAmount) ? expectedAmount: balance;
```

packages/payment-detection/src/erc777/superfluid-detector.ts
KolevDarko: Nothing strict for now. But we can ensure that each subsequent request has their recurrenceNumber as +1 from the previous one, we have previousRequestId. 
But that would add the complexity of fetching the previous request on each request creation and would add the dependency on requestLogic, currently advanced logic doesn't use request logic.

Another way is to take into consideration the expectedStartDate from the original requestId and calculate the number of months passed since then, to calculate the recurrence. In this case we depend on the original start date, is that more or less fragile?, and we also limit ourselves to 1 month duration.

I think this is enough for the MVP though

packages/advanced-logic/test/utils/payment-network/erc777/stream-create-data-generator.ts
```diff
@@ -202,21 +185,60 @@ export const requestStateNoExtensions: RequestLogicTypes.IRequest = {
...
+    value: TestData.payeeRaw.address,
+  },
+  payer: {
+    type: IdentityTypes.TYPE.ETHEREUM_ADDRESS,
+    value: TestData.payerRaw.address,
+  },
+  requestId: TestData.requestIdMock,
+  state: RequestLogicTypes.STATE.CREATED,
+  timestamp: TestData.arbitraryTimestamp,
+  version: '0.1.0',
```
alexandre-abrioux, line 241:
```suggestion
  version,
```

packages/payment-detection/src/erc777/superfluid-detector.ts
```diff
@@ -21,15 +22,69 @@ export class SuperFluidPaymentDetector extends ReferenceBasedDetector<
...
+    return request.extensions[this.paymentNetworkId].values;
+  }
+
   /**
-   * Extracts the balance and events of an address
+   * With streaming requests we have a custom balance computation.
+   * It is possible to have only one streaming event, start stream,
+   * but have 10 requests being created each month for the same stream.
+   * For the original request, the first one, the balance computation is easy because we don't care about previous expected balance.
+   * For subsequent requests we must take in to consideration how many requests were before this one and then attribute the rest of the balance to this one.
```
alexandre-abrioux, line 44:
```suggestion
   * For subsequent requests we must take into consideration how many requests were before this one and then attribute the rest of the balance to this one.
```

packages/payment-detection/test/erc777/superfluid-detector.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        recurrenceNumber: 2,
+      },
+      version: '0',
+    },
+  },
+};
+
+/**
+ * Generates transfer event with specific balance depending on the month of observation.
+ * The monthNumber is not the month ID like 1 for Jan, 2 for Feb, but instead
```
alexandre-abrioux, line 119: > but instead

but instead what? 😆

alexandre-abrioux (approved):
Great! Just some wording changes but LGTM

packages/payment-detection/src/erc777/superfluid-detector.ts
bertux: @KolevDarko the recurring period is stored in the request so we are not limited to 1 month duration, it's just the UI which we have limited for MVP to 1 month duration. For a quick demo we should be able to set shorter period to see the recurrence happening during the demo.

bertux: It should be remembered for after the MVP to not accept any recurrenceNumber but to only set it at creation from the recurrenceNumber of previousRequestId.

packages/advanced-logic/test/utils/payment-network/erc777/stream-create-data-generator.ts
```diff
@@ -160,7 +143,7 @@ const baseRequestState = {
     value: TestData.payeeRaw.address,
   },
   currency: {
-    network: network,
+    network: RINKEBY,
```
bertux, line 146: by keeping the variable name network you can remove : network as Alex suggested it for version.
So that's another argument to keep naming it network.

bertux (changes requested):
network variable should not be renamed to RINKEBY

bertux (approved):
thank you for the naming fix.

PR #905, opened 2022-08-24
fix: salt in declarative pn extension

benjlevesque:


KolevDarko (approved):
Good but one test for Bitcoin declarative is not expecting the salt and is failing

leoslr (approved):

PR #907, opened 2022-08-25
fix(payment-processor): within tests - revoke some approvals and use equalTo instead of eq

olivier7delf:
- revoke approvals for two old tests
- use .equalTo instead of .eq -> the tests were not working (correct one test also)

### DIFF
--- a/packages/payment-processor/test/payment/erc20.test.ts
+++ b/packages/payment-processor/test/payment/erc20.test.ts
@@ -14,15 +14,24 @@ import {
   hasErc20Approval,
   checkErc20Allowance,
 } from '../../src/payment/erc20';
+import {
+  getProxyAddress,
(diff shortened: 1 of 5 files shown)


KolevDarko (approved):
Great catch on the tests 👍

packages/payment-processor/test/payment/erc20.test.ts
```diff
@@ -14,15 +14,24 @@ import {
   hasErc20Approval,
   checkErc20Allowance,
 } from '../../src/payment/erc20';
+import {
+  getProxyAddress,
+  revokeErc20Approval,
+} from '@requestnetwork/payment-processor/src/payment/utils';
```
benjlevesque, line 20: wrong import

benjlevesque (approved):

PR #908, opened 2022-08-29
feat: add streamEventName to streaming event parameters

KolevDarko:
## Description of the changes
Adding stream event names to streaming event parameters in the balance object.
The reason for this is that we need to be able to detect stream specific events like start and stop stream in portal for sending notifications and knowing when to stop updating the balance.


benjlevesque (commented):
discussed in DM: events should already contain the necessary information

benjlevesque (approved):

leoslr (approved):

PR #912, opened 2022-09-07
feat(smart-contracts):batch conversion xdeployer

olivier7delf:
Deploy batch conversion proxies on each network


packages/smart-contracts/scripts-create2/constructor-args.ts
```diff
@@ -49,6 +49,20 @@ export const getConstructorArgs = (contract: string, network?: string): string[]
...
+      if (!network) {
+        throw new Error(
+          'Batch conversion contract requires network parameter to get correct address of erc20FeeProxy, erc20ConversionFeeProxy, ethereumFeeProxy, and ethereumConversionFeeProxy',
+        );
+      }
+      return [
+        '0x0000000000000000000000000000000000000000',
+        '0x0000000000000000000000000000000000000000',
+        '0x0000000000000000000000000000000000000000',
+        '0x0000000000000000000000000000000000000000',
```
KolevDarko, line 62: Why all 0 addresses? You can get the correct addresses with getArtifact for this specific network

KolevDarko (approved):

packages/smart-contracts/scripts-create2/constructor-args.ts
olivier7delf: I think it can be tricky because we still have some proxies that have different addresses depending of the network.
In my mind, this would lead to the generation of a different proxy address depending of the network

KolevDarko: That's ok, I see we can update them later

MantisClone:
@alexandre-abrioux @leoslr @olivier7delf I'd like to close this PR because it is obsolete. Any objections?

PR #914, opened 2022-09-07
fix: superfluid block type

leoslr:
Switch back the block type to number.


KolevDarko (commented):
@leoslr you need to do the same thing on line 116

KolevDarko (approved):

benjlevesque (approved):

PR #921, opened 2022-09-26
feat: near conversion payment processor

leoslr:
## Description of the changes

Payment processor for any-to-near (only the commit starting from `near payment with conversion`
[Asana task](https://app.asana.com/0/1202922067864171/1202201668706294)

## Blocker 

[x] [This PR](https://github.com/RequestNetwork/requestNetwork/pull/922) should be merged first to limit conflicts

### DIFF
--- a/packages/payment-processor/src/payment/utils-near.ts
+++ b/packages/payment-processor/src/payment/utils-near.ts
@@ -94,7 +94,7 @@ export const processNearPaymentWithConversion = async (
   currency: string,
   feeAddress: string,
   feeAmount: BigNumberish,
-  maxRateTimespan: string = '0',
+  maxRateTimespan = '0',
   version = '0.1.0',
 ): Promise<void> => {
   if (version !== '0.1.0') {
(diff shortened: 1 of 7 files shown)


packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -77,3 +80,57 @@ export const processNearPayment = async (
     throw new Error(`Could not pay Near request. Got ${e.message}`);
   }
 };
+
+/**
+ * Export used for mocking only.
+ */
+export const processNearPaymentWithConversion = async (
```
yomarion, line 87: I think this one deserves a bit of documentation for amount decimals (fiat at least), as well as the timespan unit, the currency... Similar to the contract itself.

alexandre-abrioux (commented):
To respect previous naming conventions, instead of having:
```
src/payment-detection/
  near/
    near-conversion-detector.ts
    near-detector.ts
```
; shouldn't you have the following?
```
src/payment-detection/
  any/
    any-to-near-proxy.ts
    retrievers/
      any-to-near-proxy.ts
```

yomarion:
> To respect previous naming conventions, instead of having:
> 
> ```
> src/payment-detection/
>   near/
>     near-conversion-detector.ts
>     near-detector.ts
> ```
> 
> ; shouldn't you have the following?
> 
> ```
> src/payment-detection/
>   any/
>     any-to-near-proxy.ts
>     retrievers/
>       any-to-near-proxy.ts
> ```

@alexandre-abrioux  as discussed earlier, I think with our current knowledge both structures would be valid, and the current split clarifies the code a bit, we'll keep this file hierarchy for now.

PR #925, opened 2022-10-04
chore(currency): export AggregatorsMap

yomarion:
## Description of the changes
AggregatorsMap is used by CurrencyManager constructor, so it should be exported.


bertux (approved):
good renaming

AdamDowson (approved):

alexandre-abrioux:
@yomarion @benjlevesque Should the `AggregatorsMap` type be moved to the `types` package if it's used across packages? Asking because not sure if we strictly follow this logic for all types.

benjlevesque:
hmm maybe 🤔. I have mixed feelings about the `types` package... For some abstractions it's useful to share interfaces, but for types like this, I don't know. I'm not against it anyway...

alexandre-abrioux:
+1

PR #927, opened 2022-10-04
chore: correcting Near network naming

bertux:
## Description of the changes
Context: we initially thought aurora was the network ID for Near, but it's not, it's the name of the EVM-compatible NEAR sidechain.

Remove as many references as possible to aurora, keep just enough for retro-compatibility. Ideally, we'd replace aurora naming on request parsing only, on storage, so most libraries can ignore it.


julienbrg (approved):
If there isn't any open (unpaid) payment requests, it would make sense to completely give up the retro-compatibility.

bertux:
too big and risky PR, will split in stacked PRs to make it incremental.

PR #928, opened 2022-10-05
Fix/replace rinkeby with goerli

omarspx:
## Description of the changes

- Removed all ref to Rinkeby on our documentation
- Added Goerli to documentation
- Fixed docker image on docker/helm setup
- Fixed a few links


omarspx:
Replaced rinkeby-gateway with goerli (which doesn't exist yet). PR cannot be merged as of now.

PR #929, opened 2022-10-05
fix(eip1559): use fork

benjlevesque:
See https://github.com/RequestNetwork/requestNetwork/pull/765


KolevDarko (approved):

alexandre-abrioux (approved):

benjlevesque:
This hack is unnecessary at the moment. Major providers accept int as parameters. The cleaner solution is still https://github.com/RequestNetwork/requestNetwork/pull/765

PR #931, opened 2022-10-05
refactor(data-access): thegraph-data-access package

benjlevesque:
# Description
Introduce a new package to easily re-use thegraph data access. 
Move non-graph logic to their relevant places

# Usage

## Full 
```ts
import { RequestNetworkBase } from '@requestnetwork/request-client.js';
import { TheGraphDataAccess } from '@requestnetwork/thegraph-data-access';
import { Wallet } from 'ethers';
import { IpfsStorage, EthereumStorageEthers } from '@requestnetwork/ethereum-storage';

const signer = Wallet.createRandom();
const storage = new EthereumStorageEthers({
  ipfsStorage: new IpfsStorage({}),
  signer,
  network: 'private',
});
const dataAccess = new TheGraphDataAccess({
  graphql: { url: 'http://localhost:8000/subgraphs/name/RequestNetwork/request-storage' },
  network: 'private',
  storage,
});
const rn = new RequestNetworkBase({ dataAccess });

const { requestId } = await rn.createRequest(...)

const request = await rn.fromRequestId(requestId);

await request.accept(...)
```

## Read only
```ts
import { RequestNetworkBase } from '@requestnetwork/request-client.js';
import { TheGraphDataAccess } from '@requestnetwork/thegraph-data-access';
import { Wallet } from 'ethers';

const dataAccess = new TheGraphDataAccess({
  graphql: { url: 'http://localhost:8000/subgraphs/name/RequestNetwork/request-storage' },
  network: 'private',
});
const rn = new RequestNetworkBase({ dataAccess });

// Cannot create request
// await rn.createRequest()

const request = await rn.fromRequestId('...');
console.log(request.requestId);

// Cannot modify request
// await request.accept(...)
```


packages/request-node/src/thegraph-node.ts
```diff
@@ -25,12 +28,17 @@ export class TheGraphRequestNode extends RequestNodeBase {
...
       graphql: { url },
-      ipfsStorage,
+      storage,
       network,
-      signer,
-      logger: logger,
+      logger,
     });
 
     super(dataAccess, ipfsStorage, store, logger);
```
yomarion, line 44: Have you kept `ipfsStorage` on purpose? It looks weird so maybe add a comment to explain

packages/thegraph-data-access/.nycrc
```diff
@@ -0,0 +1,19 @@
+{
```
yomarion, line 1: Is this file versioned on purpose?

packages/thegraph-data-access/src/data-access.ts
```diff
@@ -8,20 +8,25 @@ import { Block } from '@requestnetwork/data-access';
...
-  ipfsStorage: StorageTypes.IIpfsStorage;
-  graphql: { url: string } & RequestInit;
-  signer: Signer;
+type TheGraphDataAccessBaseOptions = {
   network: string;
   logger?: LogTypes.ILogger;
   pendingStore?: PendingStore;
 };
 
+export type TheGraphDataAccessOptions = TheGraphDataAccessBaseOptions & {
```
yomarion, line 21:
```suggestion
export type TheGraphDataReadOptions = TheGraphDataAccessBaseOptions & {
```
Might make the difference with the other type more explicit?

yomarion: It's also consistent with the  `TheGraphDataRead` class name

packages/types/src/storage-types.ts
```diff
@@ -2,14 +2,21 @@ import { EventEmitter } from 'events';
...
+export interface IStorageRead {
+  initialize: () => Promise<void>;
   read: (dataId: string) => Promise<IEntry>;
   readMany: (dataIds: string[]) => Promise<IEntry[]>;
   getData: (options?: ITimestampBoundaries) => Promise<IEntriesWithLastTimestamp>;
   getIgnoredData: () => Promise<IEntry[]>;
+}
+
+/** Interface of the storage */
+export interface IStorage extends IStorageRead, IStorageWrite {
```
yomarion, line 19: 💅

yomarion (approved):

packages/thegraph-data-access/src/data-access.ts
benjlevesque: i removed datawrite and moved the `storage` to `TheGraphDataAccessOptions`, which now makes sense like this.

packages/thegraph-data-access/.nycrc
benjlevesque: yes, we have this on every package... nyc is the tool for test coverage

alexandre-abrioux (commented):
Nice 😀

packages/request-node/src/thegraph-node.ts
alexandre-abrioux: @benjlevesque I understand what @yomarion means (I think). I agree that it's weird to have an `ipfsStorage` object in the `RequestNodeBase` constructor, knowing that we pass down a `dataAccess` object that should already be a layer on top of IPFS. The issue is that it doesn't look possible to abstract—in the data-access layer—what `ipfsStorage` is used for in the request-node: the `/ipfsAdd` endpoint. This endpoint only persists documents on IPFS without persisting anything on-chain. I don't think it makes sense to add a generic method in data-access that only persist documents off-chain (like `persistTransactionOffChain` for instance). The data-access interface shouldn't contain methods that are tied to the underlying technology used for persisting IMO. Honestly, I think this endpoint is a security hole anyway, allowing users to persist documents on IPFS without persisting anything on-chain means we disregard spamming protection and the IPFS node could quickly become filled with fake documents. I'd propose we completely remove that endpoint, I don't think it is used anywhere, right?

benjlevesque: Yeah sorry we solved this in DM, my bad. 

this endpoint should be challenged for sure. It is related to the option I removed from the client. I believe it _should_ be possible to use 100% the client as a front-end library, but it doesn't need to be first-class like it is today

alexandre-abrioux: Ha! Get it, thanks!

PR #936, opened 2022-10-06
feat: minimal ERC20 deployment for Ronin chain

bertux:
## Description of the changes
Ronin chain is permissioned, so I've added a script to let Ronin team deploy minimal ERC20 payment contract by themselves.


bertux:
it was just to show differences, this branch should stay alone.

PR #943, opened 2022-10-09
chore(eth-fee-proxy-pn): update supported networks

alexandre-abrioux:


benjlevesque (approved):

packages/advanced-logic/src/extensions/payment-network/ethereum/fee-proxy-contract.ts
```diff
@@ -12,7 +12,20 @@ export default class EthereumFeeProxyPaymentNetwork<
   public constructor(
     extensionId: ExtensionTypes.ID = ExtensionTypes.ID.PAYMENT_NETWORK_ETH_FEE_PROXY_CONTRACT,
     currentVersion: string = CURRENT_VERSION,
-    public supportedNetworks: string[] = ['mainnet', 'rinkeby', 'goerli', 'private'],
+    public supportedNetworks: string[] = [
```
yomarion, line 15: Is it still relevant? Or shouldn't we map some kind of EVM-compatible pn?

yomarion (approved):

packages/advanced-logic/src/extensions/payment-network/ethereum/fee-proxy-contract.ts
alexandre-abrioux: I wondered why we did not compute this directly from the artifacts, since we have the `CURRENT_VERSION` constant in the PN could we just check that the contract is deployed for this version and that would be good enough?
https://github.com/RequestNetwork/requestNetwork/blob/a9e3f3e093843a1187f624be21833a2dbae0f87b/packages/smart-contracts/src/lib/artifacts/EthereumFeeProxy/index.ts#L58-L68

yomarion: It's not obvious, but it could be a way. The underlying question is: should we constrain library users and prevent the creation of requests that cannot be paid? I think that a warn could be enough.

bertux (approved):

olivier7delf (approved):

PR #945, opened 2022-10-10
refactor(advanced-logic): add evm-based PN

alexandre-abrioux:
This PR adds a composable `evm-based` payment network. It decouples the advanced-logic layer from the current state of deployed smart contracts. See discussion: https://github.com/RequestNetwork/requestNetwork/pull/943#discussion_r990978153

This PR, if merged, would supersede the following: https://github.com/RequestNetwork/requestNetwork/pull/943

### DIFF
--- a/packages/advanced-logic/src/extensions/payment-network/erc777/stream.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/erc777/stream.ts
@@ -1,25 +1,23 @@
 import { ExtensionTypes, RequestLogicTypes, TypesUtils } from '@requestnetwork/types';
 import ReferenceBasedPaymentNetwork from '../reference-based';
 import Utils from '@requestnetwork/utils';
+import EvmBasedPaymentNetwork from '../evm-based';
 const CURRENT_VERSION = '0.1.0';
 
 /**
  * Implementation of the payment network to pay in ERC777, including third-party fees payment, based on a reference provided to a proxy contract.
  */
 export default class Erc777StreamPaymentNetwork<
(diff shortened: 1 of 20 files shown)


benjlevesque (commented):
I wonder if this is even required... 
having the network hard coded in the library feels wrong; 

could we rely on smart-contracts for that? or be more permissive?

benjlevesque:
@alexandre-abrioux the advanced logic doesn't depend on smart-contracts. Maybe the solution would be to move the check on the network/version in the request-client? Or, maybe easier, in payment-detection. 
Yes I think that makes sense: the paymentNetwork factory can take care of that check I think...

packages/advanced-logic/src/extensions/payment-network/erc777/stream.ts
```diff
@@ -1,25 +1,23 @@
...
-      'xdai',
-      'mumbai',
-      'rinkeby',
-      'goerli',
-      'arbitrum-rinkeby',
-    ],
+    public readonly extensionId: ExtensionTypes.ID = ExtensionTypes.ID
+      .PAYMENT_NETWORK_ERC777_STREAM,
+    public readonly currentVersion: string = CURRENT_VERSION,
+    public readonly supportedNetworks: string[] = EvmBasedPaymentNetwork.EVM_NETWORKS,
```
KolevDarko, line 20: @alexandre-abrioux  ERC777 is not supported on all networks, most importantly mainnet is not supported

alexandre-abrioux: @KolevDarko I took a different approach in the end as advised by Benji. I'm working on it here: https://github.com/RequestNetwork/requestNetwork/pull/950

PR #946, opened 2022-10-10
chore: update celo explorer

alexandre-abrioux:


kevindavee (approved):

benjlevesque (approved):
just be careful that celoscan might require an api key, or have a stricter rate limiting

bertux (approved):

PR #950, opened 2022-10-11
refactor: check for contract deployment before creating pn

alexandre-abrioux:
supersedes https://github.com/RequestNetwork/requestNetwork/pull/945 and https://github.com/RequestNetwork/requestNetwork/pull/943

## Context

The `advanced-logic` layer is a stateless singleton that contains an instance of all payment networks (PN). When created (see [its class](https://github.com/RequestNetwork/requestNetwork/blob/312ce4a0f1826607edc60ccc3c6ca9e8c6bd1643/packages/advanced-logic/src/advanced-logic.ts)), it has no knowledge of what chain(s) it will be used on. It only needs an instance of the currency manager. It has no other dependency.

## Issue / Solution

It is a pain to manually update PN whenever we want them to support a new chain, especially when it could be done automatically via our smart contract artifacts. Without adding new dependencies on the `advanced-logic` object, this can be done from the `payment-detection` layer. PN should have zero knowledge of what chain(s) they support. Routing from the chain/version to the payment network should be done during the instantiation of the payment detector.

From this point:
  - `isValidAddress` in PNs now only accepts an address (no more `network` parameter), this means:
     - PN supporting EVM chains then support EVM addresses ([that was already the case](https://github.com/RequestNetwork/requestNetwork/blob/312ce4a0f1826607edc60ccc3c6ca9e8c6bd1643/packages/advanced-logic/src/extensions/payment-network/address-based.ts#L163))
     - PN can now only support one type of address, so all Near-related PN had to be split in two: one for Near mainnet and one for Near testnet, the same way we already had one for BTC and one for BTC-testnet
  - PN don't accept a `supportedNetworks` parameter anymore. Classes that extend `native-token` and `any-to-native` still have a `supportedNetworks` array as they need some kind of deeper level of routing. However, those are supper specific and not supposed to change that often once they are set.
  - everything Near related has been moved to its own directory in `advanced-logic`, like we did in `payment-detection`
  - the advanced-logic package typing has been updated (removed `any` whenever possible)

Ideally, those changes should not impact the behavior of the library when used for payment detection.


packages/advanced-logic/src/extensions/content-data.ts
```diff
@@ -9,11 +9,8 @@ const CURRENT_VERSION = '0.1.0';
 export default class ContentDataExtension<
   TCreationParameters extends ExtensionTypes.ContentData.ICreationParameters = ExtensionTypes.ContentData.ICreationParameters,
 > extends AbstractExtension<TCreationParameters> {
-  public constructor(
-    public extensionId: ExtensionTypes.ID = ExtensionTypes.ID.CONTENT_DATA,
-    public currentVersion: string = CURRENT_VERSION,
-  ) {
-    super(ExtensionTypes.TYPE.CONTENT_DATA, extensionId, currentVersion);
+  public constructor() {
+    super(ExtensionTypes.TYPE.CONTENT_DATA, ExtensionTypes.ID.CONTENT_DATA, CURRENT_VERSION);
```
alexandre-abrioux, line 13: Type of change made for classes that never get extended

packages/advanced-logic/src/extensions/payment-network/any-to-native.ts
```diff
@@ -1,14 +1,14 @@
 import { FeeReferenceBasedPaymentNetwork } from './fee-reference-based';
 import { ExtensionTypes, RequestLogicTypes } from '@requestnetwork/types';
-import { InvalidPaymentAddressError } from './address-based';
+import { InvalidPaymentAddressError, UnsupportedNetworkError } from './address-based';
 
 export default abstract class AnyToNativeTokenPaymentNetwork extends FeeReferenceBasedPaymentNetwork {
-  public constructor(
+  protected constructor(
```
alexandre-abrioux, line 6: abstract class constructors should be protected

packages/payment-detection/src/near/near-conversion-detector.ts
```diff
@@ -22,27 +17,7 @@ const CONTRACT_ADDRESS_MAP: IProxyContractVersion = {
...
-  public constructor({
-    advancedLogic,
-    currencyManager,
-  }: {
-    advancedLogic: AdvancedLogicTypes.IAdvancedLogic;
-    currencyManager: ICurrencyManager;
-  }) {
-    super(
-      PaymentTypes.PAYMENT_NETWORK_ID.ANY_TO_NATIVE,
-      advancedLogic.extensions.anyToNativeToken[0],
```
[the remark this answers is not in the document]
benjlevesque: good news

packages/payment-detection/src/erc20/fee-proxy-contract.ts
```diff
@@ -23,9 +23,11 @@ export abstract class ERC20FeeProxyPaymentDetectorBase<
   TPaymentEventParameters extends PaymentTypes.IERC20FeePaymentEventParameters,
 > extends FeeReferenceBasedDetector<TExtension, TPaymentEventParameters> {
   /**
+   * @param paymentNetworkId
    * @param extension The advanced logic payment network extensions
+   * @param _currencyManager
```
benjlevesque, line 28:
```suggestion
   * @param currencyManager
```
not sure why we need these `@param` comments though ^^

packages/payment-detection/src/native-token-detector.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ReferenceBasedDetectorOptions } from './types';
+
+/**
+ * Handle payment detection for native token payment
+ */
+export abstract class NativeTokenPaymentDetector extends ReferenceBasedDetector<
+  ExtensionTypes.PnReferenceBased.IReferenceBased,
+  PaymentTypes.IETHPaymentEventParameters
+> {
+  public constructor({
```
benjlevesque, line 13:
```suggestion
  protected constructor({
```

packages/payment-detection/src/any-to-native-detector.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ReferenceBasedDetectorOptions } from './types';
+
+/**
+ * Handle payment detection for native token payment with conversion
+ */
+export abstract class AnyToNativeDetector extends AnyToAnyDetector<
+  ExtensionTypes.PnAnyToEth.IAnyToEth,
+  PaymentTypes.IETHPaymentEventParameters
+> {
+  public constructor({
```
benjlevesque, line 13:
```suggestion
  protected constructor({
```

benjlevesque (approved):
amazing! great work. Happy to see those `supportedNetworks` go away 👍

packages/advanced-logic/src/extensions/payment-network/near/any-to-near.ts
```diff
@@ -21,28 +23,10 @@ export default class AnyToNearPaymentNetwork extends AnyToNativeTokenPaymentNetw
...
-      case undefined:
-        return this.isValidMainnetAddress(address) || this.isValidTestnetAddress(address);
-      default:
-        throw new UnsupportedNetworkError(networkName, this.supportedNetworks);
-    }
-  }
-
-  private isValidMainnetAddress(address: string): boolean {
+  protected isValidAddress(address: string): boolean {
     return this.isValidAddressForSymbolAndNetwork(address, 'NEAR', 'aurora');
```
yomarion, line 27: I don't understand why we get rid of the testnet addresses validation. FYI they have different validation rules compared to mainnet.

packages/payment-detection/src/erc20/address-based.ts
```diff
@@ -4,11 +4,9 @@ import {
   PaymentTypes,
   RequestLogicTypes,
 } from '@requestnetwork/types';
-import { BalanceError } from '../balance-error';
 import Erc20InfoRetriever from './address-based-info-retriever';
 
 import { PaymentDetectorBase } from '../payment-detector-base';
-const supportedNetworks = ['mainnet', 'rinkeby', 'goerli', 'private'];
```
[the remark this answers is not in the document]
yomarion: Yes, I agree. If there was such a limitation, it would be in the info-retriever. 

Maybe throwing if no (default) provider exists for the network, within `getTransferEvents()`, would maintain clarity for what works and what does not? And the next step depends on what this provider returns if we query a wider range of blocks than it allows, specifically, if it allows `{fromBlock: 0, toBlock: 'latest'}`.

As we expect the result to be incomplete in most cases, while `getTransferEvents()` should be TheGraph-based, shouldn't we add a `console.warn` for address-based info retrieval?

packages/advanced-logic/src/advanced-logic.ts
```diff
@@ -63,10 +65,11 @@ export default class AdvancedLogic implements AdvancedLogicTypes.IAdvancedLogic
       erc777Stream: new Erc777Stream(),
       feeProxyContractEth: new FeeProxyContractEth(),
       anyToEthProxy: new AnyToEthProxy(currencyManager),
-      nativeToken: [new NearNative()],
-      anyToNativeToken: [new AnyToNear(currencyManager)],
+      nativeToken: [new NearNative(), new NearTestnetNative()],
+      anyToNativeToken: [new AnyToNear(currencyManager), new AnyToNearTestnet(currencyManager)],
```
yomarion, line 69: If we go down that line, can we add `anyToEthProxy` and `feeProxyContractEth` to these 2 generic PNs? And deprecate the "separate" use of these extensions? It does not add anything while we just have Near.

yomarion (approved):
Really clean, well done.

packages/payment-detection/src/any-to-native-detector.ts
alexandre-abrioux: done in https://github.com/RequestNetwork/requestNetwork/pull/950/commits/b3916cac36caa8784ae927e51f68410bba53f957

alexandre-abrioux: I had to add some code in classes that extend it:
```ts
  constructor(args: NativeDetectorOptions) {
    super(args);
  }
```

packages/payment-detection/src/native-token-detector.ts
alexandre-abrioux: done, see comment for `any-to-native-detector.ts`

packages/payment-detection/src/erc20/fee-proxy-contract.ts
alexandre-abrioux: https://github.com/RequestNetwork/requestNetwork/pull/950/commits/e57b267a1ac0b14e51f44ff95fe90b465a280409

packages/advanced-logic/src/extensions/payment-network/near/any-to-near.ts
alexandre-abrioux: Near testnet now has its separate class: https://github.com/RequestNetwork/requestNetwork/blob/e57b267a1ac0b14e51f44ff95fe90b465a280409/packages/advanced-logic/src/extensions/payment-network/near/any-to-near-testnet.ts#L17

packages/advanced-logic/src/advanced-logic.ts
alexandre-abrioux: I agree! We should deprecate `any-to-eth-proxy` and instead use `any-to-native-token`. Not sure however that `eth-fee-proxy-contract` and `native-token` are compatible as the latest doesn't support fees as far as I know. I'll keep this PR a refactor without changing payment networks, but let's keep this in mind for the next steps.

packages/payment-detection/src/payment-network-factory.ts
```diff
@@ -103,36 +103,50 @@ export class PaymentNetworkFactory {
...
       advancedLogic: this.advancedLogic,
       currencyManager: this.currencyManager,
       ...this.options,
     });
+
+    if (detector.extension && 'getDeploymentInformation' in detectorClass) {
+      // this throws when the contract isn't deployed and was mandatory for payment detection
+      (detectorClass as ContractBasedDetector).getDeploymentInformation(
+        network,
+        detector.extension.currentVersion,
```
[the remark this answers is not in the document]
alexandre-abrioux: Good catch! Done in https://github.com/RequestNetwork/requestNetwork/pull/950/commits/e57154b5a0da5ca446644ee93f98c2fed75da95d. By default we use the latest version for validation, but we also have the option to pass down the proper version.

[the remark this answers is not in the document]
benjlevesque: good job

PR #955, opened 2022-10-13
feat: add near-testnet naming conversion

bertux:
## Description of the changes
Accept near-testnet network name by mapping it to aurora-test in all actions

### DIFF
--- a/packages/advanced-logic/src/extensions/payment-network/any-to-near.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/any-to-near.ts
@@ -4,7 +4,7 @@ import { UnsupportedNetworkError } from './address-based';
 import AnyToNativeTokenPaymentNetwork from './any-to-native';
 
 const CURRENT_VERSION = '0.1.0';
-const supportedNetworks = ['aurora', 'aurora-testnet'];
+const supportedNetworks = ['aurora', 'aurora-testnet', 'near-testnet'];
 
 export default class AnyToNearPaymentNetwork extends AnyToNativeTokenPaymentNetwork {
   public constructor(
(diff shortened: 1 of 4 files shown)


packages/advanced-logic/src/extensions/payment-network/any-to-near.ts
```diff
@@ -57,6 +58,11 @@ export default class AnyToNearPaymentNetwork extends AnyToNativeTokenPaymentNetw
     if (!extensionAction.parameters.network || extensionAction.parameters.network.length === 0) {
       throw Error('network is required');
     }
+    // FIXME: Needed during naming migration from aurora to near
```
yomarion, line 61:
```suggestion
    // FIXME: Remove when the mistaken network name 'aurora' is gone
```

yomarion: + Other comments

packages/advanced-logic/src/extensions/payment-network/near-native.ts
```diff
@@ -27,6 +27,7 @@ export default class NearNativePaymentNetwork extends NativeTokenPaymentNetwork
       case 'aurora':
         return this.isValidMainnetAddress(address);
       case 'aurora-testnet':
+      case 'near-testnet':
```
yomarion, line 30: I think if you add this case we should increase the CURRENT_VERSION. `near-testnet` is not valid in `0.2.0`, both are valids in `0.3.0` and we can then invalidate `aurora` in `0.4.0`.

yomarion (commented):
Why do you map aurora to near for any-to-native but support near on near-native ?

packages/advanced-logic/src/extensions/payment-network/near-native.ts
bertux: that's why to avoid that I will add the temporary mapping as for any-to-native

bertux:
replaced by https://github.com/RequestNetwork/requestNetwork/pull/956

PR #956, opened 2022-10-14
chore: support 'near-testnet' aside 'aurora-testnet'

bertux:
## Description of the changes
This is a first temporary step towards replacing all the mistaken mentions of `aurora`.

This step is non-breaking and does not change the way we create or store requests.
It only focuses on testnet.

Target design after the whole change:
- Bump the request-logic version, and the concerned extension versions. These versions will only support `near` and `near-testnet`
- Keep supporting `aurora` and `aurora-testnet` for earlier versions of the extensions and request-logic
- The support for `aurora` and `aurora-testnet` should impact a small portion of the code, where we read the storage layer.


benjlevesque (approved):
any reason not to support near in the same PR?

KolevDarko (approved):

yomarion:
> any reason not to support near in the same PR?

@benjlevesque , we could, we just want to rollout the whole plan for testnet first, for more peace of mind.

PR #958, opened 2022-10-14
chore: adding tNEAR for near-testnet

bertux:
## Description of the changes
need new native token declared on new chain to be able to test the integration in currency-api and then from end to end

### DIFF
--- a/packages/currency/src/native.ts
+++ b/packages/currency/src/native.ts
@@ -61,12 +61,6 @@ export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { na
       symbol: 'BNB',
       decimals: 18,
       name: 'BNB',
-      network: 'bsctest',
-    },
-    {
-      symbol: 'BNB',
-      decimals: 18,
-      name: 'BNB',
       network: 'bsc',
     },
     {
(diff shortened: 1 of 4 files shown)


packages/currency/src/native.ts
```diff
@@ -57,12 +57,6 @@ export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { na
       name: 'Fantom',
       network: 'fantom',
     },
-    {
-      symbol: 'BNB',
-      decimals: 18,
-      name: 'BNB',
-      network: 'bsctest',
-    },
```
alexandre-abrioux, line 65: I don't understand this change

alexandre-abrioux (commented):
Why the need to rename this, what's wrong with `NEAR-testnet`? I can't find any documentation on this new `tNEAR` symbol online, that's why I'm asking this.

KolevDarko (approved):

packages/currency/src/native.ts
bertux: it's just cleaning because it's not used.

alexandre-abrioux (approved):
Saw in DM: it's to remove confusion between the new near-testnet chain (renamed from aurora-testnet) and the NEAR-testnet currency.

bertux:
superseded by https://github.com/RequestNetwork/currency-api/pull/236

PR #964, opened 2022-10-18
fix: keep the same symbol name

bertux:
## Description of the changes
Come back to same symbol name for native token of near-testnet than aurora-testnet it will replace.


bertux:
Still need to update the hash of the native token in near-testnet aggregator.

olivier7delf (approved):

PR #973, opened 2022-10-26
feat: approval amount for escrow

leoslr:
## Description of the changes

There was no way to set the approval amount for escrow transaction. This fixes it.
I'm finishing a PR on the transaction-builder on portal where the approval amount can be defined. It's better if it also applies to escrow for consistency

### DIFF
--- a/packages/payment-processor/src/payment/erc20-escrow-payment.ts
+++ b/packages/payment-processor/src/payment/erc20-escrow-payment.ts
@@ -26,10 +26,16 @@ export function prepareErc20EscrowApproval(
   request: ClientTypes.IRequestData,
   paymentTokenAddress: string,
   signerOrProvider: providers.Provider | Signer = getProvider(),
+  amount?: BigNumber,
   overrides?: ITransactionOverrides,
 ): IPreparedTransaction {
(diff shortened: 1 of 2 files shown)


packages/payment-processor/src/payment/erc20-escrow-payment.ts
```diff
@@ -26,10 +26,16 @@ export function prepareErc20EscrowApproval(
   request: ClientTypes.IRequestData,
   paymentTokenAddress: string,
   signerOrProvider: providers.Provider | Signer = getProvider(),
+  amount?: BigNumber,
```
benjlevesque, line 29: as discussed with @olivier7delf , I believe we should stop adding positionnal parameters to payment-processor methods. There should be always the same number of parameters, in the same order, and one of these should be the options, dependant on the PN. 

So typically, `paymentTokenAddress` and `amount` could be grouped in the same param.

benjlevesque (approved):

packages/payment-processor/src/payment/erc20-escrow-payment.ts
leoslr: Yes I agree. we'll need to make a refacto about this at some point. I kind of did it a while back but with the new features we kind of diverged from it.

PR #980, opened 2022-11-03
fix: update all chainlink conversion path version

leoslr:
## Description of the changes

On Fantom and Matic, we have two versions of the chainlinkConversionPath contract.
Each of them are used by different contracts.
However the toolbox only updates the latest version of the chailinkConverPath contract. 

This aims to update all versions, this way we're sure that the new aggregators that we add are available on all versions.


benjlevesque (changes requested):
As far as I remember, we need all `0.1.0` artifacts to be in `0.2.0`. And I'm not talking about the toolbox... The Current version will be the highest. So if you omit mainnet & others, I think this can cause issues.

leoslr:
> As far as I remember, we need all `0.1.0` artifacts to be in `0.2.0`. And I'm not talking about the toolbox... The Current version will be the highest. So if you omit mainnet & others, I think this can cause issues.

@benjlevesque  You're right, I did not remember about this. then for `mainnet` I'll put the same address in `0.2.0` as in `0.1.0.`

benjlevesque (approved):

yomarion (approved):

PR #983, opened 2022-11-07
feat: optimism deployment

leoslr:
## Description of the changes

- Deploy request contracts on Optimism.
- Update the plugin `hardhat-etherscan` to add custom networks.

Fixes several issues:
- Introduces a delay between each contract verification, otherwise, the API limit of 5 calls/s is reached.
- Remove the nonce parameter in the `script-create-2` admin tasks. It was causing issues since the transactions were conditional.
- Fixes the conditions in the `script-create-2` admin tasks. They were wrong.

### DIFF
--- a/packages/smart-contracts/hardhat.config.ts
+++ b/packages/smart-contracts/hardhat.config.ts
@@ -125,6 +125,11 @@ export default {
       chainId: 43114,
       accounts,
     },
+    optimism: {
+      url: url('optimism'),
+      chainId: 10,
+      accounts,
+    },
   },
   etherscan: {
     apiKey: {
--- a/packages/smart-contracts/hardhat.config.ts
+++ b/packages/smart-contracts/hardhat.config.ts
@@ -151,11 +151,22 @@ export default {
       // xdai
       xdai: process.env.GNOSISSCAN_API_KEY,
       // other networks don't need an API key, but you still need
+      optimism: process.env.OPTIMISM_API_KEY,
       // to specify one; any string placeholder will work
       sokol: 'api-key',
       aurora: 'api-key',
       auroraTestnet: 'api-key',
     },
+    customChains: [
+      {
+        network: 'optimism',
+        chainId: 10,
+        urls: {
+          apiURL: 'https://api-optimistic.etherscan.io/api',
+          browserURL: 'https://optimistic.etherscan.io/',
+        },
+      },
+    ],
   },
   typechain: {
     outDir: 'src/types',
--- a/packages/smart-contracts/scripts-create2/constructor-args.ts
+++ b/packages/smart-contracts/scripts-create2/constructor-args.ts
@@ -20,9 +20,11 @@ export const getConstructorArgs = (contract: string, network?: string): string[]
       return [
         '0x0000000000000000000000000000000000000000',
         '0x0000000000000000000000000000000000000000',
+        // FIXME: This is not right the NativeTokenHash is not the same accross all networks
+        //        It should be set to 0x0, and then updated through setup functions
+        //        These functions needs to be implemented in the contract before it can deployed using xdeployer
         '0x39e19aa5b69466dfdc313c7cda37cb2a599015cd',
       ];
-      // TODO setupETHConversionProxy
     }
     case 'ERC20SwapToConversion': {
       return [getAdminWalletAddress(contract)];
--- a/packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
+++ b/packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
@@ -11,14 +11,13 @@ export const BATCH_FEE = 3;
 export const updateChainlinkConversionPath = async (
   contract: any,
   network: string,
-  nonce: number,
   gasPrice: BigNumber,
+  version = undefined,
 ): Promise<void> => {
   const currentChainlinkAddress = await contract.chainlinkConversionPath();
-  const chainlinkConversionPathAddress = chainlinkConversionPath.getAddress(network, '0.1.0');
+  const chainlinkConversionPathAddress = chainlinkConversionPath.getAddress(network, version);
   if (currentChainlinkAddress !== chainlinkConversionPathAddress) {
     await contract.updateConversionPathAddress(chainlinkConversionPathAddress, {
-      nonce: nonce,
       gasPrice: gasPrice,
     });
   }
@@ -27,46 +26,35 @@ export const updateChainlinkConversionPath = async (
 export const updateSwapRouter = async (
   contract: any,
   network: string,
-  nonce: number,
   gasPrice: BigNumber,
 ): Promise<void> => {
   const currentSwapRouter = await contract.swapRouter();
   if (currentSwapRouter !== uniswapV2RouterAddresses[network]) {
     await contract.setRouter(uniswapV2RouterAddresses[network], {
-      nonce: nonce,
       gasPrice: gasPrice,
     });
   }
 };
 
-export const updateRequestSwapFees = async (
-  contract: any,
-  nonce: number,
-  gasPrice: BigNumber,
-): Promise<void> => {
+export const updateRequestSwapFees = async (contract: any, gasPrice: BigNumber): Promise<void> => {
   const currentFees = await contract.requestSwapFees();
   if (currentFees !== REQUEST_SWAP_FEES) {
-    await contract.updateRequestSwapFees(REQUEST_SWAP_FEES, { nonce: nonce, gasPrice: gasPrice });
+    await contract.updateRequestSwapFees(REQUEST_SWAP_FEES, { gasPrice: gasPrice });
   }
 };
 
-export const updateBatchPaymentFees = async (
-  contract: any,
-  nonce: number,
-  gasPrice: BigNumber,
-): Promise<void> => {
+export const updateBatchPaymentFees = async (contract: any, gasPrice: BigNumber): Promise<void> => {
   const currentFees = await contract.batchFee();
   if (currentFees !== BATCH_FEE) {
-    // Log is useful to have a direct view on was is being updated
+    // Log is useful to have a direct view on what is being updated
     console.log(`currentFees: ${currentFees.toString()}, new fees: ${BATCH_FEE}`);
-    await contract.setBatchFee(BATCH_FEE, { nonce: nonce, gasPrice: gasPrice });
+    await contract.setBatchFee(BATCH_FEE, { gasPrice: gasPrice });
   }
 };
 
 export const updatePaymentErc20FeeProxy = async (
   contract: any,
   network: string,
-  nonce: number,
   gasPrice: BigNumber,
 ): Promise<void> => {
   const erc20FeeProxy = artifacts.erc20FeeProxyArtifact;
@@ -74,7 +62,6 @@ export const updatePaymentErc20FeeProxy = async (
   const currentAddress = await contract.paymentErc20FeeProxy();
   if (currentAddress !== erc20FeeProxyAddress) {
     await contract.setPaymentErc20FeeProxy(erc20FeeProxyAddress, {
-      nonce: nonce,
       gasPrice: gasPrice,
     });
   }
@@ -83,7 +70,6 @@ export const updatePaymentErc20FeeProxy = async (
 export const updatePaymentEthFeeProxy = async (
   contract: any,
   network: string,
-  nonce: number,
   gasPrice: BigNumber,
 ): Promise<void> => {
   const ethereumFeeProxy = artifacts.ethereumFeeProxyArtifact;
@@ -91,7 +77,6 @@ export const updatePaymentEthFeeProxy = async (
   const currentAddress = await contract.paymentEthFeeProxy();
   if (currentAddress !== ethereumFeeProxyAddress) {
     await contract.setPaymentEthFeeProxy(ethereumFeeProxyAddress, {
-      nonce: nonce,
       gasPrice: gasPrice,
     });
   }
--- a/packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
+++ b/packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
@@ -37,15 +37,16 @@ export const updateSwapRouter = async (
 };
 
 export const updateRequestSwapFees = async (contract: any, gasPrice: BigNumber): Promise<void> => {
-  const currentFees = await contract.requestSwapFees();
-  if (currentFees !== REQUEST_SWAP_FEES) {
+  const currentFees: BigNumber = await contract.requestSwapFees();
+  if (!currentFees.eq(REQUEST_SWAP_FEES)) {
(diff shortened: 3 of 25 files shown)


packages/smart-contracts/src/lib/artifacts/ERC20EscrowToPay/index.ts
```diff
@@ -33,6 +33,11 @@ export const erc20EscrowToPayArtifact = new ContractArtifact<ERC20EscrowToPay>(
           address: '0x4BA012eae4d64da79Bd6bcdBa366803fCe701A4C',
           creationBlockNumber: 18086337,
         },
+        // FIXME: Address changed due to contract metadata change since last deployment
```
benjlevesque, line 36: I don't understand what there is to fix?

packages/smart-contracts/scripts-create2/contract-setup/setupERC20SwapToConversion.ts
```diff
@@ -23,29 +23,26 @@ export const setupERC20SwapToConversion = async (
...
-      const signer = wallet.connect(provider);
-      const ERC20SwapToConversionConnected = await ERC20SwapToConversionContract.connect(signer);
-      const adminNonce = await signer.getTransactionCount();
-      const gasPrice = await provider.getGasPrice();
+      try {
+        let provider;
+        if (network === 'celo') {
+          provider = utils.getCeloProvider();
+        } else {
+          provider = utils.getDefaultProvider(network);
```
benjlevesque, line 31: shouldn't `getDefaultProvider` handle this condition?

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
```diff
@@ -23,24 +23,26 @@ export const setupBatchPayments = async (
...
-        provider = utils.getCeloProvider();
-      } else {
-        provider = utils.getDefaultProvider(network);
-      }
-      const wallet = new hre.ethers.Wallet(hre.config.xdeploy.signer, provider);
-      const signer = wallet.connect(provider);
-      const batchPaymentConnected = await batchPaymentContract.connect(signer);
-      const adminNonce = await signer.getTransactionCount();
-      const gasPrice = await provider.getGasPrice();
+      try {
```
benjlevesque, line 26: see next file

packages/smart-contracts/scripts-create2/constructor-args.ts
```diff
@@ -20,9 +20,11 @@ export const getConstructorArgs = (contract: string, network?: string): string[]
       return [
         '0x0000000000000000000000000000000000000000',
         '0x0000000000000000000000000000000000000000',
+        // FIXME: This is not right the NativeTokenHash is not the same accross all networks
```
benjlevesque, line 23: this scares me a bit... don't you think there should be something in the code preventing wrong execution? A FIXME will most likely be forgotten

packages/smart-contracts/scripts-create2/contract-setup/setupErc20ConversionProxy.ts
```diff
@@ -0,0 +1,43 @@
...
+    hre.config.xdeploy.networks.map(async (network) => {
+      try {
+        let provider;
+        if (network === 'celo') {
+          provider = utils.getCeloProvider();
+        } else {
+          provider = utils.getDefaultProvider(network);
+        }
+        const wallet = new hre.ethers.Wallet(hre.config.xdeploy.signer, provider);
+        const signer = wallet.connect(provider);
```
benjlevesque, line 30: isn't this equivalent?
```suggestion
        const signer = new hre.ethers.Wallet(hre.config.xdeploy.signer).connect(provider);
```

benjlevesque (approved):
Not enough knowledge to challenge the deployment part...

packages/smart-contracts/src/lib/artifacts/ERC20EscrowToPay/index.ts
leoslr: Nothing to fix, just a comment to explain why the address is not the same as for other networks.

packages/smart-contracts/scripts-create2/contract-setup/setupERC20SwapToConversion.ts
leoslr: If I remember correctly, when we introduce the celo specific case with Yoann we wanted to make it part of the `getDefaultProvider` but there were some blockers.

packages/smart-contracts/src/lib/artifacts/ERC20EscrowToPay/index.ts
benjlevesque: then remove `FIXME`?

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
leoslr: We do not want to stop the execution. 
The admin tasks happen on all networks, so if it fails on one we don't want to prevent them to run on others.

If the admin task fails, it will most likely be due to missing funds. And a message is displayed in the console. 
The advantage is that in the event of any error, the person deploying the contracts just has to start the script again:
- Deployments already made will be skipped,
- Deployments that failed will be run,
- Admin tasks already performed will be skipped,
- Admin tasks that failed will be run

packages/smart-contracts/scripts-create2/contract-setup/setupErc20ConversionProxy.ts
leoslr: Right.
I've made some factorization.

packages/smart-contracts/scripts-create2/constructor-args.ts
leoslr: The scripts deploys the contract specified in `script-create2/utils`. `EthConversionProxy` is commented there with a FIXME, so I think it's safe.

packages/smart-contracts/hardhat.config.ts
```diff
@@ -146,11 +151,22 @@ export default {
       // xdai
       xdai: process.env.GNOSISSCAN_API_KEY,
       // other networks don't need an API key, but you still need
+      optimism: process.env.OPTIMISM_API_KEY,
```
alexandre-abrioux, line 154:
```suggestion
      // optimism
      optimism: process.env.OPTIMISM_API_KEY,
      // other networks don't need an API key, but you still need
```

packages/smart-contracts/scripts-create2/constructor-args.ts
olivier7delf: Maybe raise an error here => nobody can do the error if there is a fast deployment.

packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
```diff
@@ -11,88 +13,99 @@ export const BATCH_FEE = 3;
...
-  nonce: number,
-  gasPrice: BigNumber,
-): Promise<void> => {
-  const currentFees = await contract.requestSwapFees();
-  if (currentFees !== REQUEST_SWAP_FEES) {
-    await contract.updateRequestSwapFees(REQUEST_SWAP_FEES, { nonce: nonce, gasPrice: gasPrice });
+export const updateRequestSwapFees = async (contract: any, gasPrice: BigNumber): Promise<void> => {
+  const currentFees: BigNumber = await contract.requestSwapFees();
+  if (!currentFees.eq(REQUEST_SWAP_FEES)) {
+    console.log(`currentFees: ${currentFees.toString()}, new fees: ${REQUEST_SWAP_FEES}`);
```
olivier7delf, line 46: Great to have the logs.
It might be better to put it just after the tx.wait(1), in case there is an error. If we see it si logged, it means it is really deployed

packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
```diff
@@ -11,88 +13,99 @@ export const BATCH_FEE = 3;
...
-  gasPrice: BigNumber,
-): Promise<void> => {
-  const currentFees = await contract.batchFee();
-  if (currentFees !== BATCH_FEE) {
-    // Log is useful to have a direct view on was is being updated
+export const updateBatchPaymentFees = async (contract: any, gasPrice: BigNumber): Promise<void> => {
+  const currentFees: BigNumber = await contract.batchFee();
+  if (!currentFees.eq(BATCH_FEE)) {
+    // Log is useful to have a direct view on what is being updated
     console.log(`currentFees: ${currentFees.toString()}, new fees: ${BATCH_FEE}`);
```
olivier7delf, line 56: same for the log

packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
```diff
@@ -11,88 +13,99 @@ export const BATCH_FEE = 3;
...
+  version = undefined,
 ): Promise<void> => {
   const currentChainlinkAddress = await contract.chainlinkConversionPath();
-  const chainlinkConversionPathAddress = chainlinkConversionPath.getAddress(network, '0.1.0');
+  const chainlinkConversionPathAddress = chainlinkConversionPath.getAddress(network, version);
   if (currentChainlinkAddress !== chainlinkConversionPathAddress) {
-    await contract.updateConversionPathAddress(chainlinkConversionPathAddress, {
-      nonce: nonce,
+    const tx = await contract.updateConversionPathAddress(chainlinkConversionPathAddress, {
       gasPrice: gasPrice,
```
alexandre-abrioux, line 23: Note for later: `gasPrice` is legacy and more expensive most of the time, we should migrate to EIP-1559 gas fee.

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
olivier7delf: +1 Léo
I got this issue and I added it myself to deploy the batch on every network easily.

packages/smart-contracts/scripts-create2/contract-setup/setupErc20ConversionProxy.ts
```diff
@@ -0,0 +1,38 @@
+import { erc20ConversionProxy } from '../../src/lib';
+import { HardhatRuntimeEnvironmentExtended } from '../types';
+import {
+  getSignerAndGasPrice,
+  updateChainlinkConversionPath,
+  updatePaymentErc20FeeProxy,
+} from './adminTasks';
+
+/**
+ * Updates the values of the batch fees of the BatchPayments contract, if needed
```
olivier7delf, line 10: it updates the erc20Fee proxy used by erc20conversion proxy, and the chainlink

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
alexandre-abrioux: What if there is a miss config that will generate errors on all networks for all contracts? Won't this waste a lot of gas? In case of an error, can we stop and only save the artifacts for what's already been deployed, and then we try again after fixing the mistake:
- what's already been deployed will be skipped
- what's missing will be run and caught back

; do you think that's feasible?

packages/smart-contracts/scripts/deploy-payments.ts
```diff
@@ -298,18 +297,16 @@ export async function deployAllPaymentContracts(
...
     // #region MAIN - Deployments
 
-    // Batch 1
-    await jumpToNonce(args, hre, 1);
-    const { address: erc20FeeProxyAddress } = await runEasyDeployment({
-      contractName: 'ERC20FeeProxy',
-      artifact: erc20FeeProxyArtifact,
-      nonceCondition: 1,
-    });
+    // Batch 1 - REMOVED -> Deployment CREATE2
```
olivier7delf, line 300: Is it meaningful to say "Step 1" instead of "Batch 1"?

packages/smart-contracts/scripts-create2/deploy.ts
```diff
@@ -52,14 +54,25 @@ export const deployWithCreate2FromList = async (
...
-      case 'EthConversionProxy':
       case 'ERC20FeeProxy':
-      case 'Erc20ConversionProxy': {
+      case 'EthereumFeeProxy': {
         const constructorArgs = getConstructorArgs(contract);
         await deployOneWithCreate2({ contract, constructorArgs }, hre);
         break;
       }
+      case 'EthConversionProxy': {
+        // FIXME: not deployable through xdeployer yet. Check FIXME in getConstrucotrArgs function
```
alexandre-abrioux, line 64: Just to be safe:
```suggestion
        // FIXME: not deployable through xdeployer yet. Check FIXME in getConstrucotrArgs function
        throw new Error('EthConversionProxy not deployable through xdeployer yet');
```

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
```diff
@@ -23,24 +23,18 @@ export const setupBatchPayments = async (
   );
   await Promise.all(
     hre.config.xdeploy.networks.map(async (network) => {
```
olivier7delf, line 25: I am having a concern regarding the mapping, I used a loop on my branch to make logs easier to read:
logs:
network: mainnent
update ERC20 address into blabla
update Eth address into blabla
...
network: matic
update ERC20 address into blabla
update Eth address into blabla
...

With a mapping, is it still readable?

olivier7delf (approved):
Great, it will make deployment cleaner and easier to use!

KolevDarko (approved):

packages/smart-contracts/src/lib/artifacts/Erc20SwapConversion/index.ts
```diff
@@ -79,6 +79,11 @@ export const erc20SwapConversionArtifact = new ContractArtifact<ERC20SwapToConve
           address: '0x3b4837C9F4A606b71e61FD56Db6241781194df92',
           creationBlockNumber: 21837165,
         },
+        // FIXME: Deployment address changed due to an update of the contract metadata
```
alexandre-abrioux, line 82:
```suggestion
        // Deployment address changed due to an update of the contract metadata
```

alexandre-abrioux (approved):

packages/smart-contracts/scripts-create2/contract-setup/setupBatchPayments.ts
leoslr: It won't waste gas since if the transaction is doomed to fail on all networks, it won't be fired because gas estimation won't work.
However it might lead to wasting gas sometimes, it should be small amounts since most chains are cheap aside mainnet.

The person deploying the contracts can stop the script anytime, save the artifacts, and restart the script when he wants to it will automatically catch up where he stopped.

packages/smart-contracts/scripts-create2/contract-setup/adminTasks.ts
leoslr: I'll create a tech discovery ticket to not forget about that

PR #985, opened 2022-11-08
feat: optimism config

leoslr:
## Description of the changes

Adds configuration for optimism network.
This PR is complementary to https://github.com/RequestNetwork/requestNetwork/pull/983

### DIFF
--- a/packages/currency/src/conversion-aggregators.ts
+++ b/packages/currency/src/conversion-aggregators.ts
@@ -48,6 +48,7 @@ const noConversionNetworks: AggregatorsMap = {
   xdai: {},
   avalanche: {},
   bsc: {},
+  optimism: {},
 };
 
 /**
--- a/packages/currency/src/erc20/networks/optimism.ts
+++ b/packages/currency/src/erc20/networks/optimism.ts
@@ -1,0 +1,20 @@
+import { TokenMap } from './types';
+
+// List of the supported bsc network tokens
+export const supportedOptimismERC20: TokenMap = {
+  '0x7F5c764cBc14f9669B88837ca1490cCa17c31607': {
+    name: 'USD Coin',
+    symbol: 'USDC',
+    decimals: 6,
+  },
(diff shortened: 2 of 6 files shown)


packages/currency/src/conversion-aggregators.ts
```diff
@@ -48,6 +48,7 @@ const noConversionNetworks: AggregatorsMap = {
   xdai: {},
   avalanche: {},
   bsc: {},
+  optimism: {},
```
benjlevesque, line 51: can we change something to avoid needing this?

KolevDarko (approved):

packages/currency/src/erc20/networks/optimism.ts
```diff
@@ -0,0 +1,20 @@
+import { TokenMap } from './types';
+
+// List of the supported bsc network tokens
+export const supportedOptimismERC20: TokenMap = {
```
benjlevesque, line 4: do we do that for other networks? it seems unnecessary

benjlevesque (approved):
is there a doc somewhere that lists all places to update? it's still quite a lot of things to do just to add a chain; we should aim at slowly removing the need for configuration. Ideally, the currency Manager should support adding a new chain configuration without any change to the protocol's libraries

packages/currency/src/erc20/networks/optimism.ts
leoslr: We do yes

packages/currency/src/conversion-aggregators.ts
leoslr: I think we can (and should) but it needs to be thought about, it's not an easy change iirc

packages/currency/src/erc20/networks/optimism.ts
leoslr: It's used for the currency default list. So not mandatory, but I think it's good to have some currencies for each network listed there

PR #989, opened 2022-11-11
feat: add `addStakeholders`

### LINKED ISSUE #988: Add `addStakeholder`

MantisClone:
Closes #988 

# Description

This PR adds the ability to add stakeholders to an existing transaction channel. In other words, this allows invoices to be shared after they are created.

# Changes

- Transaction Manager
  - Update `ChannelParser` to search for channel key before parsing transactions.
  - Update `TransactionParser` to allow `encryptionMethod` and `keys` properties, even in transactions after the 1st one.
  - Add optional `encryptionParams` argument to `TransactionFactory.createEncryptedTransaction()`. When `encryptionParams` are given, the returned transaction object includes `keys` and `encryptionMethod`. The logic is copied from `TransactionFactory.createEncryptedTransactionInNewChannel()`.
  - Remove encryptedParam check from `TransactionManager.persistTransaction()`
- Request Logic
  - Add new `Action`:  `addStakeholders`. This action simply pushes a new `addStakeholders` event onto the request.
- Types
  - Add `IAddStakeholdersParameters` interface
  - Add `ADD_STAKEHOLDERS` to action enum.
- Request Client
  - Add new public function `addStakeholders` with required `encryptionParams` argument.

# Mental Model Changes

When a stakeholder decrypts a transaction channel, it is no longer guaranteed that their encrypted channel key will be available in the `create` transaction. If their encrypted channel key cannot be found in the `create` transaction, the stakeholder must iterate through the transactions until it finds an `addStakeholders` transaction that contains their encrypted channel key.

# Considerations

- Considered adding `addStakeholders` as an *extension* in Advanced Logic but chose to implement as an *action* in Request Logic because adding a stakeholder seemed like it warranted it's own unique action type.


vrolland:
Nice start!
Imho, the changes on `request-client.js` and `request-logic` packages look good.
Regarding the `transaction-manager`, we need deeper changes. Right now, the transaction-manager won't be able to get `keys` from a transaction that is not the first of its channel. We need to revamp `channel-parser.ts`and `transactions-parser.ts`to achieve this.
We also need to be careful about possible DoS attack if we allow to add key at any transaction in a channel.

To prepare the discussion we will have on Monday, I've started to implement (in a dirty way) these changes. I push this work on the branch [vrolland-addstakeholder](https://github.com/RequestNetwork/requestNetwork/tree/vrolland-addStakeholder). I insist it is not working, dirty, with a poor english... but it might be interesting for discussion.

PR #1001, opened 2022-11-30
feat: limited batch approval

leoslr:


benjlevesque (approved):
it would be best to have parameters pass in an object, less confusing (and less breaking!)

leoslr:
@benjlevesque I agree. The whole package needs to be refactored toward this

olivier7delf (approved):

PR #1002, opened 2022-12-01
feat: moonbeam

leoslr:
## Description of the changes

Deployment on moonbeam.

Unrelated: Deployment of the batch contract on optimism - was missing.

### DIFF
--- a/packages/currency/src/conversion-aggregators.ts
+++ b/packages/currency/src/conversion-aggregators.ts
@@ -49,6 +49,7 @@ const noConversionNetworks: AggregatorsMap = {
   avalanche: {},
   bsc: {},
   optimism: {},
+  moonbeam: {},
 };
 
 /**
--- a/packages/currency/src/native.ts
+++ b/packages/currency/src/native.ts
@@ -105,6 +105,12 @@ export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { na
       name: 'Optimism Ether',
       network: 'optimism',
     },
+    {
+      symbol: 'GLMR',
+      decimals: 18,
+      name: 'Glimmer',
+      network: 'moonbeam',
+    },
   ],
   [RequestLogicTypes.CURRENCY.BTC]: [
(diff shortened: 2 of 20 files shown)


packages/smart-contracts/scripts-create2/utils.ts
```diff
@@ -10,7 +10,6 @@ export const create2ContractDeploymentList = [
   'EthereumProxy',
   'EthereumFeeProxy',
   'ERC20FeeProxy',
-  'Erc20ConversionProxy',
```
leoslr, line 13: This contract is deployed through the old script.

KolevDarko (approved):

packages/currency/src/conversion-aggregators.ts
```diff
@@ -49,6 +49,7 @@ const noConversionNetworks: AggregatorsMap = {
   avalanche: {},
   bsc: {},
   optimism: {},
+  moonbeam: {},
```
leoslr, line 52: I made a proposal in https://github.com/RequestNetwork/requestNetwork/pull/1003 to remove this

packages/currency/src/native.ts
```diff
@@ -105,6 +105,12 @@ export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { na
       name: 'Optimism Ether',
       network: 'optimism',
     },
+    {
```
benjlevesque, line 108: this makes me wonder why we have native tokens in currency-api...

benjlevesque (approved):

packages/currency/src/native.ts
benjlevesque: oh you're right! no declarative don't make sense in here. But we should remove all detected then, it's duplicated

leoslr: The detected aren't registered in currency-api

benjlevesque: OK thanks

PR #1003, opened 2022-12-01
chore: refacto supported conversion

leoslr:
## Description of the changes

This add a utility function in the `ContractArtifact` class to determine if a contract is deployed on a specific network.

It allows to remove the `supportedConversionNetworks` variable from the advanced-logic, which reduces the number of file to update when deploying our contracts to new chains.

### DIFF
--- a/packages/advanced-logic/package.json
+++ b/packages/advanced-logic/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@requestnetwork/currency": "0.9.0",
+    "@requestnetwork/smart-contracts": "0.29.0",
     "@requestnetwork/types": "0.36.0",
     "@requestnetwork/utils": "0.36.0",
     "@types/node": "14.14.16",
--- a/packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
@@ -1,10 +1,7 @@
-import {
-  conversionSupportedNetworks,
-  ICurrencyManager,
-  UnsupportedCurrencyError,
-} from '@requestnetwork/currency';
+import { ICurrencyManager, UnsupportedCurrencyError } from '@requestnetwork/currency';
 import { ExtensionTypes, RequestLogicTypes } from '@requestnetwork/types';
 import Erc20FeeProxyPaymentNetwork from './erc20/fee-proxy-contract';
+import { erc20ConversionProxy } from '@requestnetwork/smart-contracts';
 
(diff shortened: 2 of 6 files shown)


packages/advanced-logic/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@requestnetwork/currency": "0.9.0",
+    "@requestnetwork/smart-contracts": "0.29.0",
```
benjlevesque, line 43: my opinion is that we should not add the smart-contracts dependecy to core package like advanced-logic, because of its size.

The bundle size is already a huge issue today, let's not make it worse. 

I'm sure there's a way to revert the dependency and inject the chains information inside the advanced-logic :)

packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
```diff
@@ -135,7 +132,7 @@ export default class AnyToErc20ProxyPaymentNetwork extends Erc20FeeProxyPaymentN
       return;
     }
 
-    if (!conversionSupportedNetworks.includes(network)) {
+    if (!erc20ConversionProxy.isDeployedOnNetwork(network)) {
```
benjlevesque, line 135: just to challenge: do we absolutely need this check? Or, do we need it to be done that "low" in the protocol? maybe the client could handle it

benjlevesque (changes requested):

packages/advanced-logic/src/extensions/payment-network/any-to-erc20-proxy.ts
leoslr: I know that in the RF point of view, we could get rid of it since we make similar checks (several times, I think) before creating the request. (When we check for an existing conversion path, for example).

On the protocol level, I'm not sure this is checked elsewhere, and imo it wouldn't be a good thing for someone using the protocol as is, without performing checks on his side, to be able to create a request that cannot be paid...

benjlevesque: indeed! but the request-client could cover that. 

Other solution: do it in the `payment-detection` package. There's a long overdue refactoring to remove payment networks from advanced logic anyway...

benjlevesque: (the Factory is probably a good place)

leoslr: I'll look into it, I was not that happy to add a dependency either

PR #1004, opened 2022-12-01
refactor(payment-detection): use default provider from utils

alexandre-abrioux:
## Description of the changes

The `provider.ts` files from the `payment-detection` package and the `utils` package are identical. Let's only keep one of them!


packages/payment-detection/src/erc20/proxy-info-retriever.ts
```diff
@@ -52,9 +52,9 @@ export default class ProxyERC20InfoRetriever
     private network: string,
   ) {
     // Creates a local or default provider
-    this.provider = getDefaultProvider(this.network);
+    this.provider = Utils.getDefaultProvider(this.network);
 
-    // Setup the ERC20 proxy contract interface
+    // Set up the ERC20 proxy contract interface
```
benjlevesque, line 57: TIL https://grammarist.com/spelling/set-up-vs-setup/

packages/payment-detection/src/provider.ts
```diff
@@ -1,132 +0,0 @@
...
- * Define default URLs for networks supported by Request payment detection but not by ethers' Infura Provider
- */
-const networkRpcs: Record<string, string> = {
-  private: providers.JsonRpcProvider.defaultUrl(),
-  matic: 'https://polygon-rpc.com/',
-  fantom: 'https://rpc.ftm.tools',
-  fuse: 'https://rpc.fuse.io',
-  bsctest: 'https://data-seed-prebsc-1-s1.binance.org:8545',
-  bsc: 'https://bsc-dataseed1.binance.org/',
-  xdai: 'https://rpc.gnosischain.com/',
```
benjlevesque, line 41: the xdai URL isn't the same fyi

benjlevesque (approved):
wow, I had no clue we had both. 

this reminds me of the issue with `celo`
@leoslr, do you remember why we don't handle the `celo` case inside `getDefaultProvider`?

leoslr:
@benjlevesque I remember that we did not want to handle it within the `getDefaultProvider` because it was used by other parts of our stack and we did not want to disrupt that at the time. 
But IMO it's ok to make the change now, especially since this PR goal is specific to this matter.

kevindavee (approved):

PR #1006, opened 2022-12-05
fix: toolbox cli

leoslr:
## Description of the changes

`parseSync` raised an error preventing to use the cli.

### DIFF
--- a/packages/toolbox/src/cli.ts
+++ b/packages/toolbox/src/cli.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import * as yargs from 'yargs';
 
-yargs
+void yargs
   .env('RN')
   .scriptName('request-toolbox')
   .commandDir('./commands', {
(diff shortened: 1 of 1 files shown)


KolevDarko (approved):

alexandre-abrioux (approved):

alexandre-abrioux (changes requested):
`yargs` was updated recently, you need to update your `node_modules`, it should work as is

benjlevesque (commented):
I just tried, it's all working without this change

leoslr:
From the latest doc version:

.parseAsync([args], [context], [parseCallback])
Identical to .parse() except always returns a promise for a parsed argv object, regardless of whether an async builder, handler, or middleware is used.

.parseSync([args], [context], [parseCallback])
Identical to .parse() except an exception is thrown if an asynchronous builder, handler, or middleware is used.

Several command have async builder (e.g. addAggregators). And when running the command I had the following error:
`YError: .parseSync() must not be used with asynchronous builders, handlers, or middleware`

benjlevesque (approved):

PR #1023, opened 2022-12-22
chore(utils): replace default export with named exports

### LINKED ISSUE #1015: Utils package tree shaking
## Issue

The `utils` package makes use of a default export, meaning that user code applications have to import the whole package even when using a small portion of it:
https://github.com/RequestNetwork/requestNetwork/blob/8ff76906005ed42e7624e73d7e2b579731508a0d/packages/utils/src/index.ts#L16-L28

## Ideas for improvement

- Replace the default export with named exports. This will allow for better tree shaking, lowering the build size of user code applications.
- Enforce the [no-default-export](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-default-export.md) ESLint rule.

roadrunner21:
…tree shaking (fixes #1015)

This change replaces the default export in the utils package with named exports, which allows for better tree shaking and smaller build sizes in user code applications. It also enforces the no-default-export ESLint rule to ensure consistent code style


packages/utils/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "import/no-default-export": "error"
+  }
+}
```
alexandre-abrioux, line 5: You can apply this rule in the `.eslintrc` file at the root of the repository, this way, we'll avoid making the same mistake in other packages. If there are many conflicts to fix in order to apply this rule globally, we could do it in a subsequent PR, that's OK.

alexandre-abrioux (approved):
Thank you for this 🙂

benjlevesque (commented):
this LGTM, but I only see updates on the `utils` package. There are many usage of that package in the monorepo, they need to be updated too.

packages/utils/.eslintrc
MantisClone: I created an issue to enforce `no-default-export` for all Request Network packages. https://github.com/RequestNetwork/requestNetwork/issues/1034

MantisClone:
@roadrunner21 

I believe the `payment-detection` tests are failing due to a circular dependency issue. Are you able to fix this?

I'm surprised that the `smart-contracts` tests are failing but I think that #1038 will fix it.

MantisClone (changes requested):
@roadrunner21 some of your commits are not signed. 
image

Here's how I recommend you fix it:

1. [Configure git to use your SSH signing key](https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key)
    ```
    $ git config --global gpg.format ssh
    $ git config --global user.signingkey ~/.ssh/id_ed25519.pub
    ```
2. [Add your SSH signing key to Github](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account#adding-a-new-ssh-key-to-your-account)
    * You can reuse your SSH “Authentication” Key as a “Signing” Key but it must be added separately.
    * Example: image
4. [Configure git to sign all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
    ```
    git config --global commit.gpgsign true
    ```
5. [Fix un-signed commits using rebase](https://superuser.com/a/1123928)
    ```
    git rebase --exec 'git commit --amend --no-edit -n -S' -i master
    ```

MantisClone (changes requested):
Hello @roadrunner21 :wave:  Thank you for your latest changes!

### Overview
I discussed these changes with @alexandre-abrioux and @benjlevesque. We concluded that many of the functions in `utils` need to be renamed to be more self-descriptive due to this refactor. 

### Details
In the past, module names like `identity` or `amount` helped clarify what a function was doing. For example: `identity.areEqual()` checks if two identities are equal and `amount.reduce()` reduces an amount.

After refactoring, these functions can be called without their module name `areEqual()` and `reduce()`. This is good because the caller need only import the functions that are used and not the entire module. But it makes the code harder to read without jumping to function definitions.

Thus, we'd like to see the following change:

* Please rename ambiguous functions. Examples:
  * `areEqual()` from the `identity` module --> `areEqualIdentities()`
  * `reduce()` from the `amount` module --> `reduceAmount()`
  * `min()` from the `bignumber` model --> `minBigNumber()`
  * `recover()` from the `signature` module --> `recoverSigner()`
  
* But please leave sufficiently descriptive functions unchanged. Examples:
  * `normalizeIdentityValue()` from the `identity` module
  * `encrypt()` from the `encryption` module
  * `retry()` from the `retry` module
  * all the functions in the `provider` module (some don't include the word "provider", but I think they're all self-descriptive enough)
  
If you have any questions, please don't hesitate to ask!

packages/utils/src/crypto/ec-utils.ts
```diff
@@ -4,7 +4,7 @@ import { ethers } from 'ethers';
 /**
  * Function to manage Elliptic-curve cryptography
  */
-export default {
+export const EcUtils = {
```
MantisClone, line 7: Can you please remove the named const here and export these functions individually like in the other utils modules? And then, can you please rename `recover` to `recoverSigner()`? The other functions don't need to be renamed.

packages/utils/test/crypto/crypto-wrapper.test.ts
```diff
@@ -1,11 +1,10 @@
-import CryptoWrapper from '../../src/crypto/crypto-wrapper';
-import utils from '../../src/utils';
+import { CryptoWrapper, unique } from '../../src';
 
 const anyData = 'this is any data!';
 const arbitraryKey = '12345678901234567890123456789012';
 
 /* eslint-disable @typescript-eslint/no-unused-expressions */
-describe('Utils.cryptoWrapper', () => {
+describe('Utils/cryptoWrapper', () => {
```
MantisClone, line 7: I appreciate the attention to detail here.

MantisClone (changes requested):
Hello, @roadrunner21 Sorry for the delay. I read through every line and left a few more change requests. 

I'm excited. Your changes will cut down the number of unnecessary imports across the code base!

packages/utils/src/crypto.ts
```diff
@@ -92,6 +114,6 @@ async function generate32BufferKey(): Promise<string> {
  * @returns a string of 8 random bytes
  */
 async function generate8randomBytes(): Promise<string> {
-  const random32Bytes = await CryptoWrapper.random32Bytes();
-  return random32Bytes.slice(0, 8).toString('hex');
+  const random32BytesHex = await random32Bytes();
```
MantisClone, line 117: Please revert the name of this variable back to `random32Bytes`

MantisClone (changes requested):

packages/utils/src/crypto.ts
roadrunner21: The variable random32Bytes caused a naming collision because random32Bytes() is now being called directly. I changed it to random32BytesHex to indicate that it holds the hexadecimal value of the random bytes returned by the function, this is to prevent naming collision.

Other variable names that we could use would be randomBytes, randomHex, randomHexBytes.

Let me know if you have any specific suggestions for the variable name.

MantisClone (approved):
Thanks for bearing with me @roadrunner21 this is great work.

PR #1024, opened 2022-12-22
feat: tombchain

leoslr:
## Description of the changes

Tombchain is a permissioned L2 evm.
Deployments: only `ERC20FeeProxy` and `EthFeeProxy.`

Additional fix:
When making the deployment, I was blocked as the Tombchain Node RPC did not support the eth_FeeHistory method used by the eip1559.
All evm nodes may not support it yet. It does not block the deployment anymore

### DIFF
--- a/packages/currency/src/native.ts
+++ b/packages/currency/src/native.ts
@@ -111,6 +111,12 @@ export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { na
       name: 'Glimmer',
       network: 'moonbeam',
     },
+    {
+      symbol: 'TOMB',
+      decimals: 18,
+      name: 'Tomb',
+      network: 'tombchain',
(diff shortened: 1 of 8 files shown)


benjlevesque (approved):

alexandre-abrioux (approved):
@leoslr FYI this is how it's done for the storage to detect if it can use EIP-1559:
https://github.com/RequestNetwork/requestNetwork/blob/0af020b8d2efd9c46ce8c8002d3f655bdb61b42f/packages/ethereum-storage/src/ethereum-tx-submitter.ts#L35-42

It could be extracted into an util function

leoslr:
@alexandre-abrioux I won't add it to this PR to keep what's related to contract deployment clean. I'll add myself a ticket to do it in the next weeks

PR #1025, opened 2022-12-23
feat: add `addStakeholders`

### LINKED ISSUE #988: Add `addStakeholder`

MantisClone:
Continued from #989 
Closes #988 

# Description

This PR adds the ability to add stakeholders to an existing transaction channel. In other words, this allows invoices to be shared after they are created.

# Changes

- Transaction Manager
  - Update `ChannelParser.decryptAndCleanChannel()` to search and decrypt channel key before parsing transactions.
  - Update `TransactionParser.parsePersistedTransaction()` to handle transactions that add new encrypted channel keys in subsequent transactions after creation. The `enryptionMethod` property only appears in the creation transaction.
  - Add optional `encryptionParams` argument to `TransactionFactory.createEncryptedTransaction()`. When `encryptionParams` are given, the returned transaction object includes the `keys` property. The logic is copied from `TransactionFactory.createEncryptedTransactionInNewChannel()`.
  - Remove `encryptedParam` check from `TransactionManager.persistTransaction()`
- Request Logic
  - Add new `Action`:  `addStakeholders`. This action simply pushes a new `addStakeholders` event onto the request.
- Types
  - Add `IAddStakeholdersParameters` interface
  - Add `ADD_STAKEHOLDERS` to action enum.
- Request Client
  - Add new public function `addStakeholders` with required `encryptionParams` argument.

# Mental Model Changes

When a stakeholder decrypts a transaction channel, it is no longer guaranteed that their encrypted channel key will be available in the `create` transaction. If their encrypted channel key cannot be found in the `create` transaction, the stakeholder must iterate through the transactions until it finds an `addStakeholders` transaction that contains their encrypted channel key.

# Considerations

- Considered adding `addStakeholders` as an *extension* in Advanced Logic but chose to implement as an *action* in Request Logic because adding a stakeholder seemed like it warranted it's own unique action type.


MantisClone:
I'm going to extract the `DataAccess` injection feature into a new PR so it's easier to review. Please see #1047

MantisClone:
I'm excited to share that my [`request-client-js-add-stakeholders` script](https://github.com/RequestNetwork/requestNetwork/blob/139afc0dfb24e550ca27a2319b96d92fe8eeb501/packages/usage-examples/src/request-client-js-add-stakeholders.ts) appears to be working as intended.

It creates two `RequestNetwork` instances with different `DecryptionProvider` instances and shared `MockDataAccess` and `MockStorage` instances. Using the first `RequestNetwork` instance, I create an encrypted request and add a stakeholder to it. Using the 2nd `RequestNetwork` instance, I'm able to successfully fetch and decrypt the request. :tada:

MantisClone:
The idea for future iterations:
1. Config to ignore addStakeholders actions

yomarion (commented):
(review in progress)

packages/usage-examples/package.json
```diff
@@ -24,7 +24,7 @@
     "clean": "shx rm -rf dist tsconfig.tsbuildinfo tsconfig.build.tsbuildinfo",
     "lint": "eslint . --fix",
     "lint:check": "eslint .",
-    "start": "ts-node src/request-client-js-declarative-request.ts && ts-node src/request-client-js-erc20-request.ts && ts-node src/request-logic-clear-request.ts && ts-node src/request-logic-encrypted-request.ts",
+    "start": "ts-node src/request-client-js-add-stakeholders.ts",
```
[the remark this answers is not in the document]
MantisClone: fixed

packages/transaction-manager/src/transactions-factory.ts
```diff
@@ -136,6 +139,62 @@ export default class TransactionsFactory {
 
     const encryptedDataSerialized: string = MultiFormat.serialize(encryptedData);
 
-    return { encryptedData: encryptedDataSerialized };
+    if (encryptionParams.length === 0) {
+      return { encryptedData: encryptedDataSerialized };
+    } else {
+      // TODO: Refactor, duplicate with createEncryptedTransactionInNewChannel
```
yomarion, line 145: Do you plan to fix the TODO before merging? If not use `FIXME` instead

packages/transaction-manager/src/transactions-factory.ts
```diff
@@ -136,6 +139,62 @@ export default class TransactionsFactory {
...
+      if (
+        !encryptionParams.every(
+          (encryptionParam: EncryptionTypes.IEncryptionParameters) =>
+            encryptionParam.method === EncryptionTypes.METHOD.ECIES,
+        )
+      ) {
+        throw new Error(`encryptionParams method must be all: ${EncryptionTypes.METHOD.ECIES}`);
+      }
+
+      // TODO: Check if channel already includes new encryptionParams
```
yomarion, line 156: TODO here

yomarion (approved):
All good except the remaining TODOs

packages/transaction-manager/src/transactions-factory.ts
MantisClone: Changed to FIXME

packages/transaction-manager/src/transactions-factory.ts
MantisClone: Removing this TODO.  We don't actually want to check if the channel already includes the new encryptionParams.  This would be an O(n) operation. This is not worth doing when adding new encrypted taransactions.

MantisClone: Removed.

KolevDarko (approved):

PR #1041, opened 2023-01-03
chore: add utility function to check eip1559 compliance

leoslr:
## Description of the changes

Small refacto to regroup logic related to eip1559 compliance
[Asana task](https://app.asana.com/0/1203350802794948/1203600402430065)

### DIFF
--- a/packages/utils/src/providers.ts
+++ b/packages/utils/src/providers.ts
@@ -1,3 +1,5 @@
+import { LogTypes } from '@requestnetwork/types';
+
 import { providers, constants } from 'ethers';
 
 type ProviderFactory = (network: string | undefined) => providers.Provider | string;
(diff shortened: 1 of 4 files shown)


packages/utils/src/providers.ts
```diff
@@ -147,9 +149,22 @@ const getCeloProvider = (): providers.Provider => {
...
 };
 
+const isEip1559Supported = async (provider: any, logger: LogTypes.ILogger): Promise<boolean> => {
+  try {
+    await provider.send('eth_feeHistory', [1, 'latest', []]);
+    return true;
+  } catch (e) {
+    logger.warn(
+      'This RPC provider does not support the "eth_feeHistory" method: switching to legacy gas price',
+    );
```
benjlevesque, line 159: I think this log should be removed from the utility

packages/utils/src/providers.ts
```diff
@@ -147,9 +149,22 @@ const getCeloProvider = (): providers.Provider => {
   return provider;
 };
 
+const isEip1559Supported = async (provider: any, logger: LogTypes.ILogger): Promise<boolean> => {
```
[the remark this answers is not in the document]
benjlevesque: a hack you can do is to type it as `providers.Provider`, then cast it to `JsonRpcProvider`.

Not great, but still better than `any` ;)

benjlevesque (approved):

alexandre-abrioux (approved):
nice!

leoslr:
@benjlevesque this does not work. As soon as the type of the parameter is related to the `ethers` library it raise the error, it's related to how we export functions in the `utils` package.

benjlevesque:
> @benjlevesque this does not work. As soon as the type of the parameter is related to the `ethers` library it raise the error, it's related to how we export functions in the `utils` package.

OK, so maybe https://github.com/RequestNetwork/requestNetwork/pull/1023 will solve it...

yomarion (approved):

packages/utils/src/providers.ts
leoslr: I kept it there but made it optional.

PR #1042, opened 2023-01-04
Made many suggested changed to subgraph docs

PaulieB14:


MantisClone (changes requested):
Hey @PaulieB14 I glanced at this PR. I don't see any changes to the subgraph docs. Are you sure you pushed them to Github?

MantisClone:
Github thinks this was merged because the source and target branches are identical.  But that's because the target branch was synchronized with upstream `master`, and this PR doesn't actually add any new commits.

PR #1048, opened 2023-01-16
chore(smart-contracts): publish to tenderly

alexandre-abrioux:
## Description of the changes

Adds a command to import contracts in a Tenderly account

### Related
- https://app.asana.com/0/1203350802794955/1203755698846365/f
- https://app.asana.com/0/1203350802794955/1203755698846366/f


packages/smart-contracts/scripts-create2/utils/chains.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type ChainConfig = {
+  name: string;
+  chainId: number;
+  rpcUrls: string[];
+};
+
+const cachedChainConfigs: Record<string, ChainConfig> = {};
+
+export const getChainConfig = async (chainName: string) => {
```
alexandre-abrioux, line 11: - moved this function out from the `toolbox` utils to the `smart-contracts` utils
- added a simple cache

packages/smart-contracts/scripts-create2/utils/utils.ts
```diff
@@ -1,5 +1,5 @@
 import { Contract } from 'ethers';
```
benjlevesque, line 1: rename this file `index`? or anything else than `utils` 😬

benjlevesque (commented):
nice! 

A few comments:
- what will happen on chains not supported by tenderly? 
- could we avoid relying on the (request finance) currency API, in a core library of the protocol? It seems it's used only to get the chainId, which can probably be found with a call to the RPC, that I think is already available in Hardhat configuration?

packages/smart-contracts/src/lib/ContractArtifact.ts
```diff
@@ -94,6 +94,24 @@ export class ContractArtifact<TContract extends Contract> {
     }));
   }
 
+  /**
+   * Retrieve all addresses for all versions for all networks
+   * @returns the addresses of the deployed contract and the associated network and version.
+   */
+  getAllAddressesFromAllNetworks(): { version: string; address: string; networkName: string }[] {
```
kevindavee, line 101: No enum or string union for the `networkName`?

kevindavee (approved):

packages/smart-contracts/src/lib/ContractArtifact.ts
alexandre-abrioux: not sure that we have this ☹️

leoslr (approved):

packages/smart-contracts/src/lib/ContractArtifact.ts
alexandre-abrioux: fixed in https://github.com/RequestNetwork/requestNetwork/pull/1056

benjlevesque (approved):

KolevDarko (approved):

PR #1056, opened 2023-02-03
refactor: chain configurations

alexandre-abrioux:
## Description of the changes

This PR adds chain configuration files. Those configs allow us to easily convert Request Network's internal chain names (`mainnet`, `aurora`, etc...) to standard chain IDs.


packages/currency/test/currencyManager.test.ts
```diff
@@ -592,63 +598,64 @@ describe('CurrencyManager', () => {
...
   describe('Validate currencies', () => {
-    describe('Valid cases', () => {
-      it.each([
-        {
-          currency: {
-            type: RequestLogicTypes.CURRENCY.ISO4217,
-            value: 'FIAT',
-          },
-          label: 'ISO4217 Currency',
+    const currencies: { currency: StorageCurrency; label: string }[] = [
```
alexandre-abrioux, line 601: Here I just extracted the object to type it (no change was made to the object)

packages/ethereum-storage/src/ethereum-utils.ts
```diff
@@ -1,50 +1,46 @@
...
+ */
 
 /**
- * Collection of utils functions related to Ethereum
+ * Get the name of the Ethereum network from its id
+ *
+ * @param networkId Id of the network
+ * @return name of the network
  */
-export default {
```
alexandre-abrioux, line 19: I removed the default export here (cc @MantisClone) and exported these utilities in the index file of the package. It allows other packages to filter storage chains amongst all supported EVM chains.

packages/request-node/src/thegraph-node.ts
```diff
@@ -28,7 +24,11 @@ export class TheGraphRequestNode extends RequestNodeBase {
         })
       : undefined;
 
-    const network = getNetworkFromId(config.getStorageNetworkId());
+    const network = getEthereumStorageNetworkNameFromId(config.getStorageNetworkId());
+    if (!network) {
+      throw new Error(`Storage network not supported: ${config.getStorageNetworkId()}`);
+    }
```
alexandre-abrioux, line 30: This is how the newly exported `ethereum-storage` utility becomes convenient.

packages/types/src/extensions/pn-any-to-erc20-types.ts
```diff
@@ -1,10 +1,12 @@
 import * as PnAnyToAnyConversion from './pn-any-to-any-conversion-types';
+import { EvmChainName } from '../currency-types';
 
 /** Any to ERC20 reference-based payment network extension interface */
 export type IAnyToERC20 = PnAnyToAnyConversion.IConversionReferenceBased<ICreationParameters>;
 
 /** Parameters for the creation action */
 export interface ICreationParameters extends PnAnyToAnyConversion.ICreationParameters {
+  network?: EvmChainName;
```
alexandre-abrioux, line 9: precise `network` type for `pn-any-to-erc20`

benjlevesque (commented):
great job on the chain configuration 👏 

I'm not totally convinced by the `EvmChainName` though... Imo, we could consider chains as a purely configurable: in the future, chain configuration could even include smart contract addresses, and could be injected to the client. That would make typing the chain name irrelevant, and even wrong. 

I think you cover the problem very well with the `assertChainSupported` util, and it's probably enough.

alexandre-abrioux:
_(saw on a call with @benjlevesque)_

## Use case 1

`EvmChainName` / `BtcChainName` are nice to have with our current "static" configuration as they prevent typos in chain names in:
  - the [erc20 currency definition list](https://github.com/RequestNetwork/requestNetwork/blob/bae483032d22007eadfb5ebac292034b94723bb9/packages/currency/src/erc20/chains/index.ts)
  - the [erc777 currency definition list](https://github.com/RequestNetwork/requestNetwork/blob/ed3b672e343031df64272fafbdb69619f2d8c61e/packages/currency/src/erc777/chains/index.ts#L5)
  - the [contract artifact lists](https://github.com/RequestNetwork/requestNetwork/blob/20000587318107e97742688f69ba561868e39f8f/packages/smart-contracts/src/lib/artifacts/index.ts)

This use case will be voided once we implement a proper way of dynamically configuring chains. Like @benjlevesque said:

> in the future, chain configuration could even include smart contract addresses and could be injected to the client

At this point, we should implement all safeguards at run time (when the library is first loaded, for instance), which would replace those compile-time safeguards.

## Use case 2

`EvmChainName` / `BtcChainName` can shield features that should only be used on EVM or BTC chains against using them on other chains. Example: ERC20 payment extensions should only be used on EVM chains.

This is also a nice to have, but having a union type is not future-proof (it's not compatible with a dynamic configuration). What we could do is implement proper classes to represent chains in our codebase, instead of representing them as strings. That way:
- `type EvmChainName` would become `interface EvmChain`
- `type BtcChainName` would become `interface BtcChain`

The feature-shielding could be preserved. The assertion of compatibility would be made when first creating the chain class, something like:

```ts
class EvmChain implements  {
  contructor (public name) {
    // here we would check that the chain is supported from the dynamic chain configuration
    assertChainSupported(name);
  }
}
```

## Conclusion

Overall, those types were hard to add but will be very easy to replace by `interfaces` or by simple `string` aliases once we move forward with dynamic configurations. We could merge this as a first step and iterate on it.

Please let me know if you have other concerns or suggestions!

## Misc

A suggestion was to use the newly added `satisfies` keyword instead of manually creating a union type (cf. https://github.com/RequestNetwork/requestNetwork/pull/1057).

Unfortunately, as the chain configurations are declared in `@requestnetwork/currency`, `@requestnetwork/types` [need those types too](https://github.com/RequestNetwork/requestNetwork/blob/ed3b672e343031df64272fafbdb69619f2d8c61e/packages/types/src/payment-types.ts#L38), this would create a circular dependency between `@requestnetwork/types` and `@requestnetwork/currency`.

packages/types/src/currency-types.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * List of supported EVM chains
+ */
+export type EvmChainName =
+  | 'alfajores'
+  | 'arbitrum-one'
+  | 'arbitrum-rinkeby'
+  | 'aurora'
+  | 'aurora-testnet'
```
yomarion, line 9: The name `aurora` was a mistake, and we still haven't fixed it. We don't support the actual (EVM) Aurora chain yet. But we mistakenly use `aurora` as an identifier for Near, which is not an EVM. Should you have `EvMChainName` and `NonEvms`?

packages/types/src/extensions/pn-any-to-erc20-types.ts
yomarion: I'd like to know if @leoslr thinks this is compatible with the approach of calling all fungible tokens `ERC20`, even on non-EVMs.

leoslr: It won't be compatible no. 
Currently the tests are pasing as `aurora` and `aurora-testnet` are declared as `EvmChainName`.

I think we should keep the `ChainName` from the underlying interface. Calling all fungible token ERC20 is a langage abuse, but in the context of creating requests and detecting the payment, the logic remains the same.

packages/currency/src/types.ts
```diff
@@ -1,13 +1,33 @@
-import { RequestLogicTypes } from '@requestnetwork/types';
+import { CurrencyTypes, RequestLogicTypes } from '@requestnetwork/types';
+
+/**
+ * Common types used in token configuration files
+ */
+type TokenAddress = string;
+type TokenDefinition = { name: string; symbol: string; decimals: number };
+export type TokenMap = Record<TokenAddress, TokenDefinition>;
+export type ChainDefinition = {
```
yomarion, line 9: Can you come up with another name for this type? It's not natural to "define" a chain by its currency map. Maybe `ChainDetails` or `ChainTokenMap`? Also, this type should be documented in the Currency README, it becomes key.

packages/currency/src/native.ts
```diff
@@ -1,7 +1,11 @@
-import { RequestLogicTypes } from '@requestnetwork/types';
-import { NativeCurrency, NativeCurrencyType } from './types';
+import { CurrencyTypes, RequestLogicTypes } from '@requestnetwork/types';
+import { NamedNativeCurrency } from './types';
 
-export const nativeCurrencies: Record<NativeCurrencyType, (NativeCurrency & { name: string })[]> = {
+type NativeEthCurrency = NamedNativeCurrency & { network: CurrencyTypes.EvmChainName };
```
yomarion, line 4: 1/ Generic question: why having a different type per blockchain type?

2/
Watch out: 
```
    {
      symbol: 'NEAR',
      decimals: 24,
      name: 'Near',
      network: 'aurora',
    },
```
Was defined within `RequestLogicTypes.CURRENCY.ETH`, as it seemed complex to define one Currency type per chain. These new types can enforce the confusion, maybe it's easier to make the difference at this stage with `NativeNearCurrency` (but cf. 1/).

Option 1/n: `NativeAMCurrency`, AM standing for `AccountModel`, opposed to UTXO (https://www.horizen.io/blockchain-academy/technology/expert/utxo-vs-account-model/#:~:text=The%20UTXO%20model%20is%20a,constructions%2C%20as%20well%20a%20sharding.)
Option 2/n: `NativeVMCurrency`: Common type for VM chains native tokens
...

3/ Consider usign `Evm` instead of `Eth`

packages/currency/src/erc777/chains/index.ts
```diff
@@ -0,0 +1,9 @@
+import { supportedRinkebyERC777 } from './rinkeby';
+import { TokenMap } from '../../types';
+import { CurrencyTypes } from '@requestnetwork/types';
+
+export const supportedNetworks: Partial<Record<CurrencyTypes.EvmChainName, TokenMap>> = {
+  rinkeby: supportedRinkebyERC777,
+};
+
+export type { TokenMap };
```
yomarion, line 9: I think this export is now useless.

packages/currency/src/chains/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function genericAssertChainSupported<T extends string>(
+  chainKey: string,
+  supportedChains: T[],
+): asserts chainKey is T {
+  if (!(supportedChains as string[]).includes(chainKey))
+    throw new Error(`Unsupported chain ${chainKey}`);
+}
+
+export const addNativeCurrenciesToChains = (
```
yomarion, line 13: TODO: quick doc

packages/currency/src/chains/evm/index.ts
```diff
@@ -0,0 +1,76 @@
...
+export const chainNames = Object.keys(chains) as CurrencyTypes.EvmChainName[];
+
+// add native currencies
+addNativeCurrenciesToChains(chains, RequestLogicTypes.CURRENCY.ETH);
+
+/**
+ * Asserts if a specific chain is supported across EVM-type supported chains
+ * @param chainName
+ */
+export function assertChainSupported(
```
yomarion, line 66: Consider renaming `assertEVMChainSupported` to avoid mistakes with the global one? (Same remark for BTC)

packages/currency/src/chains/evm/mainnet.ts
```diff
@@ -0,0 +1,7 @@
+import { TokenMap } from '../../types';
```
yomarion, line 1: This file structure should be referenced in the `README` as it looks key.

packages/currency/src/chains/evm/index.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as SokolDefinition from './sokol';
+import * as TombchainDefinition from './tombchain';
+import * as XDaiDefinition from './xdai';
+
+export const chains: Record<CurrencyTypes.EvmChainName, EvmChainDefinition> = {
+  alfajores: AlfajoresDefinition,
+  'arbitrum-one': ArbitrumOneDefinition,
+  'arbitrum-rinkeby': ArbitrumRinkebyDefinition,
+  aurora: AuroraDefinition,
+  'aurora-testnet': AuroraTestnetDefinition,
```
yomarion, line 34: (Repeated comment)
`aurora` and `aurora-testnet` are actually not supported and have been used as IDs for Near chains.

packages/currency/src/erc20/chains/fantom.ts
```diff
@@ -1,4 +1,4 @@
-import { TokenMap } from './types';
+import { TokenMap } from '../../types';
 
 export const supportedFantomTokens: TokenMap = {
```
yomarion, line 3: Consider renaming `supportedFantomERC20` for consistency

packages/currency/src/chains/evm/aurora-testnet.ts
```diff
@@ -0,0 +1 @@
+export const chainId = 1313161555;
```
yomarion, line 1: (Repeated comment)
Mind the file location, should ideally be moved out of `evm` until we fix the Near identification.

packages/currency/src/chains/evm/aurora.ts
```diff
@@ -0,0 +1 @@
+export const chainId = 1313161554;
```
yomarion, line 1: (Repeated comment)
Mind the file location, should ideally be moved out of `evm` until we fix the Near identification.

yomarion (commented):
(reviewed `advanced-logic`, `currency`, `types`, `toolbox` and `smart-contracts` for now)

packages/currency/src/native.ts
alexandre-abrioux: I fixed it in https://github.com/RequestNetwork/requestNetwork/pull/1056/commits/d6ae451ecd548c5f27b6686fc721891e7eb9d2cc with the following bit of code:
```ts
type NativeEthCurrency = NamedNativeCurrency & {
  network: CurrencyTypes.EvmChainName | CurrencyTypes.NearChainName;
};
```

> why having a different type per blockchain type

The goal was to have a different type per currency type.

IMO, if we can specify the type, let's do it; it brings clarity to the model. The change already revealed small flaws in the current design (which are OK flaws):
- **flaw 1**: because chain names are not unique across different chain types, a collision is possible for currency types that span across multiple chain types, like `RequestLogicTypes.CURRENCY.ETH` that spans across EVM-type and NEAR-type chains
- **flaw 2**: the currency type `RequestLogicTypes.CURRENCY.ETH`'s name is a bit confusing

I think that's great: it made it more obvious. I didn't want to change any logic in this PR, so I haven't refactored anything related to currencies. Let's keep it in mind tho! I didn't know about "UTXO vs. Account Model", thanks!

packages/currency/src/erc777/chains/index.ts
alexandre-abrioux: fixed in c7b3112a

packages/currency/src/erc20/chains/fantom.ts
alexandre-abrioux: fixed in e52eb104

packages/currency/src/types.ts
alexandre-abrioux: fixed in a446daf0

alexandre-abrioux: fixed in a446daf0

packages/currency/src/chains/evm/index.ts
alexandre-abrioux: fixed in 05ab7d5e

packages/currency/src/chains/utils.ts
alexandre-abrioux: fixed in 05ab7d5e

packages/currency/src/chains/evm/mainnet.ts
alexandre-abrioux: What do you recommend we describe? The file structure is [enforced by TS](https://github.com/RequestNetwork/requestNetwork/blob/05ab7d5ef587de53bd2d8c3e3fb8181586535ef6/packages/currency/src/chains/evm/index.ts#L26).

benjlevesque (approved):

packages/payment-detection/src/any/any-to-erc20-proxy.ts
```diff
@@ -78,7 +84,7 @@ export class AnyToERC20PaymentDetector extends ERC20FeeProxyPaymentDetectorBase<
     toAddress: string | undefined,
     paymentReference: string,
     requestCurrency: RequestLogicTypes.ICurrency,
-    paymentChain: string,
+    paymentChain: CurrencyTypes.EvmChainName,
```
leoslr, line 87: Same in this file. The detection should be agnostic on whether it's from an evm chain or not as long as `ERC20` is supported

packages/currency/src/types.ts
```diff
@@ -26,7 +43,7 @@ export type ISO4217Currency = {
 export type ERC20Currency = {
   symbol: string;
   decimals: number;
-  network: string;
+  network: CurrencyTypes.EvmChainName;
```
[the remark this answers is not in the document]
alexandre-abrioux: fixed in 97b3438c

packages/currency/src/chains/evm/mainnet.ts
alexandre-abrioux: I actually added some info in [3452fd1](https://github.com/RequestNetwork/requestNetwork/pull/1056/commits/3452fd1a3f06b413a44d9406b60b271dacd2f206) because the `testnet: true` property that is introduced in https://github.com/RequestNetwork/requestNetwork/pull/1048 is not obvious and not a required property. Let me know what you think.

leoslr (approved):

KolevDarko (approved):

PR #1077, opened 2023-02-27
fix(request-node): Add exponential backoff in fetchAndRetry

### LINKED ISSUE #1076: Fix: 429 Too Many Requests in create.request, IPFS Timeout in request-node
# Problem

* Request Node polls for request confirmation, 1 RPC per second for 30 seconds.
* Under heavy load, RPC Node Provider starts returning HTTP 429 Too Many Requests.
* As a result:
  * the create.request app throws HTTP 429 Too Many Requests
  * the request node throws IPFS Timeout errors

# Solution

Exponential backoff with no max timeout

mliu:
Closes #1076 
Closes [dw-99](https://app.dework.xyz/request-network/development-56861?taskId=e0b018b6-f29a-4e7e-a87e-030c790e792a)

This PR introduces an `exponentialBackoff` optional flag to `fetchAndRetry` calls that increases the time between retried http requests at an exponential rate. These rates are further configurable by setting parameters in `retry.ts` but defaults to an exponential growth rate of 2 and a max cap of 30 seconds (i.e. 2, 4, 8, 16, 30). Existing logic like base delay between retries and maximum retries allowed are still preserved.


MantisClone:
I should have mentioned this: DeWork recognizes branches like `dw-99/fix-the-thing` but ignores branches like `dw-99-fix-the-thing`  I'm not sure why.  

I don't want you to change the branch name because that would auto-close the PR.  Instead, could you please click the "Submit Work" button on the DeWork task and post the PR link there?

MantisClone (changes requested):
The request-node supports command-line arguments.  Can you please ensure that passing them as command-line args works and add descriptions to the README?

mliu:
> The request-node supports command-line arguments. Can you please ensure that passing them as command-line args works and add descriptions to the README?

Discussed offline, the command-line args don't apply here since this is request client executed code. This implementation is already following precedence of previous config values.

MantisClone:
@mliu I implemented the exponential backoff configuration parameters but wasn't able to push them to your fork so I pushed them to an upstream branch instead: https://github.com/RequestNetwork/requestNetwork/compare/master...dw-99-exponential-backoffs-in-retry

I think this feature may warrant a new test or two in https://github.com/RequestNetwork/requestNetwork/blob/dw-99-exponential-backoffs-in-retry/packages/utils/test/retry.test.ts

mliu:
Thanks! Added tests

MantisClone (approved):

packages/utils/test/retry.test.ts
```diff
@@ -122,4 +122,48 @@ describe('Retry', () => {
...
+
+    retry(throwFn, {
+      exponentialBackoffDelay: 1000,
+      maxExponentialBackoffDelay: 7000,
+    })();
+
+    // Should call immediately
+    expect(throwFn).toHaveBeenCalledTimes(1);
+
+    // Exponential backoff should only call a second time after 2000ms
```
MantisClone, line 142: This makes me realize that if you want to have back off like: 1 second, 2 seconds, 4 seconds, 8 seconds, then you need to set the `exponentialBackoffDelay` to 500 ms.

PR #1086, opened 2023-03-19
feat: ERC20FeeProxy on Near testnet

yomarion:


packages/transaction-manager/README.md
```diff
@@ -17,7 +17,7 @@ npm install @requestnetwork/transaction-manager
 In order to decrypt encrypted transactions, you must provide a decryption provider.
 You can see the specification of decryption provider [here](./specs/decryption-provider.md).
 
-An example of implementation of a decryption provider is available in the package: [epk-decryption](../../epk-decryption)
+An example of implementation of a decryption provider is available in the package: [epk-decryption](../epk-decryption/)
```
yomarion, line 20: Minor unrelated fix

packages/types/src/extension-types.ts
```diff
@@ -23,13 +23,13 @@ export {
 };
 
 /** Extension interface is extended by the extensions implementation */
-export interface IExtension<T = any> {
+export interface IExtension<TCreationParameters = any> {
```
yomarion, line 26: Minor unrelated clarification

yarn.lock
yomarion, line 1: TODO this yarn lock update is probably irrelevant or wrong

leoslr (approved):
I could not find any blocker.
Just a general, question, a saw that most of the test were using `aurora` / `aurora-testnet.` Are we sure they would pass as well using `near` / `near-testnet` ? 
Or that is planned for later ?

yomarion:
> I could not find any blocker. Just a general, question, a saw that most of the test were using `aurora` / `aurora-testnet.` Are we sure they would pass as well using `near` / `near-testnet` ? Or that is planned for later ?

Yes, we plan to make it a different change. With the recent strong typing of `NearChains`, it should be quite easy, we can use type the actual `NearChains` and `NearChainsAlias` differently, and little by little reduce the support surface for `NearChainsAlias`.

Thanks for your review @leoslr

packages/smart-contracts/scripts-create2/tenderly.ts
```diff
@@ -55,20 +55,26 @@ export const tenderlyImportAll = async (hre: HardhatRuntimeEnvironmentExtended):
...
-        const contractId = getTenderlyContractId(contract);
-        contracts[contractId] = {
-          name: capitalizeFirstLetter(artifactName.replace(/Artifact/i, '')),
-          ...contract,
-        };
-        versions[version] ??= new Set();
-        versions[version].add(contractId);
-        (EvmChains.isTestnet(networkName) ? testnetContracts : mainnetContracts).add(contractId);
+        try {
+          EvmChains.assertChainSupported(networkName);
```
alexandre-abrioux, line 59: Does it work if you `catch ... continue` just for this line instead of ignoring errors for the whole block? I'm afraid we might ignore unrelated errors that might be hard to debug

yarn.lock
alexandre-abrioux, line 1: Was this intended? I didn't see changes to dependencies

alexandre-abrioux (approved):
Awesome changes, I like the new chain helpers! 🙂
And thank you for the shoutout 😇

PR #1094, opened 2023-04-04
feat: batch create encrypted requests

leoslr:
#### Context

Creating multiple encrypted requests in a row can result in performance issues depending on your `SignatureProvider`.

####  Update

This PR introduces several changes to mitigate this issue:
1. The `SignatureProvider` exposes a new method: `batchSign` (optional).
2. The `RequestNetwork` client now exposes the `_batchCreateEncryptedRequests` method. To be able to use this method, the given `SignatureProvider` must implement the batchSign method.

The `SignatureProvider` will be called only once to sign the data.

#### Limitations

When calling this method, if `disableEvents` or `skipRefresh` are `false`, the `SignatureProvider` will be called multiple times during the refresh process.

### DIFF
--- a/packages/request-logic/src/types.ts
+++ b/packages/request-logic/src/types.ts
@@ -1,0 +1,13 @@
+import { IdentityTypes, RequestLogicTypes } from 'types/dist';
+
+export interface ICreateCreationActionRequestIdAndTopicsParameters {
+  requestParameters: RequestLogicTypes.ICreateParameters;
+  signerIdentity: IdentityTypes.IIdentity;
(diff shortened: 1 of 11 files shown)


packages/request-logic/src/types.ts
```diff
@@ -0,0 +1,13 @@
+import { IdentityTypes, RequestLogicTypes } from 'types/dist';
```
alexandre-abrioux, line 1: wrong import

alexandre-abrioux (approved):
Cool stuff 😎

PR #1096, opened 2023-04-07
feat: processor for ERC20FeeProxy on Near

yomarion:
## Description of the changes

feat: `payment-processor` method for paying an ERC20FeeProxy request on Near, by sending fungible tokens to a proxy contract.

chore: deployment address for mainnet.

### DIFF
--- a/packages/payment-processor/src/payment/near-fungible.ts
+++ b/packages/payment-processor/src/payment/near-fungible.ts
@@ -16,7 +16,7 @@ import { NearChains } from '@requestnetwork/currency';
  * Processes the transaction to pay a request in fungible token on NEAR with fee (Erc20FeeProxy).
  * @param request the request to pay
  */
-export async function payFungibleNearRequest(
+export async function payNearFungibleRequest(
   request: ClientTypes.IRequestData,
   walletConnection: WalletConnection,
   amount?: BigNumberish,
--- a/packages/payment-processor/src/payment/near-fungible.ts
+++ b/packages/payment-processor/src/payment/near-fungible.ts
@@ -1,0 +1,64 @@
+import { BigNumberish } from 'ethers';
+import { WalletConnection } from 'near-api-js';
+
+import { ClientTypes, ExtensionTypes } from '@requestnetwork/types';
+
+import {
+  getRequestPaymentValues,
+  validateRequest,
+  getAmountToPay,
+  getPaymentExtensionVersion,
+} from './utils';
+import {
+  INearTransactionCallback,
+  isReceiverReady,
+  processNearFungiblePayment,
+} from './utils-near';
+import { NearChains } from '@requestnetwork/currency';
+
+/**
+ * Processes the transaction to pay a request in fungible token on NEAR with fee (Erc20FeeProxy).
+ * @param request the request to pay
+ */
+export async function payFungibleNearRequest(
+  request: ClientTypes.IRequestData,
+  walletConnection: WalletConnection,
+  amount?: BigNumberish,
+  callback?: INearTransactionCallback,
+): Promise<void> {
+  validateRequest(request, ExtensionTypes.PAYMENT_NETWORK_ID.ERC20_FEE_PROXY_CONTRACT);
+
+  const { paymentReference, paymentAddress, feeAddress, feeAmount, network } =
+    getRequestPaymentValues(request);
+
+  if (!paymentReference) {
+    throw new Error('Cannot pay without a paymentReference');
+  }
+
+  if (!network || !NearChains.isChainSupported(network)) {
+    throw new Error('Should be a Near network');
+  }
+  NearChains.assertChainSupported(network);
+
+  const amountToPay = getAmountToPay(request, amount).toString();
+  const version = getPaymentExtensionVersion(request);
+
+  if (!(await isReceiverReady(walletConnection, request.currencyInfo.value, paymentAddress))) {
+    throw new Error(
+      `The paymentAddress is not registered for the token ${request.currencyInfo.value}`,
+    );
+  }
+
+  return processNearFungiblePayment(
+    walletConnection,
+    network,
+    amountToPay,
+    paymentAddress,
+    paymentReference,
+    request.currencyInfo.value,
+    feeAddress || '0x',
+    feeAmount || 0,
+    version,
+    callback,
+  );
+}
--- a/packages/payment-processor/src/payment/near-fungible.ts
+++ b/packages/payment-processor/src/payment/near-fungible.ts
@@ -1,14 +1,10 @@
 import { BigNumberish } from 'ethers';
 import { WalletConnection } from 'near-api-js';
 
+import { erc20FeeProxyArtifact } from '@requestnetwork/smart-contracts';
 import { ClientTypes, ExtensionTypes } from '@requestnetwork/types';
 
-import {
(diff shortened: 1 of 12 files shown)


packages/payment-processor/src/payment/swap-any-to-erc20.ts
```diff
@@ -77,8 +77,8 @@ export function encodeSwapToPayAnyToErc20Request(
   signerOrProvider: providers.Provider | Signer = getProvider(),
   options: IRequestPaymentOptions,
 ): string {
-  const conversionSettings = options?.conversion;
-  const swapSettings = options?.swap;
+  const conversionSettings = options.conversion;
```
yomarion, line 80: Not related, these arguments are not optional anymore.

packages/payment-processor/src/payment/utils.ts
```diff
@@ -107,7 +106,7 @@ export function getRequestPaymentValues(request: ClientTypes.IRequestData): {
     expectedStartDate,
     tokensAccepted,
     maxRateTimespan,
-    network,
+    network: network ?? request.currencyInfo.network,
```
yomarion, line 109: Not related but it looks obvious and convenient.

packages/payment-processor/test/payment/erc20-fee-proxy-near.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        },
+      },
+    };
+
+    await expect(
+      payNearFungibleRequest(invalidRequest, mockedNearWalletConnection, undefined, undefined),
+    ).rejects.toThrowError(
+      'request cannot be processed, or is not an pn-erc20-fee-proxy-contract request',
+    );
+    expect(paymentSpy).toHaveBeenCalledTimes(0);
```
olivier7delf, line 160:
```suggestion
expect(paymentSpy).not.toHaveBeenCalled();
```

packages/payment-processor/test/payment/erc20-fee-proxy-near.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        network: 'aurora',
+      },
+    };
+
+    await expect(
+      payNearFungibleRequest(invalidRequest, mockedNearWalletConnection),
+    ).rejects.toThrowError(
+      'request cannot be processed, or is not an pn-erc20-fee-proxy-contract request',
+    );
+    expect(paymentSpy).toHaveBeenCalledTimes(0);
```
olivier7delf, line 185:
```suggestion
expect(paymentSpy).not.toHaveBeenCalled();
```

olivier7delf (approved):

packages/payment-processor/src/payment/near-fungible.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const amountToPay = getAmountToPay(request, amount).toString();
+
+  if (!(await isReceiverReady(walletConnection, request.currencyInfo.value, paymentAddress))) {
+    throw new Error(
+      `The paymentAddress is not registered for the token ${request.currencyInfo.value}`,
+    );
+  }
+  const proxyAddress = erc20FeeProxyArtifact.getAddress(network, 'near');
+  if (!(await isReceiverReady(walletConnection, request.currencyInfo.value, proxyAddress))) {
```
leoslr, line 47: Just to make it clear, everytime we support a new ERC20 on near, we will have to fund our proxy contract with  at least `MIN_STORAGE_FOR_FUNGIBLE` of this token, correct ?

packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -148,3 +164,92 @@ export const processNearPaymentWithConversion = async (
...
+      amount,
+      msg: JSON.stringify({
+        fee_address: feeAddress,
+        fee_amount: feeAmount,
+        payment_reference: paymentReference,
+        to,
+      }),
+    },
+    gas: GAS_LIMIT_CONVERSION_TO_NATIVE,
+    amount: '1000000000000000000000000 '.toString(), // 1 yoctoNEAR deposit is mandatory for ft_transfer_call
```
leoslr, line 197:
```suggestion
    amount: '1000000000000000000000000'.toString(), // 1 yoctoNEAR deposit is mandatory for ft_transfer_call
```

packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -28,14 +29,29 @@ export const isValidNearAddress = async (nearNetwork: Near, address: string): Pr
...
+  if (!token || token.type === RequestLogicTypes.CURRENCY.ETH) {
+    return nearWalletConnection
+      .account()
+      .state()
+      .then((state) => {
+        const balance = BigNumber.from(state?.amount ?? '0');
+        return balance.gte(amount);
+      });
+  }
+  if (token.type === RequestLogicTypes.CURRENCY.ERC20) {
```
leoslr, line 43: In this case, shouldn't we also check the account solvency in native currency for:
- The transaction costs
- The mandatory deposit of 1 yoctoNEAR

If you add the condition, having the associated test would be nice

packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -148,3 +164,92 @@ export const processNearPaymentWithConversion = async (
...
+const MIN_STORAGE_FOR_FUNGIBLE = '1250000000000000000000';
+
+/**
+ * Stores the minimum deposit amount on the `paymentAddress` account for `tokenAddress`.
+ * This does not check the existing deposit, if any, and should be called if `isReceiverReady` is false.
+ * @param walletConnection
+ * @param tokenAddress
+ * @param paymentAddress
+ */
+export const storageDeposit = async (
```
leoslr, line 217: Not directly related to this PR, but if the payee does not have the required storage deposit, how do we plan to handle this on the FE.
Is it the payer who will have to perform an additional transaction to send the storage deposit amount to the payee's account, and then perform the actual payment ?

alexandre-abrioux (approved):
Looks good to me!

@leoslr had good comments/questions, I'm also interested in those. Apart from that, approved 👌

packages/payment-processor/src/payment/near-fungible.ts
yomarion: @leoslr Yes, that's correct. I plan to check how other dapps do, especially, our contract should probably handle the storage in the same transaction, in a future version. cc @alexandre-abrioux

packages/payment-processor/src/payment/utils-near.ts
yomarion: That's a very good point, I did not think of that. The wallet UX is quite good when it comes to lack of NEAR (for both fees and deposit), so I think it's OK to let it handle this case.

yomarion: @leoslr  cc @alexandre-abrioux

packages/payment-processor/src/payment/utils-near.ts
yomarion: So that's a key question I haven't answered yet. I should check how other dapps do. I think we should assume that the payer pays for this storage in this case (and we should thus increase the transaction deposit).

yomarion: @leoslr  cc @alexandre-abrioux

PR #1097, opened 2023-04-12
feat: ERC20TransferableReceivable contract gas optimizations

mliu:
This pull request aims to cut gas costs on the ERC20TransferableReceivable contract. The changes proposed are based on the findings of a gas profiler and include the following optimizations:

- Remove the ERC721Enumerable functionality since it is not critical to the final product. Request Finance already knows all of a user's invoices and can map each one to their respective receivables. Previously we only used this to power our demo experience on the Huma dApp.
- Stop using ERC721URIStorage as a glorified storage field for requestID. This is a significant waste of gas since strings need to be processed and stored. Instead, revert to the base ERC721 URI storage and create a new field requestID in the ReceivableInfo struct.
    - This still allows for us to upgrade the receivables contract with metadata hosted on a storage service like IPFS.
      - One note for projects using this contract is that they won't be able to lookup the request ID directly from this contract. For Huma's integration, Request will pass the request ID during invoice factoring for Huma to verify receivables information.

Using hardhat-gas-reporter, I measured a decrease in mint call costs from 320k to 177k. These changes cut gas costs per mint by almost half, resulting in savings of approximately $8-10 per mint on mainnet with current gas prices. 

---

## Before
Screen Shot 2023-04-04 at 1 10 10 PM

## After
Screen Shot 2023-04-04 at 2 35 16 PM

### DIFF
--- a/packages/payment-processor/package.json
+++ b/packages/payment-processor/package.json
@@ -43,6 +43,7 @@
     "@requestnetwork/currency": "0.9.0",
     "@requestnetwork/payment-detection": "0.36.0",
     "@requestnetwork/smart-contracts": "0.29.0",
+    "@requestnetwork/multi-format": "0.15.10",
     "@requestnetwork/types": "0.36.0",
     "@requestnetwork/utils": "0.36.0",
     "@superfluid-finance/sdk-core": "0.5.0",
--- a/packages/payment-processor/package.json
+++ b/packages/payment-processor/package.json
@@ -43,7 +43,6 @@
     "@requestnetwork/currency": "0.9.0",
     "@requestnetwork/payment-detection": "0.36.0",
     "@requestnetwork/smart-contracts": "0.29.0",
-    "@requestnetwork/multi-format": "0.15.10",
     "@requestnetwork/types": "0.36.0",
     "@requestnetwork/utils": "0.36.0",
     "@superfluid-finance/sdk-core": "0.5.0",
--- a/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
+++ b/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
@@ -29,6 +29,7 @@ contract ERC20TransferableReceivable is ERC721 {
 
   /**
    * @notice Mapping for looking up a receivable given a paymentReference and minter address
+   * @dev Prevents minting multiple tokens for a given paymentReference and minter address
    */
   mapping(bytes32 => uint256) public receivableTokenIdMapping;
 
--- a/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
+++ b/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
@@ -28,8 +28,8 @@ contract ERC20TransferableReceivable is ERC721 {
   }
 
   /**
-   * @notice Mapping for looking up a receivable given a paymentReference and minter address
-   * @dev Prevents minting multiple tokens for a given paymentReference and minter address
+   * @notice Mapping for looking up a receivable given an initial owner address and paymentReference
+   * @dev Prevents minting multiple tokens for a given initial owner address and paymentReference
    */
   mapping(bytes32 => uint256) public receivableTokenIdMapping;
 
--- a/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
+++ b/packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
@@ -2,8 +2,7 @@
 pragma solidity ^0.8.0;
 
 import '@openzeppelin/contracts/utils/Counters.sol';
-import '@openzeppelin/contracts/token/ERC721/extensions/ERC721URIStorage.sol';
-import '@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol';
+import '@openzeppelin/contracts/token/ERC721/ERC721.sol';
 
 /**
  * @title ERC20TransferableReceivable
@@ -11,7 +10,7 @@ import '@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol';
  * @dev ERC721 contract for creating and managing unique NFTs representing receivables
  *      that can be paid with any ERC20 token
  */
-contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStorage {
+contract ERC20TransferableReceivable is ERC721 {
   using Counters for Counters.Counter;
 
   /**
@@ -31,6 +30,7 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
     address tokenAddress;
     uint256 amount;
     uint256 balance;
+    bytes32 requestID;
   }
 
   /**
@@ -158,7 +158,8 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
    * @param paymentReference A reference for the payment.
    * @param amount The amount of ERC20 tokens to be paid.
    * @param erc20Addr The address of the ERC20 token to be used as payment.
-   * @param newTokenURI The URI to be set on the minted receivable token.
+   * @param requestID The ID of the request associated with the receivable.
+   *                  Useful for retrieving details of the request from the Request Network.
    * @dev Anyone can pay for the mint of a receivable on behalf of a user
    */
   function mint(
@@ -166,7 +167,7 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
     bytes calldata paymentReference,
     uint256 amount,
     address erc20Addr,
-    string memory newTokenURI
+    bytes32 requestID
   ) external {
     require(paymentReference.length > 0, 'Zero paymentReference provided');
     require(amount > 0, 'Zero amount provided');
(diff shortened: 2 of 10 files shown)


packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -158,15 +158,16 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
...
    * @param amount The amount of ERC20 tokens to be paid.
    * @param erc20Addr The address of the ERC20 token to be used as payment.
-   * @param newTokenURI The URI to be set on the minted receivable token.
+   * @param requestID The ID of the request associated with the receivable.
+   *                  Useful for retrieving details of the request from the Request Network.
    * @dev Anyone can pay for the mint of a receivable on behalf of a user
    */
   function mint(
     address owner,
     bytes calldata paymentReference,
```
MantisClone, line 167: Change Request: 

I'm struggling to understand why we need both `requestId` and `paymentReference` parameters when minting. I suspect we can save gas if we eliminated the `paymentReference` and used the `requestId` instead.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -158,15 +158,16 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
    * @param paymentReference A reference for the payment.
    * @param amount The amount of ERC20 tokens to be paid.
    * @param erc20Addr The address of the ERC20 token to be used as payment.
-   * @param newTokenURI The URI to be set on the minted receivable token.
+   * @param requestID The ID of the request associated with the receivable.
+   *                  Useful for retrieving details of the request from the Request Network.
    * @dev Anyone can pay for the mint of a receivable on behalf of a user
    */
   function mint(
     address owner,
```
MantisClone, line 166: Change Request: 

I find the `owner` argument name confusing. `initialOwner` would be more clear.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -183,60 +184,10 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
     receivableInfoMapping[currentReceivableTokenId] = ReceivableInfo({
       tokenAddress: erc20Addr,
       amount: amount,
```
MantisClone, line 186: Comment:

I just realized that minters need to be very careful to ensure that the `tokenAddress` and `amount` provided when minting are correct and match the `owner` and `paymentReference`. If the minter uses a valid `owner` and `paymentReference` but accidentally uses an incorrect `tokenAddress` or `amount`, they would not be able to fix it. They'd have to start over with a new request before minting a new `TransferableReceivable` token.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -183,60 +184,10 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
...
-  /// @dev Overrides ERC721's supportsInterface method to include functionality from ERC721Enumerable.
-  function supportsInterface(bytes4 interfaceId)
-    public
-    view
-    override(ERC721, ERC721Enumerable)
-    returns (bool)
-  {
-    return super.supportsInterface(interfaceId);
   }
 }
```
MantisClone, line 193: Change Request: 

On [line 118](https://github.com/RequestNetwork/requestNetwork/pull/1097/files#diff-58d3632849e76bbd7f443b5978ff1800d7e3bbf7bd87b13d8152b238107229c2R118), the `paymentReference` argument is `bytes` but it could be restricted to `bytes8` which should save some gas. [ERC20FeeProxy Spec for reference](https://github.com/RequestNetwork/requestNetwork/blob/6ac8c287b8838af93e9384bc30352b334dba116a/packages/advanced-logic/specs/payment-network-erc20-fee-proxy-contract-0.1.0.md?plain=1#L14)
```solidity
bytes calldata paymentReference,
```

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -31,6 +30,7 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
     address tokenAddress;
     uint256 amount;
     uint256 balance;
+    bytes32 requestID;
   }
 
   /**
```
MantisClone, line 36: Change Request: 

Recommend adding a dev note for `receivableTokenIdMapping` (namely because this confused me for a while).

```
@dev Used by mint function to prevent minting multiple tokens for a given paymentReference and minter address
```

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -158,15 +158,16 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
    * @param paymentReference A reference for the payment.
```
MantisClone, line 158: Comment:

On [line 143 - 152](https://github.com/RequestNetwork/requestNetwork/pull/1097/files#diff-58d3632849e76bbd7f443b5978ff1800d7e3bbf7bd87b13d8152b238107229c2R143-R152) the `TransferableReceivablePayment` is emitted. Are all of these Event data fields absolutely necessary?

packages/payment-processor/package.json
```diff
@@ -43,6 +43,7 @@
     "@requestnetwork/currency": "0.9.0",
```
MantisClone, line 43: Change Request:

The description currently says:

> we store a bytes32 hash of the deserialized request ID

I think this is slightly misleading. We store the request ID which is a bytes32 hash of other data.

MantisClone (changes requested):
Nice work.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
yomarion: The only problem I see is that requestId-based detection would be inconsistent with all other payment extensions, they all use `paymentReference` instead. Maybe @vrolland can detail the scenario when knowing one's requestId is a security weakness?

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -2,16 +2,15 @@
 pragma solidity ^0.8.0;
```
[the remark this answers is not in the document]
mliu: Thank you for the thorough analysis!

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -183,60 +184,10 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
...
-    super._beforeTokenTransfer(from, to, tokenId);
-  }
-
-  /// @dev Overrides ERC721's _burn method to include functionality from ERC721URIStorage.
-  function _burn(uint256 tokenId) internal override(ERC721, ERC721URIStorage) {
-    super._burn(tokenId);
-  }
-
-  /// @dev Overrides ERC721's tokenURI method to include functionality from ERC721URIStorage.
-  function tokenURI(uint256 tokenId)
```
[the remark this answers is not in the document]
mliu: Since it's just for aesthetic purpose I think we could totally build a simple solution that makes these NFTs display nicer on marketplaces. A URL that took accepted IDs e.g. request.network/receivableImages/{tokenId} and redirected to a .svg file would be simple to build!

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
mliu: This is a good point, I can eliminate paymentReference here.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
mliu: We can remove paymentProxy, tokenAddress, and paymentId here to save some gas.

packages/payment-processor/package.json
mliu: Fixed

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: In the end, we decided to remove the `requestId` and keep the `paymentReference` - for privacy and consistency with other payment proxy contracts.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: I now know that other payment proxy smart contracts use `bytes` for the `paymentReference` but I don't understand why.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: I still think this would be a nice change. I wasn't able to add it as a suggestion because it's too far away from other edited lines.

MantisClone (changes requested):
Hey, @mliu Removal of the requestId looks great! A couple of my comments went unaddressed. Can you please address them?

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
mliu: No idea! I saw that as well and kept it as `bytes8` to stay consistent but paymentReference could definitely fit in a `bytes8`. What do you think I should use?

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
mliu: Sorry for not seeing this! Added

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: Would it be possible to benchmark how much gas is saved by using `bytes8` vs `bytes`? Consistency with other payment proxies has value but I want to understand how much we gain if we break that consistency.

MantisClone: Why did the payOwner # of calls increase from 10 to 11 between the tests?

mliu: One of the tests was to check that payOwner reverted if an empty bytes array was passed in, but I removed this test for the bytes8 benchmark. I don't think this would impact the benchmarks enough to warrant changing over to bytes8 though

MantisClone: Thanks for explaining. I agree that switching the `paymentReference` to `bytes8` doesn't make sense right now. It might make sense later if we migrated to a new version of the `ERC20FeeProxy` that also uses `bytes8`.

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
```diff
@@ -35,6 +29,7 @@ contract ERC20TransferableReceivable is ERC721, ERC721Enumerable, ERC721URIStora
 
   /**
    * @notice Mapping for looking up a receivable given a paymentReference and minter address
+   * @dev Prevents minting multiple tokens for a given paymentReference and minter address
```
MantisClone, line 32: Ever since #1091, the "minter" can be different from the initial owner. Please update the comment to make this clear. Also, please rearrange to match packing order `abi.encodePacked(owner, paymentReference)`

```suggestion
   * @notice Mapping for looking up a receivable given an initial owner address and paymentReference
   * @dev Prevents minting multiple tokens for a given initial owner address and paymentReference
```

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: I'm torn about it, but I guess I'm fine with not making this change given that it would have ripple effects through the payment-processing package.

MantisClone (changes requested):

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
mliu: Good catch!

MantisClone (approved):

yomarion (approved):

packages/smart-contracts/src/contracts/ERC20TransferableReceivable.sol
MantisClone: I learned from @vrolland that there's no good reason why `bytes` was used instead of `bytes8`. It just wasn't considered at the time. But now that all the payment proxies follow the same pattern, changing it would mean deploying new versions and migrating. I created #1113 to track this issue.

PR #1116, opened 2023-06-20
chore: remove prototype estimator

benjlevesque:
Unused package


alexandre-abrioux (approved):

yomarion (approved):

MantisClone (approved):
I'm late but LGTM! 
Out of curiosity, do we have tests that measure IPFS storage size and system throughput?

benjlevesque:
not that I know of @MantisClone 

in my 5 years working on the protocol, I've never needed or ran this package 
it's not running any automated tests
but it's in our way every time we changed something in the codebase

PR #1117, opened 2023-06-20
feat!: drop legacy storage

benjlevesque:
_BREAKING CHANGE_: Dropping legacy indexing mechanism.

Changes:
- enforce usage of TheGraph as a backend for request-node
- drop all custom indexing code (mostly DataAccess)
- refactor DataAccess to rely on an Indexer
- refactor mock to use in-memory indexer
- reduce dependency to TheGraph by abstracting most of the code transformation. TheGraph dependency is now down to only the SubgraphClient.


MantisClone:
I recognize that the legacy indexing mechanism doesn't work on many chains, is difficult to maintain, and that no one wants to spend the time to fix it. For those reasons, I agree that removing it now is the right move. 

Nonetheless, would it be possible to leave injection points such that someone could plug in their own custom indexer into the protocol?

benjlevesque:
> I recognize that the legacy indexing mechanism doesn't work on many chains, is difficult to maintain, and that no one wants to spend the time to fix it. For those reasons, I agree that removing it now is the right move.
> 
> Nonetheless, would it be possible to leave injection points such that someone could plug in their own custom indexer into the protocol?

Yes @MantisClone; that would be best indeed! These injection points are still available, but not as the Node level (same as today). I've introduced a new layer called "indexer" that represents better the way things are (it would actually be possible to keep the old code with this indexer layer, but it's not efficient so better remove it). 

This PR is still draft because I'm trying to reduce the surface of TheGraph in the code; it should be down to only the Subgraph Client, graphql queries and types. The rest could be generalised to work with any type of indexer.

packages/request-node/src/server.ts
benjlevesque, line 1: this file now only creates the server object, but doesn't start it. `startNode` is called from `bin.ts`

packages/request-node/src/requestNode.ts
benjlevesque, line 1: `requestNodeBase.ts` moved to `requestNode.ts`

packages/data-access/src/data-read.ts
benjlevesque, line 1: moved from `packages/thegraph-data-access/src/data-access.ts`

packages/data-access/src/pending-store.ts
benjlevesque, line 1: moved from thegraph-data-access

packages/smart-contracts/scripts/deploy-payments.ts
```diff
@@ -10,10 +10,10 @@ import { uniswapV2RouterAddresses, jumpToNonce } from './utils';
 import { Contract } from 'ethers';
 // eslint-disable-next-line
 // @ts-ignore Cannot find module
-import { ChainlinkConversionPath } from '../src/types/ChainlinkConversionPath';
```
benjlevesque, line 13: I had lots of issues due to dependencies (probably because removing some deps changes the resolved version of other packages) and had to upgrade some. This change is related to the typechain upgrade

packages/request-node/src/server.ts
```diff
@@ -1,42 +1,29 @@
-#!/usr/bin/env node
```
benjlevesque, line 1: this file now only creates the server object, but doesn't start it. startNode is called from bin.ts

packages/data-access/src/data-read.ts
```diff
@@ -0,0 +1,194 @@
+import { DataAccessTypes, StorageTypes } from '@requestnetwork/types';
```
benjlevesque, line 1: moved from packages/thegraph-data-access/src/data-access.ts

packages/data-access/src/pending-store.ts
```diff
@@ -0,0 +1,32 @@
+import { DataAccessTypes } from '@requestnetwork/types';
```
benjlevesque, line 1: moved from packages/thegraph-data-access/src/data-access.ts

packages/ethereum-storage/src/ethereum-utils.ts
```diff
@@ -1,8 +1,7 @@
-import { CurrencyTypes, StorageTypes } from '@requestnetwork/types';
+import { StorageTypes } from '@requestnetwork/types';
 import * as config from './config';
 
 import { BigNumber } from 'ethers';
-import { EvmChains } from '@requestnetwork/currency';
```
benjlevesque, line 5: @alexandre-abrioux FYI. 
I'm trying to reduce inter-packages dependencies... The currency package doesn't make sense in the storage I think

packages/request-client.js/src/http-metamask-data-access.ts
```diff
@@ -106,26 +103,26 @@ export default class HttpMetaMaskDataAccess extends HttpDataAccess {
...
-    const tx = await this.submitterContract.submitHash(
+    const tx = await submitterContract.submitHash(
       ipfsHash,
       /* eslint-disable no-magic-numbers */
       ethers.utils.hexZeroPad(ethers.utils.hexlify(ipfsSize), 32),
       { value: fee },
     );
 
-    const ethBlock = await this.provider.getBlock(tx.blockNumber);
+    const ethBlock = await this.provider.getBlock(tx.blockNumber || -1);
```
yomarion, line 116:
```suggestion
    const ethBlock = await this.provider.getBlock(tx.blockNumber ?? -1);
```
Any reason for not sticking with the nullish coalescing? It has impacts in edge cases like blockNumber = 0. Even if unlikely, I thought we should stick with the good practice.

packages/request-node/test/getConfirmedTransaction.test.ts
```diff
@@ -60,7 +65,7 @@ describe('getConfirmedTransaction', () => {
     expect(serverResponse!.body.result).toMatchObject({});
     // 'getConfirmedTransaction request meta'
     expect(serverResponse!.body.meta.storageMeta.state).toBe('confirmed');
-  }, 11000);
+  }, 30000);
```
yomarion, line 68: Are the 19 additional seconds needed just for the mining?

packages/thegraph-data-access/src/subgraph-client.ts
```diff
@@ -1,51 +1,110 @@
...
     });
   }
 
-  public getChannelsByTopics(topics: string[]): Promise<TransactionsBody> {
-    return this.graphql.request<TransactionsBody>(GetChannelsByTopicsQuery, {
-      topics,
-    });
+  // FIXME: this should be possible to do in a single query to the subgraph,
+  // but currently one transaction doesn't contain topics from previous ones on the same channel.
+  // This could be fixed on the Subgraph indexer code for optimization.
```
yomarion, line 53: 👍 Very clear, thanks for documenting the fixme

yomarion (approved):
Massive!

packages/request-node/test/getConfirmedTransaction.test.ts
benjlevesque: indexing of the graph. I increased a bit more than actually needed, to avoid flakiness

packages/usage-examples/src/request-client-js-erc20-request.ts
```diff
@@ -65,7 +65,7 @@ requestNetwork
   .createRequest(createParams)
   .then((request) => {
     console.log('clear request:');
-    console.log(request);
+    console.log(request.requestId);
```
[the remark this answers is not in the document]
MantisClone: I agree with @yomarion - logging the whole request makes sense in usage-example context.

MantisClone (commented):
I'm about 1/4th of the way through the review.

packages/thegraph-data-access/src/types.ts
```diff
@@ -0,0 +1,9 @@
+import { StorageTypes } from '@requestnetwork/types';
+import { DataAccessBaseOptions } from '@requestnetwork/data-access';
+
+import { RequestInit } from 'graphql-request/dist/types.dom';
```
MantisClone, line 4: What is this? Why is this imported from dist?

MantisClone (approved):
Impressive refactor! Only a few minor comments.

packages/thegraph-data-access/src/types.ts
benjlevesque: because not exported from root :)
(nb: not new, moved)

packages/request-client.js/src/http-request-network.ts
```diff
@@ -9,8 +9,8 @@ import { AxiosRequestConfig } from 'axios';
 import { PaymentNetworkOptions } from '@requestnetwork/payment-detection';
 import RequestNetwork from './api/request-network';
 import HttpDataAccess from './http-data-access';
-import MockDataAccess from './mock-data-access';
-import MockStorage from './mock-storage';
+import { MockDataAccess } from '@requestnetwork/data-access';
```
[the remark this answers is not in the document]
benjlevesque: Not sure what you mean, this is already done like this I think? can you see any place where it's not?

The change in this file is due to the Mock Storage not being in the request-client anymore

README.md
```diff
@@ -41,12 +41,11 @@ Join the [Request Discord][request-discord-url] to get in touch with us.
 
 ### Private Packages
 
-| Package                                                                | Description                                                 |
-| ---------------------------------------------------------------------- | ----------------------------------------------------------- |
-| [`@requestnetwork/integration-test`](/packages/integration-test)       | Integration test for the Request system                     |
-| [`@requestnetwork/prototype-estimator`](/packages/prototype-estimator) | Give estimates of size and throughput of the Request system |
```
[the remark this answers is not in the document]
benjlevesque: I've never seen it used. I'm in favor of dropping unused parts that cause more maintenance than they are used

packages/request-client.js/src/http-request-network.ts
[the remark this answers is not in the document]
MantisClone: I think there are places where things are imported from sibling source files instead of via a package dependency.  I'll find and share an example.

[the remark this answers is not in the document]
MantisClone: I went searching and even went so far as to enable an eslint rule [`monorepo/no-relative-import`](https://github.com/azz/eslint-plugin-monorepo#monorepono-relative-import-fixable) but I couldn't find any examples.  I stand corrected 😳

packages/request-node/src/dataAccess.ts
```diff
@@ -0,0 +1,50 @@
...
+  ipfsStorage: StorageTypes.IIpfsStorage,
+  logger: LogTypes.ILogger,
+): DataAccessTypes.IDataAccess {
+  const graphNodeUrl = config.getGraphNodeUrl();
+
+  const network = getEthereumStorageNetworkNameFromId(config.getStorageNetworkId()) as any;
+  if (!network) {
+    throw new Error(`Storage network not supported: ${config.getStorageNetworkId()}`);
+  }
+  EvmChains.assertChainSupported(network);
```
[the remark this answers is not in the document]
benjlevesque: I agree. I'd like to change this a little bit, in another PR. 
This call is there mostly for type assertion

alexandre-abrioux (approved):
👏

packages/thegraph-data-access/src/subgraph-client.ts
benjlevesque: Issue https://github.com/RequestNetwork/storage-subgraph/issues/9

benjlevesque:
NB: from now on, the `IPFS_URL` environment replaces `IPFS_HOST`, `IPFS_PORT` and `IPFS_PROTOCOL`

PR #1121, opened 2023-06-30
fix(gas-fee): catch estimation errors

alexandre-abrioux:
## Description of the changes

It can happen, rarely, that `@rainbow-me/fee-suggestions` fails to estimate gas fees, especially on chains other than Ethereum mainnet (related to https://github.com/rainbow-me/fee-suggestions/issues/23).

This PR catches,  logs, and handles those errors by returning an empty object, thus letting the RPC provider choose how to configure gas fee / gas price (either by using defaults for the EIP-1559 transaction fields or falling back to a legacy, non-EIP 1559, transaction).

### DIFF
--- a/packages/utils/src/estimate-gas-fees.ts
+++ b/packages/utils/src/estimate-gas-fees.ts
@@ -1,6 +1,7 @@
 import { BigNumber, constants, providers } from 'ethers';
 import { suggestFees } from '@rainbow-me/fee-suggestions';
 import { maxBigNumber } from './index';
+import { LogTypes } from '@requestnetwork/types';
 
 /**
  * The function estimates gas fee with EIP-1559.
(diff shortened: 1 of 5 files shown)


packages/utils/src/estimate-gas-fees.ts
```diff
@@ -11,36 +12,45 @@ import { maxBigNumber } from './index';
...
+      );
+      return {};
+    }
+
+    return {
+      maxPriorityFeePerGas,
+      maxFeePerGas,
+    };
+  } catch (e) {
+    logger.error(`estimateGasFees error: ${e}`);
```
benjlevesque, line 51: will this log correctly and not `[Object object]`?

benjlevesque (approved):
great. If you need to test, [this block](https://gnosisscan.io/block/28704480) has the issue

kevindavee (approved):

packages/utils/src/estimate-gas-fees.ts
alexandre-abrioux: I've added a test, all good 👍

benjlevesque (approved):

PR #1115, opened 2023-06-20
refactor(request node): config

benjlevesque:
Cleanup the configuration file of the request node, for better use of typescript and better readability

PR #1125, opened 2023-07-12
fix(request-node): log config

benjlevesque:
The latest version of the node doesn't log anything due to a bug introduced in #1115

### DIFF
--- a/packages/request-node/src/config.ts
+++ b/packages/request-node/src/config.ts
@@ -197,10 +197,7 @@ export const getLogLevel = (): LogTypes.LogLevel => {
 };
 
 /** logMode defines the log format to display: `human` is a more readable log, `machine` is better for parsing */
-export const getLogMode = (): LogMode => {
-  const logModeStr = getOption<keyof typeof LogMode>('logMode', 'LOG_MODE');
-  return LogMode[logModeStr] || defaultValues.log.mode;
-};
+export const getLogMode = makeOption('logMode', 'LOG_MODE', defaultValues.log.mode);
 
 /**
  * Get the minimum delay between getLastBlockNumber calls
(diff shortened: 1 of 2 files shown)


kevindavee (approved):

packages/request-node/src/config.ts
```diff
@@ -190,19 +190,14 @@ export function getMnemonic(): string {
...
 }
 
-/**
- * Get log configs: level and mode, from command line argument, environment variables or default values.
- * logLevel is the maximum level of messages we will log
- * logMode defines the log format to display: `human` is a more readable log, `machine` is better for parsing
- *
- * @returns the log level
- */
-export function getLogConfig(): { logLevel: LogTypes.LogLevel; logMode: LogMode } {
```
benjlevesque, line 200: removing this beacuse I'm dropping it in my next PR, this is not the fix

packages/request-node/src/config.ts
```diff
@@ -190,19 +190,14 @@ export function getMnemonic(): string {
...
-  return {
-    logLevel: getOption('logLevel', 'LOG_LEVEL', defaultValues.log.level),
-    logMode: getOption('logMode', 'LOG_MODE', defaultValues.log.mode),
-  };
-}
+/** logLevel is the maximum level of messages we will log */
+export const getLogLevel = (): LogTypes.LogLevel => {
+  const logLevelStr = getOption<keyof typeof LogTypes.LogLevel>('logLevel', 'LOG_LEVEL');
+  return LogTypes.LogLevel[logLevelStr] || defaultValues.log.level;
```
benjlevesque, line 196: this is the fix

MantisClone:
The bug was that `LogLevel` is an enum backed by a number but the `LOG_LEVEL` env variable takes `DEBUG`, not `0`.

MantisClone:
I deployed the Goerli Gateway and the logging is fixed. :tada: 
image

PR #1171, opened 2023-09-18
fix: remove hardcoded chainlink rate age (#1171)

### LINKED ISSUE #1144: Fix `encodeSwapToPayAnyToErc20Request()`: Don't hardcode `_chainlinkMaxRateTimespan` to 0
# Problem

SDK function `encodeSwapToPayAnyToErc20Request` hardcodes `_chainlinkMaxRateTimespan` to 0 when calling the `swapTransferWithReference()` function on the `ERC20SwapToConversion` contract. https://github.com/RequestNetwork/requestNetwork/blob/d539717009de82fe1195a73cb9f4ddbb0f59cd46/packages/payment-processor/src/payment/swap-any-to-erc20.ts#L170

This means that the caller cannot specify their tolerance for stale price feed data - they always accept infinitely old price feed data.

# Solution
Callers should be allowed to set their own value for `_chainlinkMaxRateTimespan` and therefore set their own tolerance for stale price feed data.

---

Migrated from Asana: https://app.asana.com/0/1203912381456855/1205377095268375/f

OjusWiZard:
Fixes #1144

Hi, a new contributor here!
Trying to fix the issue #1144

## Description of the changes
I've added an optional parameter `maxRateAge` in ~`ISwapSettings`~ `IConversionSettings`. It defaults to 0.
We can keep it required too, wdyt?

### DIFF
--- a/packages/payment-processor/src/payment/swap-any-to-erc20.ts
+++ b/packages/payment-processor/src/payment/swap-any-to-erc20.ts
@@ -167,6 +167,6 @@ export function encodeSwapToPayAnyToErc20Request(
     feeToPay, // _requestFeeAmount: BigNumberish,
     feeAddress || constants.AddressZero, // _feeAddress: string,
     Math.round(swapSettings.deadline / 1000), // _uniswapDeadline: BigNumberish,
-    0, // _chainlinkMaxRateTimespan: BigNumberish,
+    swapSettings.maxRateAge || 0, // _chainlinkMaxRateTimespan: BigNumberish,
   ]);
 }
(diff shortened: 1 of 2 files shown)


packages/payment-processor/src/payment/swap-any-to-erc20.ts
```diff
@@ -167,6 +167,6 @@ export function encodeSwapToPayAnyToErc20Request(
     feeToPay, // _requestFeeAmount: BigNumberish,
     feeAddress || constants.AddressZero, // _feeAddress: string,
     Math.round(swapSettings.deadline / 1000), // _uniswapDeadline: BigNumberish,
-    0, // _chainlinkMaxRateTimespan: BigNumberish,
+    swapSettings.maxRateAge || 0, // _chainlinkMaxRateTimespan: BigNumberish,
```
yomarion, line 170:
```suggestion
    swapSettings.maxRateAge ?? 0, // _chainlinkMaxRateTimespan: BigNumberish,
```

yomarion (commented):
Thanks for your contribution! Shouldn't this new param come from `conversionSettings` instead of `swapSettings`?

OjusWiZard:
> Thanks for your contribution! Shouldn't this new param come from `conversionSettings` instead of `swapSettings`?

That makes sense. Thanks for the review!

MantisClone (approved):
Thank you for your contribution!

PR #1180, opened 2023-09-29
fix: storage check for USDC on Near is custom

yomarion:
## Description of the changes

The NEAR-USDC storage check (and initiator) are not named according to the [standard](https://docs.near.org/develop/relevant-contracts/ft).


benjlevesque (approved):
this could use a comment/link to doc referring to this problem

yomarion:
> this could use a comment/link to doc referring to this problem

@benjlevesque , I agree but I couldn't find any 🤷

MantisClone (approved):

PR #1184, opened 2023-10-02
fix: solana address validation

leoslr:
## Description of the changes

The validation was broken when trying to validate ERC20 currencies on Solana

### DIFF
--- a/packages/currency/src/currencyManager.ts
+++ b/packages/currency/src/currencyManager.ts
@@ -246,7 +246,7 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
       case RequestLogicTypes.CURRENCY.ERC777:
         if (NearChains.isChainSupported(currency.network)) {
           return isValidNearAddress(address, currency.network);
-        } else if (currency.network === 'tron') {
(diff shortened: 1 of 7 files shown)


benjlevesque (approved):

packages/currency/src/currencyManager.ts
```diff
@@ -246,7 +246,7 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
       case RequestLogicTypes.CURRENCY.ERC777:
         if (NearChains.isChainSupported(currency.network)) {
           return isValidNearAddress(address, currency.network);
-        } else if (currency.network === 'tron') {
+        } else if (currency.network === 'tron' || currency.network === 'solana') {
```
alexandre-abrioux, line 249: @leoslr Is it the fix or is the change of version of `multicoin-address-validator` also needed?

alexandre-abrioux (approved):

packages/currency/src/currencyManager.ts
leoslr: Both are needed:
 - We need to tackle solana case here
 - The lib update was needed as well, version 5.2 didn't handle solana

MantisClone (approved):

PR #1196, opened 2023-10-11
feat: request network add sdk package

### LINKED ISSUE #1181: Create a new "request-network-sdk" package that bundles all the packages for interacting with the protocol.
# Problem(s)

1. It's unclear what the term "Request Network SDK" refers to. It's only an internal term that we use to describe the Request Network packages.
2. Installing the Request Network SDK requires many separate package installations. 

# Possible Solution

Create a new "request-network-sdk" package that bundles all of the packages that Builders typically install so that they only need to install 1 package to get all of the Request Network functionality in their app. The new package should be on the requestNetwork monorepo, not in a new repository.

# What if they don't want ALL the functionality in their app?

We can tree-shake so the unused packages are removed from the final app.
We can't yet tree-shake so the unused modules within packages are removed from the final app. #1148 will add this ability to tree-shake at the module level.

skiv71:
Fixes #1181

## Changes


packages/request-network-sdk/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/RequestNetwork/requestNetwork.git"
+  },
+  "homepage": "https://github.com/RequestNetwork/requestNetwork/tree/master/packages/request-network-sdk.js#readme",
+  "bugs": {
+    "url": "https://github.com/RequestNetwork/requestNetwork/issues"
+  },
+  "license": "MIT",
+  "dependencies": {
```
benjlevesque, line 35: the references in here should match the version in the `package.json` of these packages. Right now, you are resolving the version on NPM, not the monorepo's version

skiv71: Hi @benjlevesque thanks for this info, it really helped. Turned out it was because I used the @next modifier (as this has been promoted since I started).

Changed to the correct version(s) and all it well.

The latest commit also includes tsconfig changes to produce es6 output.

packages/request-network-sdk/README.md
```diff
@@ -0,0 +1,46 @@
+# @requestnetwork/request-network-sdk
+
+`@requestnetwork/request-network-sdk` is a ES6 typescript library part of the [Request Network protocol](https://github.com/RequestNetwork/requestNetwork).
+This package allows you to interact with the Request blockchain through [Request nodes](https://github.com/RequestNetwork/requestNetwork/blob/master/packages/request-node). This client side library uses Request nodes as servers, connected in HTTP. See the Request node documentation for more details on their API.
```
alexandre-abrioux, line 4:
```suggestion
This package lets you interact with the Request network through [Request Nodes](https://github.com/RequestNetwork/requestNetwork/blob/master/packages/request-node). This client-side library uses Request Nodes as its backend, connected in HTTP. See the [Request Node's API documentation](https://github.com/RequestNetwork/requestNetwork/tree/master/packages/request-node) for more details.
```

alexandre-abrioux (approved):

MantisClone:
@skiv71 The parent issue #1181 has been closed as not planned, so I'm closing this PR.

PR #1213, opened 2023-10-23
feat: payment processor batch payment native token

leoslr:
## Description of the changes

Payment-processor supports for batch payment of ETH requests with / without conversion.

Additional change:
Add and use `ETH-private` instead of using `ETH-mainnet` on local chain.


packages/payment-processor/test/payment/swap-erc20-fee-proxy.test.ts
```diff
@@ -197,7 +197,7 @@ describe('swap-erc20-fee-proxy', () => {
 
       // Swap and pay
       const tx = await swapErc20FeeProxyRequest(validRequest, wallet, {
-        deadline: Date.now() + 10000,
+        deadline: Date.now() + 1000000,
```
alexandre-abrioux, line 200: Do you remember why you needed to make this change?

alexandre-abrioux (approved):

packages/payment-processor/test/payment/batch-proxy.test.ts
```diff
@@ -430,16 +478,70 @@ describe('erc20-batch-conversion-proxy', () => {
...
             // Calculation of expectedAmountToPay when there there is no fee USD limit
-            //   expectedAmount:      1.00
-            //   feeAmount:        +   .02
-            //                     =  1.02
+            //   expectedAmount:    55 000
+            //   feeAmount:        +     2
             //   AggEurUsd.sol     x  1.20
             //   AggDaiUsd.sol     /  1.01
             //   BATCH_CONV_FEE      x  1.003
-            //   (exact result)    =  1.215516831683168316 (over 18 decimals for this ERC20)
```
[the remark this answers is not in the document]
leoslr: The comments were wrong: the request amount and fees were 55000 and 2. 
I removed the one about the exact results, since it was wrong as well. Didn't add a new exact result as I didn't find it brings much value since we already have the detailed calculus

packages/payment-processor/test/payment/swap-erc20-fee-proxy.test.ts
leoslr: Sometimes the tests were too slow, and this one failed because of expired `deadline`

olivier7delf (approved):

olivier7delf:
Great job Leo!
An additional test with ERC and ETH requests paid within a batch could be useful

leoslr:
> Great job Leo! An additional test with ERC and ETH requests paid within a batch could be useful

It's done in this test `should pay heterogeneous payments (ETH/ERC20 with/without conversion)`

PR #1221, opened 2023-10-27
fix(payment-processor): USDC on Near is now standard

yomarion:
## Description of the changes
USDC appears to be standard on Near.

I could not find sources for this breaking change but executing the standard method `storage_balance_of` now works and `check_registration` breaks.


kevindavee (approved):

benjlevesque (approved):
well, nice!

KolevDarko (approved):

olivier7delf (approved):

PR #1224, opened 2023-10-27
fix: persistTransaction timeout

### LINKED ISSUE #1145: Fix Request Node `/waitForConfirmation` endpoint. It only checks `store`, should fallback to query subgraph if confirmed transaction not in `store`
# Problem
The `/waitForConfirmation` endpoint only looks in the Request Node's store to see if the request has been confirmed.
https://github.com/RequestNetwork/requestNetwork/blob/2ff5b09c5cc859e190922d0c17163e6060489bed/packages/request-node/src/request/getConfirmedTransactionHandler.ts#L26-L33

But if the subgraph confirmation times out after 100 seconds, the confirmed transaction is never added to the confirmedTransactionStore 
https://github.com/RequestNetwork/requestNetwork/blob/2ff5b09c5cc859e190922d0c17163e6060489bed/packages/thegraph-data-access/src/data-access.ts#L282-L282
https://github.com/RequestNetwork/requestNetwork/blob/2ff5b09c5cc859e190922d0c17163e6060489bed/packages/request-node/src/request/persistTransaction.ts#L86-L100

As a result, the `/waitForConfirmation` endpoint will never return a confirmed transaction if it takes longer than 100 seconds for The Graph node to index the new request.
# Bandaid
Use a faster Ethereum Node RPC in The Graph Node so it never takes longer than 100 seconds to index a new request.
# Solution
The `/waitForConfirmation` endpoint should query the storage subgraph if it cannot find the confirmed transaction in the store .

---

Migrated from Asana: https://app.asana.com/0/1203912381456855/1205385812620545/f

MantisClone:
Resolves #1145 

# Context
The `persistTransaction()` method of the `http-data-access` implementation works like this:

1. it calls the `request-node` to persist a new transaction with `POST /persistTransaction`
2. then it waits for the transaction to be confirmed in a retry loop: it calls `GET /getConfirmedTransaction` multiple times until it does not receive an HTTP error anymore (normally it's NOT FOUND 404). By default, it polls 500 times with a 3-second delay for a total of 25 minutes! :scream: 
3. Meanwhile, the `request-node` tries to confirm the transaction via the `storage-subgraph`. However, if the graph node is running slow, perhaps due to a rate-limited Ethereum Node RPC, this transaction confirmation will timeout after 30 seconds.
4. IF the confirmation via the storage subgraph times out, the `request-node` never again tries to confirm it, and the `/getConfirmedTransaction` will forever return NOT FOUND 404

# Changes

* Reduce the default `getConfirmationMaxRetry` from 500 to 30 and the default `getConfirmationRetryDelay` from 3 seconds to 1 second.
* Improve error logging `request-node` server side to better explain what happened in the case of a timeout.
* Improve error logging on the `http-data-access` client side to explain to the `persistTransaction()` caller that they must continue polling the `getTransactionsByChannelId()` if they still need to confirm that their transaction was stored.
* Improve error handling in the `request-node`: Store an `Error` in the `confirmedTransactionStore` when the `POST /persistTransaction` endpoint fails. Anyone who calls the `GET /getConfirmedTransaction` can see the `Error`.

# TODO

~~- [ ] Add tests~~


MantisClone:
I converted to Draft because this change probably warrants some tests.

yomarion (approved):

alexandre-abrioux (approved):

MantisClone:
I decided to skip writing tests.

Issue #1227, opened 2023-11-02
Node v18 migration impossible with eccrypto


alexandre-abrioux:
# Issue

Migrating a codebase that uses the Request Network library to Node.js 18 is virtually impossible because of the dependency [eccrypto](https://github.com/bitchan/eccrypto).

# Context

## Native version incompatibility

`eccrypto` contains a "native" [C++ binary](https://github.com/bitchan/eccrypto/blob/master/ecdh.cc) that is compiled during installation of the package. This C++ code depends on the [nan.h](https://github.com/nodejs/nan/blob/main/nan.h) header file developed by Node and that contains native abstractions for Node.js. This header file has received breaking changes in Node v18, and compiling the C++ code now gives the following error:
```bash
In file included from ../../nan/nan.h:2884:
../../nan/nan_typedarray_contents.h: In constructor 'Nan::TypedArrayContents<T>::TypedArrayContents(v8::Local<v8::Value>)':
../../nan/nan_typedarray_contents.h:34:43: error: 'class v8::ArrayBuffer' has no member named 'GetContents'
   34 |       data   = static_cast<char*>(buffer->GetContents().Data()) + byte_offset;
      |                                           ^~~~~~~~~~~
```
This is also described here: https://github.com/bitchan/eccrypto/issues/94

## Browser version incompatibility

The issue described above should not, in theory, cause too much trouble because `eccrypto` also contains a [browser](https://github.com/bitchan/eccrypto/blob/master/browser.js) (JS-only) implementation that is also the fallback on Node.js in case the C++ binary is not found (see [here](https://github.com/bitchan/eccrypto/blob/a4f4a5f85ef5aa1776dfa1b7801cad808264a19c/index.js#L23)).

However, because of an issue with the library, data encrypted with the Native C++ module has a high probability of not being decryptable by the browser (JS-only) implementation. This means that data encrypted by Node.js before the upgrade to Node18, with the help of the C++ binary, cannot be decrypted after the upgrade to Node18 because we're now forced to use the browser (JS-only) fallback implementation.

This is also described here:
- https://github.com/bitchan/eccrypto/issues/56
- https://github.com/bitchan/eccrypto/issues/63
- this one describes it best: https://github.com/bitchan/eccrypto/issues/81
- (probably also related) https://github.com/bitchan/eccrypto/issues/90

# How to reproduce

## Native version incompatibility

With the following Dockerfile:

```Dockerfile
[...]

Issue #1207, opened 2023-10-21
Migrate from C implementation of eccrypto to JavaScript implementation eccrypto-js


MantisClone:
# Problem

C implementation of eccrypto requires compilation that is slow and takes a lot of time to build.

# Proposed Solution

Use a pure JavaScript implementation like [eccrypto-js](https://github.com/pedrouid/eccrypto-js)

alexandre-abrioux:
This issue seems more important than before since the discovery of https://github.com/RequestNetwork/requestNetwork/issues/1227. I will try to have a look at it soon.

Issue #1178, opened 2023-09-27
Parameterize CircleCI build for Node 16 and 18


MantisClone:
# Problem

The CircleCI build doesn't test Node v18. The fear is that we'll introduce a regression that breaks either v16 or v18 compatibility.

# Details

Node 16 is EOL so maybe we should deprecate it and remove support for Node 16 instead of keeping it.
https://endoflife.date/nodejs

# Solution

Parameterize the build step using a [matrix job](https://circleci.com/docs/using-matrix-jobs/) to take in a variable for the node docker image.

MantisClone:
This was closed as unplanned because Node 16 support was dropped in https://github.com/RequestNetwork/requestNetwork/pull/1229#issuecomment-1792651512

PR #1229, opened 2023-11-02
chore: eccrypto alternative

alexandre-abrioux:
## Description of the changes

This is an attempt to fix https://github.com/RequestNetwork/requestNetwork/issues/1227, and if possible, should also resolve https://github.com/RequestNetwork/requestNetwork/issues/1207.
It contains a new set of tests to ensure the transition from `eccrypto` covers all cases.

### eccrypto-js

What I tried but did **_not_** work was to replace `eccrypto` with `eccrypto-js`, like proposed in https://github.com/RequestNetwork/requestNetwork/issues/1207. Unfortunately, the `eccryptoNativeData` dataset could be decrypted but the `eccryptoBrowserData` could not. It looks like the library contains some of the same faults as `eccrypto` (see https://github.com/pedrouid/eccrypto-js/issues/34)

### @toruslabs/eccrypto

What worked was to replace `eccrypto` with [`@toruslabs/eccrypto`](https://www.npmjs.com/package/@toruslabs/eccrypto) which is a fork of the initial library, re-implemented in Typescript, that fixes the initial bugs, and that contains no native C++ binary. However, there was a requirement to move to Node18.


MantisClone:
It's worth noting that this change **drops support for Node 16**. It makes sense to do this because [Node 16 security support ended on September 11, 2023](https://endoflife.date/nodejs). Thus, I just closed #1178 as not planned.

MantisClone:
It's worth noting that this change **drops support for Node 16**. It makes sense to do this because [Node 16 security support ended on September 11, 2023](https://endoflife.date/nodejs). Thus, I just closed #1178 as not planned.

packages/utils/test/crypto/ec-utils.test.ts
```diff
@@ -159,6 +161,19 @@ describe('Utils/EcUtils', () => {
         ),
       ).rejects.toThrowError('The encrypted data is not well formatted');
     });
+
+    it('should be compatible with previous implementation of eccrypto', async () => {
+      const dataTypes = { native: eccryptoNativeData, browser: eccryptoBrowserData } as const;
+      for (const dataType in dataTypes) {
```
benjlevesque, line 167: use it.each?

benjlevesque (approved):
great work 👏

packages/utils/test/crypto/ec-utils.test.ts
alexandre-abrioux: done in https://github.com/RequestNetwork/requestNetwork/pull/1229/commits/9f792560b62e59850390b1c4011ad1c5a89a87d4

MantisClone (approved):
Looks good 👍 great work 💯

olivier7delf (approved):

PR #1231, opened 2023-11-07
ci: save test results in CircleCI

alexandre-abrioux:
## Description of the changes

This PR aims to store test results in CircleCI. This way, we can have access to test results directly in the "test" tab of each CircleCI job without the need to inspect the logs. This will also unlock CircleCI's Insights and we will have access to flaky test statistics.


benjlevesque (approved):
yeah! nice

MantisClone (approved):
Nice.

PR #1236, opened 2023-11-13
chore: migrate to @noble/curves and ecies/js

alexandre-abrioux:
## Description of the changes

- use `@noble/curves` for signature instead of `secp256k1`
- use `ecies/js` for encryption (which uses `@noble/curves` under the hood) instead of `eccrypto`'s fork
- `@noble/curves` is an audited library, and now the industry standard used by `ethers` and `viem`

## What's missing
For the moment we can't decrypto old data because `ecies/js` uses `AES-256-GCM` while `eccrypto` uses `AES-256-CBC`, see https://github.com/ecies/js/issues/747

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "@requestnetwork/types": "0.37.0",
     "@toruslabs/eccrypto": "4.0.0",
+    "eciesjs": "^0.4.5",
     "ethers": "5.5.1",
     "secp256k1": "4.0.2",
     "tslib": "2.5.0"
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -40,11 +40,10 @@
     "test:watch": "yarn test --watch"
   },
   "dependencies": {
+    "@noble/curves": "1.2.0",
     "@requestnetwork/types": "0.37.0",
-    "@toruslabs/eccrypto": "4.0.0",
-    "eciesjs": "^0.4.5",
+    "eciesjs": "0.4.5",
     "ethers": "5.5.1",
-    "secp256k1": "4.0.2",
     "tslib": "2.5.0"
   },
   "devDependencies": {
--- a/packages/utils/src/crypto/ec-utils.ts
+++ b/packages/utils/src/crypto/ec-utils.ts
@@ -1,6 +1,8 @@
 import { publicKeyConvert, ecdsaRecover } from 'secp256k1';
 import { ethers } from 'ethers';
-import { Ecies, decrypt, encrypt } from '@toruslabs/eccrypto';
+// import { Ecies, decrypt, encrypt } from '@toruslabs/eccrypto';
+
+import { decrypt, encrypt, ECIES_CONFIG } from 'eciesjs';
 
 /**
  * Function to manage Elliptic-curve cryptography
@@ -13,6 +15,11 @@ export {
   ecRecover,
   ecSign,
 };
+
+ECIES_CONFIG.ellipticCurve = 'secp256k1';
+ECIES_CONFIG.isEphemeralKeyCompressed = false;
+ECIES_CONFIG.symmetricAlgorithm = 'aes-256-gcm';
+ECIES_CONFIG.symmetricNonceLength = 16;
 
 /**
  * Function to derive the address from an EC private key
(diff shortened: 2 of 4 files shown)


MantisClone:
I thought we were using GCM also, according to the old encryption audit. Is the CBC code actually used for anything?

https://request.network/en/2020/07/07/request-encryption-audit-completed-by-cure53/
> Action: Use the authenticated encryption algorithm aes-256-gcm instead of aes-256-cbc and remove the hash.
> Status: fixed – [Commit 1](https://github.com/RequestNetwork/requestNetwork/commit/d58f101f9f76e408671dd1edb0d67863d1c8abd5) & [Commit 2](https://github.com/RequestNetwork/requestNetwork/commit/969bebeb99b4bc2fdd31405a162934cfdff6db05)

alexandre-abrioux:
@MantisClone There are two encryption stages. First, the request is encrypted using AES-256-GCM (symmetric encryption). The generated encryption key is later referred to as the channel key. This channel key is then encrypted with each of the shareholders' public keys using ECIES. This two-stage encryption scheme aims to reduce the size of the encrypted data; otherwise, we would have to encrypt the whole request as many times as there are shareholders. Instead, we encrypt the channel key multiple times, which is a minimal overhead. More details over [here](https://github.com/RequestNetwork/requestNetwork/blob/master/packages/transaction-manager/specs/encryption.md#description).

This PR only concentrates on the ECIES part. I didn't know this before, but ECIES also uses AES under the hood, and `eccrypto` uses `AES-256-CBC`. AEC "cbc" mode is not recommended because it has no "aead" tag, so I'm trying to find a way to upgrade while keeping compatibility for previously encrypted requests.

MantisClone:
@alexandre-abrioux now that ecies/js added support for AES-256-CBC in https://github.com/ecies/js/pull/748, do you still think that it's possible to migrate to @noble/curves and ecies/js?

Are we blocked by the fact that ecies/js does not support the SHA512 key derivation function?

https://github.com/ecies/js/issues/747#issuecomment-1809586196:
> Another point is that eccrypto is using sha512 (specifically the first 32 bytes of `SHA512(ECDH(ephermeralPublicKey, publicKey)`) to derive the shared key (https://github.com/bitchan/eccrypto/blob/master/index.js#L219 or https://github.com/torusresearch/eccrypto/blob/master/src/index.ts#L248) but we are using [HDKF-SHA256](https://github.com/ecies/py/issues/82), this also makes it incompatible.

https://github.com/ecies/js/issues/747#issuecomment-1809953172:
> **Since we don't have a plan to support SHA512 as a key derivation function, you may need to think about a migration strategy.**

packages/utils/src/crypto/ec-utils-legacy.ts
```diff
@@ -0,0 +1,81 @@
...
+const equalConstTime = (b1: Uint8Array, b2: Uint8Array): boolean => {
+  if (b1.length !== b2.length) {
+    return false;
+  }
+  let res = 0;
+  for (let i = 0; i < b1.length; i++) {
+    res |= b1[i] ^ b2[i];
+  }
+  return res === 0;
+};
```
alexandre-abrioux, line 45: Using the Web Crypto API (available in Node.js as well as browsers) this whole code (`hmacSha256Verify` + `equalConstTime`) could be replaced by:
```ts
globalThis.crypto.subtle.verify("HMAC", key, sig, msg)
```

### Pros:
- no need to use a library (`@noble/hashes`)
- the Web Crypto API is async so non-thread blocking

### Cons
- the Web Crypto API is async so the calling code needs to be async too (but that's OK since `ecEncrypt` and `ecDecrypt` were already async in the previous implementation)
- the Web Crypto API is only available in secure context (= HTTPS)

Let me know what you would prefer, I can easily make the switch if needed.

tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "strict": true,
     "baseUrl": "./packages",
     "skipLibCheck": true,
-    "moduleResolution": "node",
+    "moduleResolution": "nodenext",
```
alexandre-abrioux, line 13: Needed to load `@ecies/ciphers/aes`. The value `node` is deprecated anyway, we shouldn't use it, see https://www.typescriptlang.org/tsconfig/#moduleResolution

PR #1238, opened 2023-11-14
refactor: currency types

benjlevesque:
Move all types to the `types` package for better consistency with other types and as a first step to reduce unnecessary dependencies to the currency package.

BREAKING: types must be updated with `CurrencyTypes` prefix, imported from `@requestnetwork/types`  package.
BREAKING: the `currencies` field has been removed from the `RequestNetwork` client parameters. Use `currencyManager: new CurrencyManager(CURRENCIES)` instead.


packages/advanced-logic/src/extensions/payment-network/near/any-to-near-testnet.ts
```diff
@@ -1,8 +1,8 @@
-import { ICurrencyManager } from '@requestnetwork/currency';
+import { CurrencyTypes } from 'types/dist';
```
benjlevesque, line 1:
```suggestion
import { CurrencyTypes } from '@requestnetwork/types';
```

packages/currency/src/chains/evm/data/bsc.ts
```diff
@@ -1,7 +1,7 @@
-import { TokenMap } from '../../../types';
+import { CurrencyTypes } from 'types/dist';
```
benjlevesque, line 1:
```suggestion
import { CurrencyTypes } from '@reuqestnetwork/types';
```

packages/request-client.js/src/http-request-network.ts
```diff
@@ -41,8 +41,7 @@ export default class HttpRequestNetwork extends RequestNetwork {
       nodeConnectionConfig?: Partial<NodeConnectionConfig>;
       signatureProvider?: SignatureProviderTypes.ISignatureProvider;
       useMockStorage?: boolean;
-      currencies?: CurrencyInput[];
```
benjlevesque, line 44: Breaking: I suggest to remove this parameter as it's redundant with `currencyManager`

benjlevesque: between the two? I would say yes: it enables injecting your own implementation, and the CurrencyManager contains more than just a list of currencies. For instance, it contains the list of Aggregators for conversion

alexandre-abrioux (approved):

leoslr (approved):

Issue #1202, opened 2023-10-16
Reduce download time by maximizing static composition and dependency injection. Eliminate runtime configs.


MantisClone:
# Problem

Request Network uses config variables at runtime. This means that the package MUST contain code to support all possible configurations even if they're not used.

# Proposed Solution

Use dependency injection to empower the user to select the features and dependencies they want. This will involve API BREAKING CHANGES and refactoring package boundaries.

# Definition of Done

This issue is an EPIC that encompasses the following subtasks:

- [ ] #1211
- [ ] #1111
- [x] #1524
- [ ] #1525
- [ ] #1208 
- [ ] #1523

# Example

In the following code snippet, notice that the currencyManager, chains, and balanceDetectors are all injected.

```js
const requestNetwork = new RequestNetwork({
  decryptionProvider,
  signatureProvider,
  currencyManager,
  chains,
  // if no compatible balance detector is found, the balance will be null. 
  balanceDetectors: [
    new EvmTheGraphBalanceDetector(chains, currencyManager),
    new NearTheGraphBalanceDetector(chains),
    new BtcBalanceDetector(),
    new SolanaBalanceDetector(),
    new SuperfluidBalanceDetector(chains),
    // ...
  ]
})

```

# Reference

* [Request Network library refactoring](https://docs.google.com/document/d/19fxTvuI_LfZD0CqRte3lJZH51yKT6GPRGO0m0q9bQpo/edit?usp=sharing)

MantisClone:
If we (@skiv and I) were to rebuild this system from scratch, we would write it in a functional style, with less OOP and inheritance. We believe a functional style would lend itself to automated static analysis and tree-shaking. But we've already got a system that uses OOP and inheritance, so we're choosing to double down and make it manually composable via dependency injection so that the user chooses the packages that fit their use case.

MantisClone:
Closing epic in favor of sub-issues.

PR #1253, opened 2023-11-16
feat(payment-detection): add subgraphMinIndexedBlock option

alexandre-abrioux:


benjlevesque:
I feel the option is a bit awkward, I'd rather have it directly on the client, see my [proposal](https://github.com/RequestNetwork/requestNetwork/compare/subgraph-min-indexed-block...subgraph-client-optons?expand=1)

benjlevesque:
Also, I wonder if grouping the options is a good move, considering https://github.com/RequestNetwork/requestNetwork/issues/1202 and https://github.com/RequestNetwork/requestNetwork/issues/1211

PR #1255, opened 2023-11-17
ci: move lint out of build

benjlevesque:
the format & lint takes about a minute. It shouldn't be part of the `build` stage, but in a separate `lint` stage (together with package.json lint)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,12 +76,6 @@ jobs:
           paths:
             - ~/.cache/yarn
       - run:
-          name: Verify format
-          command: yarn format:check
-      - run:
-          name: Verify lint
-          command: yarn lint:check
-      - run:
           name: Build types
           command: yarn workspace @requestnetwork/types run build
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -102,14 +96,21 @@ jobs:
 
       - persist_to_workspace:
           root: *working_directory
-          paths: .
-  lint-package-json:
+          paths:
+            - ./
```
benjlevesque, line 100: just remove a warning from the json schema

.circleci/config.yml
```diff
@@ -102,14 +96,21 @@ jobs:
...
+  lint:
```
benjlevesque, line 101: rename job to match what it does

.circleci/config.yml
```diff
@@ -102,14 +96,21 @@ jobs:
...
     working_directory: *working_directory
     steps:
       - attach_workspace:
           at: *working_directory
+      - run:
+          name: Verify format
+          command: yarn format:check
+      - run:
+          name: Verify lint
+          command: yarn lint:check
```
benjlevesque, line 113: main change

.circleci/config.yml
```diff
@@ -714,9 +719,6 @@ workflows:
                 - master
     jobs:
       - build
-      - lint-package-json:
-          requires:
-            - build
```
benjlevesque, line 719: unnecessary in nighly build

MantisClone (approved):
LGTM. Thanks for the excellent comments 👍

alexandre-abrioux (approved):
👍

PR #1261, opened 2023-11-23
fix: zksync networks names to lowercase

KolevDarko:
## Description of the changes

We rely on that all network names are in lowercase in Invoicing, with the mixed letters the UI could not match the currency chain with the chainList from currency API.

Also added the batch contract address from zksync testnet.

### DIFF
--- a/packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
+++ b/packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
@@ -65,6 +65,10 @@ export const batchPaymentsArtifact = new ContractArtifact<BatchPayments>(
           address: '0x0C41700ee1B363DB2ebC1a985f65cAf6eC4b1023',
           creationBlockNumber: 19545614,
         },
+        zksynceratestnet: {
+          address: '0x276A92F78aA527b8e157B0E47cEB63b4239dfa0b',
+          creationBlockNumber: 13814862,
+        },
       },
     },
   },
--- a/packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
+++ b/packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
@@ -61,7 +61,7 @@ export const batchPaymentsArtifact = new ContractArtifact<BatchPayments>(
           address: '0x0DD57FFe83a53bCbd657e234B16A3e74fEDb8fBA',
           creationBlockNumber: 35498688,
         },
-        zkSyncEra: {
+        zksyncera: {
           address: '0x0C41700ee1B363DB2ebC1a985f65cAf6eC4b1023',
           creationBlockNumber: 19545614,
         },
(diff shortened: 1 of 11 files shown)


AdamDowson (approved):

leoslr (approved):
@KolevDarko Did you also update the contracts (Batch I think) to modify the `nativeHash` that was previously used ?

packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
```diff
@@ -61,10 +61,14 @@ export const batchPaymentsArtifact = new ContractArtifact<BatchPayments>(
           address: '0x0DD57FFe83a53bCbd657e234B16A3e74fEDb8fBA',
           creationBlockNumber: 35498688,
         },
-        zkSyncEra: {
+        zksyncera: {
           address: '0x0C41700ee1B363DB2ebC1a985f65cAf6eC4b1023',
```
MantisClone, line 65: 1. Why is this address identical to the one listed for zkSync Era [`BatchConversionPayments`](https://github.com/RequestNetwork/requestNetwork/pull/1261/files#diff-a7c454cdf39eb8aa49bbdbd7dd4ed16ab0ab8271d6fffb702c06ce687ee57101R68)

2. Why does the [zkSync Era explorer](https://explorer.zksync.io/address/0x0C41700ee1B363DB2ebC1a985f65cAf6eC4b1023#contract) show the contract name "BatchConversionPayments" instead of "BatchPayments"?
image

packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
```diff
@@ -61,10 +61,14 @@ export const batchPaymentsArtifact = new ContractArtifact<BatchPayments>(
...
           creationBlockNumber: 19545614,
         },
+        zksynceratestnet: {
+          address: '0x276A92F78aA527b8e157B0E47cEB63b4239dfa0b',
```
MantisClone, line 69: Why does the zkSync Era Goerli explorer show the contract name "BatchConversionPayments" instead of "BatchPayments"?
image

MantisClone (commented):
@KolevDarko @benjlevesque @leoslr I have a couple questions. I just want to make sure the correct addresses are documented in the lib/artifacts/../index.ts files.

packages/smart-contracts/src/lib/artifacts/BatchPayments/index.ts
KolevDarko: The BatchConversionPayments is the contract that is used in production, so that's why that one is deployed.

At the time I thought BatchPayments was the one we were using but if we see the block explorer on mainnet: https://etherscan.io/address/0x0dd57ffe83a53bcbd657e234b16a3e74fedb8fba it hasn't been used for more than a year.

So if you want to clean things up, we can just remove the entries from BatchPayments/index.ts for both prod and test zksync.

MantisClone: Added issue to fix :+1: https://github.com/RequestNetwork/requestNetwork/issues/1303

PR #1262, opened 2023-11-23
feat(payment-detection): add minIndexedBlock option for TheGraph

alexandre-abrioux:
supersedes https://github.com/RequestNetwork/requestNetwork/pull/1253

## Context

When using TheGraph decentralized network, different GQL queries can target different indexers. For consistency, we can ask to target only indexers that have processed specific blocks using a block filter.

For more info, see [here](https://thegraph.com/docs/en/cookbook/upgrading-a-subgraph/) or [here](https://thegraph.com/blog/how-to-migrate-ethereum-subgraph/), look for `number_gte` in these pages.

## Description of the changes

- Added a new `minIndexedBlock` option to TheGraph client in order to handle this

PR #1267, opened 2023-11-27
fix(payment-detection): differenciate decentralized network case

alexandre-abrioux:
This is a fixup for https://github.com/RequestNetwork/requestNetwork/pull/1262 and https://github.com/RequestNetwork/requestNetwork/pull/1265

## Context

When using TheGraph decentralized network, different GQL queries can target different indexers. For consistency, we can ask to target only indexers that have processed specific blocks using a block filter.

For more info, see [here](https://thegraph.com/docs/en/cookbook/upgrading-a-subgraph/) or [here](https://thegraph.com/blog/how-to-migrate-ethereum-subgraph/), look for `number_gte` in these pages.

## Description of the change

When we don't have the data for which minimum block an indexer should have processed, we must set a default value. After several unsuccessful tries, here is what I've found regarding generating a default value that works with both a normal TheGraph Node and the decentralized gateway:

| Block Filter                      | TheGraph Node                                             | TheGraph Gateway (decentralized)                                            |
|-----------------------------------|-----------------------------------------------------------|-----------------------------------------------------------------------------|
| `undefined`                         | ✅                                                         | ❌ `Invalid query: Failed to determine block constraints.`                     |
| `null`                         | ✅                                                         | ❌ `Invalid query: Failed to determine block constraints.`                     |
| empty object: `{}`                   | ❌ `Invalid value provided for argument block: Object({})` | ✅                                                                           |
| default to zero: `{ number_gte: 0 }` | ✅                                                         | ❌ `Invalid query: Requested block before minimum startBlock of manifest: 0` |

This shows that no default value is compatible with both setups. We need to differentiate between both. I've added a new `decentralizedNetwork` boolean option that auto-populates depending on the URL value and allows to distinguish between both cases.

### DIFF
--- a/packages/payment-detection/src/thegraph/client.ts
+++ b/packages/payment-detection/src/thegraph/client.ts
@@ -38,10 +38,23 @@ export type TheGraphClientOptions = {
   timeout?: number;
   /** constraint to select indexers that have at least parsed this block */
   minIndexedBlock?: number | undefined;
+  /** is this client targeting TheGraph decentralized network? */
+  decentralizedNetwork?: boolean;
 };
 
 const extractClientOptions = (options?: TheGraphClientOptions) => {
   return pick(options, 'timeout');
+};
+
+const populateSdkWithOptions = (
+  sdk: TheGraphClient<any>,
+  url: string,
+  options?: TheGraphClientOptions,
+) => {
+  sdk.options = {
+    decentralizedNetwork: url.startsWith('https://gateway-arbitrum.network.thegraph.com/'),
+    ...options,
+  };
 };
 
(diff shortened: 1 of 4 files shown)


KolevDarko (approved):

olivier7delf (approved):

benjlevesque (approved):
Wow, great find... Strange that TheGraph doesn't expect the same input on hosted vs Network 🤔  I wonder if this is intentional.

I'd prefer an approach where the complexity related to the Network is kept centralized in the `client` , and abstracted in the `info-retriever`, but your solution works just fine! For the record, [here](https://github.com/RequestNetwork/requestNetwork/compare/master...thegraph-client-proposal?expand=1)'s what I suggest, but not a blocker
Great work

PR #1268, opened 2023-11-28
feat(advanced-logic): use currency manager instance

kevindavee:
### Context
Initiating a `CurrencyManager` is expensive. I was debugging the `computeRequestId` function. I'm sharing the logs of my test.

Screenshot 2023-11-28 at 12 09 16

Based on my test, `computeRequestId` is executed for 130ms. I'm using a M1 Macbook Pro. External network calls (API/DB calls) only take around 20ms _(from verifying signature input until signed)_. So, we still have 110ms of processing time, and it's all local execution. I suspected two things: for loops or a computing-intensive operation such as signing/hashing/encrypting, etc. But apparently, the process that takes much time is `CurrencyManager.getDefault()`. You can see on the logs there are several `.getDefault()` calls, and it takes approximately 15ms. A total of 6 calls contribute 90ms from the 110ms local execution 🥲

### Change
Inject instantiated `currencyManager`. I think we should re-use the `currencyManager` that has been instantiated in the `AdvancedLogic` class and pass it down to the payment networks unless there are specific reasons why certain payment networks can't use the `currencyManager` from the `AdvancedLogic` class.

### DIFF
--- a/packages/advanced-logic/src/advanced-logic.ts
+++ b/packages/advanced-logic/src/advanced-logic.ts
@@ -56,26 +56,30 @@ export default class AdvancedLogic implements AdvancedLogicTypes.IAdvancedLogic 
     erc20TransferableReceivable: Erc20TransferableReceivablePaymentNetwork;
   };
 
+  private currencyManager: ICurrencyManager;
+
   constructor(currencyManager?: ICurrencyManager) {
     if (!currencyManager) {
       currencyManager = CurrencyManager.getDefault();
     }
+
+    this.currencyManager = currencyManager;
     this.extensions = {
       addressBasedBtc: new AddressBasedBtc(),
-      addressBasedErc20: new AddressBasedErc20(),
+      addressBasedErc20: new AddressBasedErc20(currencyManager),
       addressBasedTestnetBtc: new AddressBasedTestnetBtc(),
       contentData: new ContentData(),
       anyToErc20Proxy: new AnyToErc20Proxy(currencyManager),
       declarative: new Declarative(),
-      ethereumInputData: new EthereumInputData(),
-      feeProxyContractErc20: new FeeProxyContractErc20(),
-      proxyContractErc20: new ProxyContractErc20(),
-      erc777Stream: new Erc777Stream(),
-      feeProxyContractEth: new FeeProxyContractEth(),
+      ethereumInputData: new EthereumInputData(currencyManager),
+      feeProxyContractErc20: new FeeProxyContractErc20(currencyManager),
+      proxyContractErc20: new ProxyContractErc20(currencyManager),
+      erc777Stream: new Erc777Stream(currencyManager),
+      feeProxyContractEth: new FeeProxyContractEth(currencyManager),
       anyToEthProxy: new AnyToEthProxy(currencyManager),
-      nativeToken: [new NearNative(), new NearTestnetNative()],
+      nativeToken: [new NearNative(currencyManager), new NearTestnetNative(currencyManager)],
       anyToNativeToken: [new AnyToNear(currencyManager), new AnyToNearTestnet(currencyManager)],
-      erc20TransferableReceivable: new Erc20TransferableReceivablePaymentNetwork(),
+      erc20TransferableReceivable: new Erc20TransferableReceivablePaymentNetwork(currencyManager),
     };
   }
 
@@ -173,7 +177,7 @@ export default class AdvancedLogic implements AdvancedLogicTypes.IAdvancedLogic 
 
   public getFeeProxyContractErc20ForNetwork(network?: string): FeeProxyContractErc20 {
     return NearChains.isChainSupported(network)
-      ? new FeeProxyContractErc20(undefined, undefined, network)
+      ? new FeeProxyContractErc20(this.currencyManager, undefined, undefined, network)
       : this.extensions.feeProxyContractErc20;
   }
 
--- a/packages/advanced-logic/src/extensions/payment-network/address-based.ts
+++ b/packages/advanced-logic/src/extensions/payment-network/address-based.ts
@@ -1,4 +1,4 @@
-import { CurrencyManager, UnsupportedCurrencyError } from '@requestnetwork/currency';
+import { ICurrencyManager, UnsupportedCurrencyError } from '@requestnetwork/currency';
 import {
   CurrencyTypes,
   ExtensionTypes,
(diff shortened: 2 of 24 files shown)


KolevDarko (approved):
Good idea, I see there are a lot of for cycles when preparing the currencies.

Another way would be to store a static field the CurrencyManager itself, that way you would modify just one file. But this also works.

packages/request-client.js/test/index.test.ts
```diff
@@ -2031,10 +2032,19 @@ describe('request-client.js', () => {
     });
 
     describe('allows overriding the default currencies', () => {
+      const currencyManager = new CurrencyManager(CurrencyManager.getDefaultList());
```
kevindavee, line 2035: I think this test was inaccurate. Previously, the `AddressBasedPaymentNetwork` initialize it's own currency manager; hence the request can be created. Suppose that we only have `_TEST` currency in the currency manager.

packages/advanced-logic/src/advanced-logic.ts
```diff
@@ -56,26 +56,30 @@ export default class AdvancedLogic implements AdvancedLogicTypes.IAdvancedLogic
...
   };
 
+  private currencyManager: ICurrencyManager;
+
   constructor(currencyManager?: ICurrencyManager) {
     if (!currencyManager) {
       currencyManager = CurrencyManager.getDefault();
     }
+
+    this.currencyManager = currencyManager;
```
benjlevesque, line 66: I would go even further and force the injection at that level. This will enable removing the dependency from advanced-logic to the `currency` package (more in https://github.com/RequestNetwork/requestNetwork/pull/1238)

```suggestion
  constructor(private currencyManager: ICurrencyManager) {
```

packages/advanced-logic/src/extensions/payment-network/near-native.ts
benjlevesque, line 1: strange. why is this empty?

packages/request-client.js/test/index.test.ts
benjlevesque: I don't understand what you mean? I don't think this was inacurate, why do you need to add ETH?

packages/advanced-logic/src/advanced-logic.ts
benjlevesque: If you prefer, leave it like this and I'll update in my PR

benjlevesque (approved):

packages/request-client.js/test/index.test.ts
kevindavee: We need ETH to validate the address.

alexandre-abrioux (commented):
I agree with @KolevDarko , I think it would be easier to save the default instance after the first call to `getDefault()`, directly in the `CurrencyManager` like so:

```ts
export class CurrencyManager {

  private static defaultInstance: CurrencyManager;

  /**
   * Returns a default instance of CurrencyManager based on default lists
   */
  static getDefault(): CurrencyManager {
    if (this.defaultInstance) return this.defaultInstance;
    this.defaultInstance = new CurrencyManager(
      CurrencyManager.getDefaultList(),
      CurrencyManager.getDefaultLegacyTokens(),
    );
    return this.defaultInstance;
  }

}
```

What do you think?

kevindavee:
That would work as well @alexandre-abrioux @KolevDarko . However, I still think we have inconsistency issues across payment networks (some are using injected currencies, some are not). And for another reason, to remove dependency with the currency package, as what Benji suggest.

packages/request-client.js/test/index.test.ts
benjlevesque: ah got it. It worked before because the default currency manager was instanciated?

KolevDarko:
> That would work as well @alexandre-abrioux @KolevDarko . However, I still think we have inconsistency issues across payment networks (some are using injected currencies, some are not). And for another reason, to remove dependency with the currency package, as what Benji suggest.

Ok, that's a good benefit, to decouple the packages. 

We could do both then, or at least add a comment in the getDefault method that it's heavy

packages/request-client.js/test/index.test.ts
kevindavee: Yep. So the instantiated one contains `_TEST` currency only, and we have another currency manager that uses default list.

packages/request-client.js/test/index.test.ts
```diff
@@ -2031,10 +2032,19 @@ describe('request-client.js', () => {
...
       const requestNetwork = new RequestNetwork({
         signatureProvider: TestData.fakeSignatureProvider,
         useMockStorage: true,
         currencies: [
+          {
+            network: 'mainnet',
+            type: RequestLogicTypes.CURRENCY.ETH,
+            decimals: ETH!.decimals,
+            symbol: ETH!.symbol,
+          },
```
benjlevesque, line 2047:
```suggestion
      const currencyManager = CurrencyManager.getDefault();
      const ETH = currencyManager.getNativeCurrency(RequestLogicTypes.CURRENCY.ETH, 'mainnet');

      const requestNetwork = new RequestNetwork({
        signatureProvider: TestData.fakeSignatureProvider,
        useMockStorage: true,
        currencies: [
          ETH,
```

kevindavee: Screenshot 2023-11-28 at 16 31 57


Can't do it like that 😢

alexandre-abrioux:
@KolevDarko @kevindavee I agree lets do both then

packages/advanced-logic/src/extensions/payment-network/near-native.ts
MantisClone: It's strange that this `near-native.ts` appears here. It lives in the `near/` directory.

kevindavee: Maybe we forgot to delete it 😬 It's there since a while.
https://github.com/RequestNetwork/requestNetwork/blob/master/packages/advanced-logic/src/extensions/payment-network/near-native.ts

packages/request-client.js/test/index.test.ts
MantisClone: Dang. It would be really nice if this worked. I wonder if it would be worth a future PR to make this usage possible 🤔

MantisClone (approved):

benjlevesque:
I've open https://github.com/RequestNetwork/requestNetwork/issues/1269 for another (middle/long term) approach to this issue

PR #1286, opened 2023-12-08
feat(request-node): use subgraph for confirmation

### LINKED ISSUE #1287: Request Node confirmation backed by TheGraph
## Problem
The Request Node uses an internal store to flag confirmed transaction. 
Transactions are added to that store once the blockchain transaction is successful. 

This means only the node that created the request transaction has knowledge about the confirmation state. This doesn't have a strong impact on the flow, but adds a bit of complexity for running the node (needs to have a peristant store for this data, to avoid losing it when restarting)

Moreover, and more importantly, this means the confirmation doesn't consider the indexing on TheGraph: it's possible to have confirmation on a transaction, but get a "pending" state if fetching the request right after. This creates flakiness in the integration tests, but could also have impact on real usage. 

## Solution 
Use the Subgraph data to consider a transaction confirmed. This will make the confirmation store un-necessary, and increase the stability/accuracy. The only drawback is that the confirmation time might be increased slightly.

## Related issues
https://github.com/RequestNetwork/storage-subgraph/issues/12

benjlevesque:
Resolves #1287
Requires https://github.com/RequestNetwork/storage-subgraph/pull/13

This changes removes the handling of failed transactions. I think this is acceptable, since there is a relatively short timeout on confirmation, and failures aren't very common. Please feel free to challenge this.

### DIFF
--- a/packages/request-node/src/request/confirmedTransactionStore.ts
+++ b/packages/request-node/src/request/confirmedTransactionStore.ts
@@ -2,9 +2,10 @@ import { DataAccessTypes, StorageTypes } from '@requestnetwork/types';
 import { SubgraphClient } from '@requestnetwork/thegraph-data-access';
 
 /**
- * Class for storing confirmed transactions information
- * When 'confirmed' event is received from a 'persistTransaction', the event data are stored.
- * The client can call the getConfirmed entry point, to get the confirmed event.
+ * Class for storing confirmed transaction information
+ * When 'confirmed' event is received from a 'persistTransaction', the event data is
+ * stored and indexed by the storage subgraph. The client can call the 
+ * getConfirmedTransaction endpoint, to get the confirmed event.
  */
 export default class ConfirmedTransactionStore {
   /**
--- a/packages/request-node/src/request/confirmedTransactionStore.ts
+++ b/packages/request-node/src/request/confirmedTransactionStore.ts
@@ -4,7 +4,7 @@ import { SubgraphClient } from '@requestnetwork/thegraph-data-access';
 /**
  * Class for storing confirmed transaction information
  * When 'confirmed' event is received from a 'persistTransaction', the event data is
- * stored and indexed by the storage subgraph. The client can call the 
+ * stored and indexed by the storage subgraph. The client can call the
  * getConfirmedTransaction endpoint, to get the confirmed event.
  */
 export default class ConfirmedTransactionStore {
@@ -19,9 +19,8 @@ export default class ConfirmedTransactionStore {
   public async getConfirmedTransaction(
     transactionHash: string,
   ): Promise<DataAccessTypes.IReturnPersistTransactionRaw | undefined> {
-    const { transactions, blockNumber } = await this.subgraphClient.getTransactionsByDataHash(
-      transactionHash,
-    );
+    const { transactions, blockNumber } =
+      await this.subgraphClient.getTransactionsByDataHash(transactionHash);
     if (transactions.length === 0) {
       return;
     }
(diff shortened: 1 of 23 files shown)


alexandre-abrioux (approved):
👌

olivier7delf (approved):

packages/request-node/src/request/confirmedTransactionStore.ts
```diff
@@ -1,50 +1,52 @@
...
-  public async addFailedTransaction(transactionHash: string, error: Error): Promise<void> {
-    await this.store.set(transactionHash, error);
+  ): Promise<DataAccessTypes.IReturnPersistTransactionRaw | undefined> {
+    const { transactions, blockNumber } = await this.subgraphClient.getTransactionsByDataHash(
+      transactionHash,
+    );
+    if (transactions.length === 0) {
+      return;
+    }
+    const transaction = transactions[0];
```
MantisClone, line 27: Under what condition, if any, will multiple transactions be returned from the subgraph?

packages/request-node/src/request/confirmedTransactionStore.ts
```diff
@@ -1,50 +1,52 @@
...
-import Keyv, { Store } from 'keyv';
+import { DataAccessTypes, StorageTypes } from '@requestnetwork/types';
+import { SubgraphClient } from '@requestnetwork/thegraph-data-access';
 
 /**
  * Class for storing confirmed transactions information
  * When 'confirmed' event is received from a 'persistTransaction', the event data are stored.
  * The client can call the getConfirmed entry point, to get the confirmed event.
  */
 export default class ConfirmedTransactionStore {
```
MantisClone, line 9: The header comment could use a refresh.

I considered requesting a class name change, but in the end decided it was okay. 

```suggestion
/**
 * Class for storing confirmed transaction information
 * When 'confirmed' event is received from a 'persistTransaction', the event data is
 * stored and indexed by the storage subgraph. The client can call the 
 * getConfirmedTransaction endpoint, to get the confirmed event.
 */
export default class ConfirmedTransactionStore {
```

packages/request-node/src/server.ts
```diff
@@ -3,22 +3,36 @@ import { Logger } from './logger';
...
-        filename: initializationStoragePath,
-      })
-    : undefined;
+  const network = getNetwork();
   const storage = getDataStorage(logger);
-  const dataAccess = getDataAccess(storage, logger);
-  return new RequestNode(dataAccess, storage, store, logger);
+  const dataAccess = getDataAccess(network, storage, logger);
+
+  // the confirmed transaction feature is specific to usage with a Request Node, so isn't exposed by dataAccess.
```
MantisClone, line 29: What does this comment mean? Is there a way to make it more clear?

MantisClone (approved):
> This change removes the handling of failed transactions. I think this is acceptable, since there is a relatively short timeout on confirmation, and failures aren't very common. Please feel free to challenge this.

I agree that this is acceptable for now.

The Graph Node throws errors when reorgs occur and I think our handling of that should be better, but fixing it is beyond the scope of this PR. 

Nice work :+1:

packages/request-node/src/request/confirmedTransactionStore.ts
benjlevesque: I don't think there's a way to have a single transaction from the subgraph. Anyway, it's possible to have twice the same transcation indexed (submit twice the same transaction to different IPFS blocks, submit twice the same IPFS block onchain). Since we only care about "confirmed or not", the other transactions than first don't really matter.

packages/request-node/src/request/confirmedTransactionStore.ts
benjlevesque: thanks 👍

packages/request-node/src/server.ts
benjlevesque: it means that this confirmation flow doesn't really matter outside of the usage with Client and Node, so it doesn't fit in the underlying packages / interfaces.

is this better? 
 ```suggestion
  // we access the subgraph client directly, not through the data access, because this feature is specific to RN use with Request Node. Without a node, the confirmation process would be different, so this doesn't fit in the data access layer
```

MantisClone: Yes. this is more clear. OK to commit :+1:

PR #1289, opened 2023-12-11
chore: allow for graphql middlewares

alexandre-abrioux:
## Description of the changes

This PR updates all our GraphQL packages to expose a new feature offered by the latest versions of `graphql-request` which are middlewares. They can be used to mutate requests sent to the Graph Node, or simply to log requests and responses for debugging purposes. See an example here: https://github.com/jasonkuhrt/graphql-request/blob/main/examples/other-middleware.ts

The new version of `graphql-request` has seen multiple breaking changes, and options accepted by the GraphQL client are now different. One noticeable change is the absence of the `timeout` option. Instead, a whole instance of `fetch` can be passed down.

Instead of handling GraphQL options at the Request Network protocol level, I've let the user code handle this. We now pass down all option to the GraphQL client and do not interfere with it.

For reference, here is how we can create a `fetch` instance to handle a `timeout`:
https://github.com/jasonkuhrt/graphql-request/issues/103#issuecomment-866646237


packages/payment-detection/package.json
```diff
@@ -47,20 +47,21 @@
...
     "tslib": "2.5.0"
   },
   "devDependencies": {
-    "@graphql-codegen/cli": "1.21.5",
-    "@graphql-codegen/typescript": "1.22.1",
-    "@graphql-codegen/typescript-document-nodes": "1.17.12",
-    "@graphql-codegen/typescript-graphql-request": "3.2.1",
-    "@graphql-codegen/typescript-operations": "1.18.0",
-    "@graphql-codegen/typescript-resolvers": "1.19.2",
+    "@babel/helper-get-function-arity": "^7.16.7",
```
alexandre-abrioux, line 58: Without this new package generating the graphql code result in the following error:
```
Error: Cannot find module '@babel/helper-get-function-arity'
```

I cannot find anything online regarding this, which surprises me, but adding this package solves the issue.

benjlevesque (approved):

leoslr (approved):

PR #1302, opened 2023-12-18
fix: sunset NEAR conversion for now

yomarion:
## Description of the changes

The previous oracle we implemented is down. We should investigate supporting Pyth network instead.

Removed:
- references to Flux oracles
- Near pairs from default conversion pairs (=> this is debatable, see More details)

Attempts to pay a Near request with conversion throws with `Payments with on-chain conversions are currently disabled on Near.`

### More details
Removing Near pairs is debatable, it's quite likely that we will add them back soon. If we think of this array as a stable default set on top of which implementers add their implementations, it's a poor choice and it's up to implementers to remove these values from the default values.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205942259962169


alexandre-abrioux (approved):

leoslr (approved):

olivier7delf (approved):

MantisClone (approved):
I'm okay with removing the fluxCurrencyPairs from the default list for now. 👍

PR #1304, opened 2023-12-19
chore(toolbox): sepolia and fix double transaction

benjlevesque:
- adding a default RPC for sepolia
- fixing double transactions, when contract versions have the same address (this is a hack so that requests can have PN version 2.0 or 2.1)
- add sepolia to default currency list

### DIFF
--- a/packages/toolbox/src/commands/chainlink/contractUtils.ts
+++ b/packages/toolbox/src/commands/chainlink/contractUtils.ts
@@ -16,11 +16,11 @@ export const runUpdate = async <T extends 'updateAggregator' | 'updateAggregator
 
   for (const { version, contract } of contractsWithVersion) {
     console.log(
-      `${dryRunText} will call ${method} on chainlinkConversionPath version ${version} at ${contract.address} (${args.network}))`,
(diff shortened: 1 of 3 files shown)


packages/toolbox/src/commands/chainlink/contractUtils.ts
```diff
@@ -16,11 +16,11 @@ export const runUpdate = async <T extends 'updateAggregator' | 'updateAggregator
 
   for (const { version, contract } of contractsWithVersion) {
     console.log(
-      `${dryRunText} will call ${method} on chainlinkConversionPath version ${version} at ${contract.address} (${args.network}))`,
+      `${dryRunText}will call ${method} on chainlinkConversionPath version ${version} at ${contract.address} (${args.network}))`,
     );
     console.log(JSON.stringify(params));
     if (args.dryRun) {
-      process.exit();
+      continue;
```
benjlevesque, line 23: dryRun was not showing the 2 transactions

KolevDarko (approved):

leoslr (approved):

PR #1318, opened 2024-01-09
feat(ethereum-storage): gas fee multiplier & max gas fee

alexandre-abrioux:
## Description of the changes

This PR introduces two new options to the `EthereumTransactionSubmitter`:
- `gasPriceMax`: this parameter can be used to protect against gas price spikes and adds an upper threshold
- `gasPriceMultiplier`: this parameter can be used to add a margin to the computed gas fee

### DIFF
--- a/packages/ethereum-storage/src/ethereum-tx-submitter.ts
+++ b/packages/ethereum-storage/src/ethereum-tx-submitter.ts
@@ -8,7 +8,7 @@ import { SimpleLogger, isEip1559Supported } from '@requestnetwork/utils';
 export type SubmitterProps = {
   signer: Signer;
   /**
-   * The minimum value for maxPriorityFeePerGas and maxFeePerGas
+   * The minimum value for maxPriorityFeePerGas and maxFeePerGas.
    * The default is zero.
    */
   gasPriceMin?: BigNumber;
--- a/packages/ethereum-storage/src/ethereum-tx-submitter.ts
+++ b/packages/ethereum-storage/src/ethereum-tx-submitter.ts
@@ -7,8 +7,20 @@ import { SimpleLogger, isEip1559Supported } from '@requestnetwork/utils';
 
 export type SubmitterProps = {
   signer: Signer;
+  /**
+   * The minimum value for maxPriorityFeePerGas and maxFeePerGas
+   * The default is zero.
+   */
   gasPriceMin?: BigNumber;
+  /**
+   * The maximum value for maxFeePerGas.
+   * There is no limit if no value is set.
+   */
   gasPriceMax?: BigNumber;
+  /**
+   * A multiplier for the computed maxFeePerGas.
+   * The default is 100, which does not change the value (100 is equal to x1, 200 is equal to x2).
+   */
   gasPriceMultiplier?: number;
   network: CurrencyTypes.EvmChainName;
   logger?: LogTypes.ILogger;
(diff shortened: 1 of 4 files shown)


packages/utils/test/normalize-gas-fees.test.ts
```diff
@@ -18,4 +18,30 @@ describe('Normalize Gas Fees', () => {
...
+    const normalizedGasFees = await normalizeGasFees({
+      logger: console,
+      suggestFees: async () => ({
+        baseFee: '20000000000', // 20 Gwei
+        maxPriorityFee: '2000000000', // 2 Gwei
+      }),
+      gasPriceMultiplier: 200, // x2
+    });
+    expect(normalizedGasFees.maxPriorityFeePerGas?.toString()).toBe('2000000000'); // 2 Gwei
+    expect(normalizedGasFees.maxFeePerGas?.toString()).toBe('44000000000'); // (20 + 2) x 2 = 44 Gwei
```
olivier7delf, line 45: I have a question, according to the formula:
Total Fee = Gas unit (limits)*(Base fee + Tip)

Could it be useful to have a parameter that multiplies only the Tip? Allowing to be better than other tips, but without a gasFees value too high.
In the test, gasFees => 20 + (2) x 2 = 24 Gwei

I might be lacking of context (google didn't help that much).

packages/ethereum-storage/src/ethereum-tx-submitter.ts
```diff
@@ -8,6 +8,8 @@ import { SimpleLogger, isEip1559Supported } from '@requestnetwork/utils';
 export type SubmitterProps = {
   signer: Signer;
   gasPriceMin?: BigNumber;
+  gasPriceMax?: BigNumber;
+  gasPriceMultiplier?: number;
```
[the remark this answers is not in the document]
alexandre-abrioux: Thanks, I've added comments on all fields in https://github.com/RequestNetwork/requestNetwork/pull/1318/commits/412b1c732d487faca50a0a1060d90c6cfde2f31f

packages/utils/test/normalize-gas-fees.test.ts
alexandre-abrioux: The formula for eip-1559 is `maxFeePerGas = baseFeePerGas + maxPriorityFeePerGas`, see https://docs.alchemy.com/docs/maxpriorityfeepergas-vs-maxfeepergas

The priority fee is the fee given to validators for including our transaction. The priority fee is set by us: we use the `urgent` value of the lib `rainbow-me/fee-suggestions`, which is set to a minimum of 2Gwei, [see here](https://github.com/rainbow-me/fee-suggestions/blob/main/src/index.ts#L167). This signals the network that we are willing to pay for our transaction to be included ASAP. We have control over this value, this is not an issue. 

However, we have no control over the base fee. The base fee is set by the network, and it is burned when the transaction is included. It is approximated by the lib `rainbow-me/fee-suggestions`, but between the time of approximation and the time of the transaction, the base fee can change a bit.

Because the base fee can change, the total will also change, which is the second value we control: we can set a cap for the total value with `maxFeePerGas`. However if `maxFeePerGas` is set too low and the base fee increases between approximation and inclusion, there might be a very low amount of gas remaining for validators to take their priority fee cut.

To fix this, I introduced a multiplier that acts on the total. We can virtually increase the total (hence the approximation) and this will give room for our transaction to be included. Note that this doesn't mean we will pay more gas in the end, because miners/validators cannot cheat on the amount of gas used (defined by the EVM) and the base fee (defined by the network on each block). They can only decide which cut of the priority fee they want to take (they usually take all of it).

PR #1320, opened 2024-01-10
feat: contract update with safe

leoslr:
## Description of the changes

Soon contract updates will be performed through Safe instead of EAO to increase security and go towards more decentralization.
This PR introduce the possibility to trigger an update via a Safe wallet.

When an update is performed, the associated transaction(s) will appear in the associated Safe(s) for co-owners to sign.

#### Comments

The latest version of `@safe-global` packages  required an update to `ethers v6`. This was a breaking change, so used the previous version.


packages/smart-contracts/hardhat.config.ts
```diff
@@ -365,13 +350,14 @@ task(
...
-  'Update the latest deployed contracts from the Create2DeploymentList',
-).setAction(async (_args, hre) => {
-  await hre.run(DEPLOYER_KEY_GUARD);
-  await updateContractsFromList(hre as HardhatRuntimeEnvironmentExtended);
-});
+task('update-contracts', 'Update the latest deployed contracts from the Create2DeploymentList')
+  .addFlag('safe', 'Is the update to be performed in Safe context')
+  .setAction(async (args, hre) => {
+    const safeMode = args.safe ?? false;
+    console.log(safeMode);
```
benjlevesque, line 357:
```suggestion
```

benjlevesque: I find the name `safeMode` very confusing. It's most often used to describe a type of execution, but here it's used to describe an actual mode that's using Safe, if I'm not mistaken

benjlevesque (approved):

packages/smart-contracts/hardhat.config.ts
MantisClone: Agreed

MantisClone: Maybe `isSafeMultisig` would make it more clear?

MantisClone: Also consider clarifying all of the header comments: 

```
safeMode Is the transaction to be executed by a Safe Multisig
```

MantisClone (approved):

packages/smart-contracts/README.md
```diff
@@ -269,6 +269,12 @@ yarn hardhat update-contracts
 
 This command will output details about each update on each chain
 
+If the contracts are administrate by a Safe, use the flag `safe`:
+
+```bash
+yarn hardhat update-contracts --safe
```
yomarion, line 275: This should be default actually. And we should use `--eoa` to force the opposite.

packages/smart-contracts/scripts-create2/contract-setup/execute-contract-method.ts
```diff
@@ -0,0 +1,82 @@
...
+    const safeSdk = await Safe.create({ ethAdapter, safeAddress });
+
+    const safeTransactionData = [
+      {
+        to: contract.address,
+        data: contract.interface.encodeFunctionData(method, props),
+        value: '0',
+      },
+    ];
+    console.log(safeAddress);
```
yomarion, line 60:
```suggestion
    console.log({safeAddress});
```

packages/smart-contracts/scripts-create2/contract-setup/execute-contract-method.ts
```diff
@@ -0,0 +1,82 @@
...
+    const nonce = await safeService.getNextNonce(safeAddress);
+    const safeTransaction = await safeSdk.createTransaction({
+      safeTransactionData,
+      options: { nonce },
+    });
+    const safeTxHash = await safeSdk.getTransactionHash(safeTransaction);
+    const senderSignature = await safeSdk.signTransactionHash(safeTxHash);
+
+    await safeService.proposeTransaction({
```
yomarion, line 69: If I'm not wrong, we can expect some error at this step, and we should maybe catch the error in order to append the `network` to its message.

packages/smart-contracts/src/lib/artifacts/index.ts
```diff
@@ -24,3 +24,7 @@ export * from './RequestDeployer';
  * ERC20 Transferable Receivable
  */
 export * from './ERC20TransferableReceivable';
+/**
+ * Admin
+ */
+export * from './Admin';
```
yomarion, line 30: Does it need to be exported?

yomarion (commented):
Nice one

packages/smart-contracts/src/lib/artifacts/index.ts
leoslr: Yes we import it in another file to retrieve the safe address.
(It's not fully up to date yet, I'll make a PR when all the safes are ready)

packages/smart-contracts/scripts-create2/contract-setup/execute-contract-method.ts
leoslr: We already have this in the main calling methods.
Example:
```
console.warn(`An error occurred during the setup of ChainlinkConversionPath on ${network}`);
console.warn(err);
```

packages/smart-contracts/src/lib/artifacts/index.ts
MantisClone: Thanks for clarifying. I had wondered if this list of Safe addresses was complete yet.

PR #1333, opened 2024-01-18
fix(security): vulnerabilities + remove coveralls

benjlevesque:
## Changes
This PR upgrades Lerna from 3 to 6, there is no breaking change and the new version is actually pretty nice. We'll need to confirm publish still works. NB: Lerna was the cause of _many_ vulnerabilities. 

It also drops `coveralls`  and the `publish-coverage` step from the CI. I don't think code coverage is worth having critical vulnerabilities, and there's no fix available (the library hasn't been maintained for 3 years)

## Resolutions
Before: 
```
$ yarn audit
yarn audit v1.22.19
[...]
58 vulnerabilities found - Packages audited: 1699
Severity: 19 Moderate | 32 High | 7 Critical
Done in 2.92s.
```

After:
```
$ yarn audit
yarn audit v1.22.19
0 vulnerabilities found - Packages audited: 1611
Done in 2.58s.
```
_NB: this is optimistic, the OpenZeppelin vulnerabilities will not be removed by this PR_

### Critical
https://github.com/RequestNetwork/requestNetwork/security/dependabot/207
https://github.com/RequestNetwork/requestNetwork/security/dependabot/177
https://github.com/RequestNetwork/requestNetwork/security/dependabot/137
https://github.com/RequestNetwork/requestNetwork/security/dependabot/123
https://github.com/RequestNetwork/requestNetwork/security/dependabot/51
https://github.com/RequestNetwork/requestNetwork/security/dependabot/14

### High 
https://github.com/RequestNetwork/requestNetwork/security/dependabot/182
https://github.com/RequestNetwork/requestNetwork/security/dependabot/172
https://github.com/RequestNetwork/requestNetwork/security/dependabot/165
https://github.com/RequestNetwork/requestNetwork/security/dependabot/124
https://github.com/RequestNetwork/requestNetwork/security/dependabot/120
https://github.com/RequestNetwork/requestNetwork/security/dependabot/113
https://github.com/RequestNetwork/requestNetwork/security/dependabot/101
https://github.com/RequestNetwork/requestNetwork/security/dependabot/80
https://github.com/RequestNetwork/requestNetwork/security/dependabot/75
https://github.com/RequestNetwork/requestNetwork/security/dependabot/45

PR #1237, opened 2023-11-13
refactor: remove axios

benjlevesque:
## Motivation
- reduce size (minor impact)
- avoid versions conflicts
- reduce maintenance

Changing to MSW for mocks also has the benefit to be compatible with any client.

## Choices
- Dropping `gas-price-definer`. This is never used, since #1117 
- dropping `IpfsStore.read`. This is never used. 
- remove `maxSize` from `IpfsManager.read`. This method is kept as a convenience, as it's useful to test the `read` after upload, but is no longer used (since #1117). 
- `axios` remains a `devDependency` of `smart-contracts` and `toolbox` for now

PR #1338, opened 2024-01-19
chore(cd): fix clean script for cd

yomarion:
## Description of the changes

This fixes the issue introduced by #1328 , when running `next-release` as part of the CI.

Partial view of the problem:
```
> @requestnetwork/data-format:clean

$ shx rm -rf dist tsconfig.tsbuildinfo tsconfig.build.tsbuildinfo
rm: no such file or directory: tsconfig.tsbuildinfo
```

### DIFF
--- a/packages/advanced-logic/package.json
+++ b/packages/advanced-logic/package.json
@@ -32,7 +32,7 @@
   ],
   "scripts": {
     "build": "tsc -b tsconfig.build.json",
-    "clean": "shx rm -rf dist tsconfig.tsbuildinfo tsconfig.build.tsbuildinfo",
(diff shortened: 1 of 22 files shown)


benjlevesque (approved):

alexandre-abrioux (approved):

benjlevesque:
I don't think the issue comes from #1328, but from #1237

benjlevesque:
still not 😅 
actually #1333

yomarion:
> still not 😅 actually #1333

@benjlevesque of course! The previous one failed to build so it's a different issue. I was looking for the lerna change from previous commits but it's not in the name. Thanks.

benjlevesque:
what's strange is that the file missing is in `data-format`, which is a good argument for https://github.com/RequestNetwork/requestNetwork/pull/1328, but the build and released worked

PR #1344, opened 2024-01-25
fix: lerna publish error

MantisClone:
## Description of the changes

* Remove `useWorkspaces: true` from `lerna.json`

# Error

```
lerna ERR! ECONFIGWORKSPACES The "useWorkspaces" option has been removed. By default lerna will resolve your packages using your package manager's workspaces configuration. Alternatively, you can manually provide a list of package globs to be used instead via the "packages" option in lerna.json.
```

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,5 @@
   "lerna": "3.2.1",
   "packages": ["packages/*"],
   "version": "independent",
-  "npmClient": "yarn",
-  "useWorkspaces": true
+  "npmClient": "yarn"
 }


benjlevesque (approved):
thanks, and my bad, it's related to the recent upgrade I made... Didn't spot any breaking change 🤔

MantisClone:
I think this is the breaking change: https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes-1

skiv71 (approved):

PR #1345, opened 2024-01-25
fix: improved currency detection

KolevDarko:
The reason for this PR came about when trying to introduce native USDC and bridged USDC to Polygon and Optimism.
Similar fix was done to [Arbitrum One here](https://github.com/RequestNetwork/currency-api/pull/366) by @alexandre-abrioux.
However, the reason it worked for Arbitrum, and does not work for Polygon and Optimism is because the network name arbitrum-one has a dash in it, the full currency identifier is USDC-arbitrum-one.
Because of the two dashes, the currency parsing from CurrencyManager failed with CurrencyManager.fromSymbol and then was correctly detected by CurrencyManager.fromId.
But with chain names without dashes, the CurrencyManager.fromSymbol worked but it fetches the wrong USDC token, it doesn't prioritize the id, that's why I had to prioritize the CurrencyManager.fromId with this PR.

I think it's the more correct way of doing things because the id is exact match, if the exact match fails, then we can attempt to detect based on the symbol with or without a network etc.

Further Context on the USDC tokens in [Asana Task](https://app.asana.com/0/1206167058664729/1206100829839124)


packages/currency/src/currencyManager.ts
```diff
@@ -71,17 +71,21 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
...
     }
 
+    if (network && currencyIdentifier.indexOf(network) === -1) {
+      currencyIdentifier = `${currencyIdentifier}-${network}`;
+    }
+
+    const currencyFromId = this.fromId(currencyIdentifier);
+
+    if (currencyFromId) return currencyFromId;
+
```
KolevDarko, line 81: Main purpose of this PR is to prioritize the method CurrencyManager.fromId before fromSymbol, because it matches the direct id of the currency, it's more specific.

packages/currency/src/currencyManager.ts
```diff
@@ -71,17 +71,21 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
       return this.fromAddress(currencyIdentifier, network);
     }
 
+    if (network && currencyIdentifier.indexOf(network) === -1) {
+      currencyIdentifier = `${currencyIdentifier}-${network}`;
+    }
```
KolevDarko, line 76: In making the improvement below, we need to handle the case where we provide the network as separate parameter but the network is not within the symbol. 
There is an issue when we have both Fiat and ERC20 token with same symbol, like MNT.
MNT - fiat currency
MNT-mainnet - Ethereum token.
The simplest fix here was to add the network to the currencyIdentifier if it's not already there.

alexandre-abrioux (approved):

packages/currency/src/erc20/index.ts
```diff
@@ -17,6 +17,7 @@ export function getSupportedERC20Tokens(): ERC20Currency[] {
           network: networkName,
           decimals: token.decimals,
           symbol: token.symbol,
+          id: token.id,
```
alexandre-abrioux, line 20: @KolevDarko The tokens do not have an `id` (see the [JSON file](https://github.com/RequestNetwork/requestNetwork/blob/master/packages/currency/src/erc20/chains/sepolia.ts)), you can create it manually maybe.
```suggestion
          id: `${token.symbol}-${networkName}`,
```

packages/currency/test/currencyManager.test.ts
```diff
@@ -709,4 +709,46 @@ describe('CurrencyManager', () => {
       expect(path).toMatchObject([eur.hash, dai.hash.toLowerCase()]);
     });
   });
+
+  describe('Native and bridged USDC', () => {
```
benjlevesque, line 713: The `describe` and `it` should give details on the behaviour you are testing, not about the data.

packages/currency/src/erc20/index.ts
```diff
@@ -17,6 +17,7 @@ export function getSupportedERC20Tokens(): ERC20Currency[] {
           network: networkName,
           decimals: token.decimals,
           symbol: token.symbol,
+          id: token.id || `${token.symbol}-${networkName}`,
```
benjlevesque, line 20: This is redundant, there should be only 1 place where we compute the ID
```suggestion
          id: token.id,
```

packages/currency/src/currencyManager.ts
```diff
@@ -71,17 +71,21 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
       return this.fromAddress(currencyIdentifier, network);
     }
 
+    if (network && currencyIdentifier.indexOf(network) === -1) {
+      currencyIdentifier = `${currencyIdentifier}-${network}`;
```
benjlevesque, line 75: use the existing (static) method `currencyId`. You can change the signature to this: 

```ts
 static currencyId(currency: { symbol: string; network?: string }): string {
```

packages/currency/src/currencyManager.ts
```diff
@@ -71,17 +71,21 @@ export class CurrencyManager<TMeta = unknown> implements ICurrencyManager<TMeta>
...
+    }
+
+    const currencyFromId = this.fromId(currencyIdentifier);
+
+    if (currencyFromId) return currencyFromId;
```
benjlevesque, line 80: I don't get this. If you are using this method by passing always IDs, why not use `fromId` directly?

packages/currency/test/currencyManager.test.ts
```diff
@@ -709,4 +709,46 @@ describe('CurrencyManager', () => {
...
+    const USDC_LIST = [
+      {
+        address: '0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174',
+        name: 'USD Coin (PoS)',
+        symbol: 'USDCe',
+        network: 'matic',
+        id: 'USDC-matic',
```
benjlevesque, line 720: can you give context why the ID wouldn't be `USDCe-matic` for this token?

benjlevesque (changes requested):

MantisClone:
@KolevDarko Although you ended up fixing the behavior another way, do you still think that this PR is still the "more correct way of doing things"?
CC @benjlevesque 

I want to understand if this PR still has merit or benefit for the rest of the Request Network ecosystem.

alexandre-abrioux:
I agree with @MantisClone, I think we should keep this PR open and continue to push for this change. Retrieving a currency from the currency list with the `CurrencyManager.from()` function should first prioritize using the currency ID and then try looking in the list from the currency symbol.

### Context
While maintaining an extended list of currencies at Request Finance for our day-to-day operations (https://api.request.finance/currency/) we came across the need to rename some currencies.

For instance, on Arbitrum, the community first used a bridged USDC token. Then in 2023, Circle deployed their own version of USDC on Aritrum, called the "native USDC token", see https://www.circle.com/blog/arbitrum-usdc-now-available.

At Request Finance, we renamed the old USDC token to USDCe, and we introduced the new native token as USDC.

Behind the scenes, the ID of the legacy token did not change to prevent breaking previously created Requests. So the symbol and ID of these currencies started to drift apart:
- `USDC-arbitrum (ID)`, the legacy token, now corresponds to `USDCe (symbol)`
- `USDCn-arbitrum (ID)`, the new token, corresponds to `USDC (symbol)`

Anywhere we retrieve a currency with the `CurrencyManager.from()` function, we should make sure to first look it up by ID, because it is always more precise.

packages/currency/src/erc20/index.ts
alexandre-abrioux: Sorry, this was my suggestion. @KolevDarko I guess its not needed because the ID is already added here, in `fromInput()`: https://github.com/RequestNetwork/requestNetwork/blob/master/packages/currency/src/currencyManager.ts#L311

packages/currency/src/currencyManager.ts
alexandre-abrioux: You're right, we should always use `fromId` in user code when we are sure to have an ID as input.

In the case of `from`, we're not entirely sure what kind of input we have (ID, symbol, address). In this case, we want to prioritize looking by ID rather than symbol because it's more precise (like we already prioritize looking by address first). More context here: https://github.com/RequestNetwork/requestNetwork/pull/1345#issuecomment-1921277612

I'm sure we are over-using `from` sometimes where we could replace it with `fromId` in user code, but the path with `from` should work too and needs to be fixed.

benjlevesque: should we deprecate `from`?

alexandre-abrioux: Yes, that could be a good start 👍 adding the `@deprecated` JSDOC

alexandre-abrioux (approved):
I'll let you answer @benjlevesque's suggestions, after that it's all good for me 👍

packages/currency/test/currencyManager.test.ts
KolevDarko: I added an elaborate comment above the describe block; there are many ways to explain this, I'm open to suggestions.

alexandre-abrioux: Looks good!

benjlevesque (approved):

leoslr (approved):

PR #1349, opened 2024-01-30
chore(security): remove ipfs-unixfs

benjlevesque:
Resolves https://github.com/advisories/GHSA-mxhp-79qh-mcx6 (not spotted by Github Security vulnerabilities, but visible in GCP docker registry scanner)

`ipfs-unixfs@6.0.7` > `protobufjs@6.11.4` > `taffydb@2.6.2`

We cannot upgrade ipfs-unixfs easily due to a breaking change to drop CJS support. 

`ipfs-unixfs` was only used to read IPFS content, which we don't need anymore, since the indexing is done by the Subgraph

### DIFF
--- a/packages/ethereum-storage/package.json
+++ b/packages/ethereum-storage/package.json
@@ -46,7 +46,6 @@
     "@requestnetwork/utils": "0.40.0",
     "ethers": "5.5.1",
     "form-data": "3.0.0",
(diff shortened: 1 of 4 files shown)


alexandre-abrioux (approved):

MantisClone:
Okay :+1: It makes sense to remove this code for reading from IPFS since it's no longer used.

PR #1350, opened 2024-01-31
fix: isSolvent

leoslr:
## Description of the changes

The method `isSolvent` was broken.
```ts
const needsGas =
    !(provider as any)?.provider?.safe?.safeAddress &&
    !['Safe Multisig WalletConnect', 'Gnosis Safe Multisig'].includes(
      (provider as any)?.provider?.wc?._peerMeta?.name,
    );
```
This is not working as intended anymore, and these info are no longer on the `provider` object.

I propose instead to pass a `needsGas` value - which is true by default. It should be to the client to decide whether he wants to check for gas fees or not, the protocol can't handle the situation for every wallet.

### DIFF
--- a/packages/payment-processor/src/payment/index.ts
+++ b/packages/payment-processor/src/payment/index.ts
@@ -195,14 +195,20 @@ export async function swapToPayRequest(
  * @param providerOptions.provider the Web3 provider. Defaults to getDefaultProvider.
  * @param providerOptions.nearWalletConnection the Near WalletConnection
  */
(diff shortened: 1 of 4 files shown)


yomarion (approved):

benjlevesque (approved):

MantisClone:
:+1: Makes sense.

MantisClone:
I just realized this PR changed the interface for `hasSufficientFunds` - it now takes an object containing the parameters instead of the individual parameters directly.

PR #1351, opened 2024-01-31
feat: remove lodash

benjlevesque:
This PR removes yet another superfluous dependency that causes security warnings and increases package size: lodash.


packages/types/src/payment-types.ts
```diff
@@ -218,6 +218,10 @@ export interface IERC20FeePaymentEventParameters extends IERC20PaymentEventParam
   feeAmountInCrypto?: string;
   amountInCrypto?: string;
   tokenAddress?: string;
+
+  gasUsed?: string;
+  gasPrice?: string;
+  maxRateTimespan?: string;
```
benjlevesque, line 224: the new typing makes it clear that these were missing

alexandre-abrioux (approved):
👍

leoslr (approved):

PR #1372, opened 2024-03-29
chore(types): sorted EVMChainName

yomarion:
## Description of the changes

Base is an `EVMChainName`.


alexandre-abrioux (approved):

leoslr (approved):

packages/types/src/currency-types.ts
```diff
@@ -6,6 +6,7 @@ export type EvmChainName =
   | 'arbitrum-one'
   | 'arbitrum-rinkeby'
   | 'avalanche'
+  | 'base'
```
MantisClone, line 9: 'base' is already on line 33. I recommend leaving this change and removing it from line 33 to maintain alphabetical order.

PR #1373, opened 2024-03-29
feat: Pay before Persist experiment

MantisClone:
# Problem

In theory, we can Pay a request before persisting it in the Request Network protocol.

In other words, we create a Request in memory without persisting it. Then, we prompt the end user to send payment against that request. Finally, we persist the request - so that it may serve as a receipt for the payment.

However, the existing SDK functions assume that the request is persisted immediately after being created.

# Motivation

This shortens the critical path for a user making a payment. This is especially desirable for payment apps - in which a payer wants to buy something. The workflow is driven by the _payer_ instead of the payee, so why make them wait for the request to be persisted before allowing them to send a payment?

# Changes

* feat: public funcs for creating a request in memory without persisting

# Why were these variables/functions chosen?

I traced the [RequestNetwork.createRequest()](https://github.com/RequestNetwork/requestNetwork/blob/create-without-publish/packages/request-client.js/src/api/request-network.ts#L78) and the [RequestLogic.acceptRequest()](https://github.com/RequestNetwork/requestNetwork/blob/create-without-publish/packages/request-logic/src/request-logic.ts#L136) functions and changed all the internal variables/functions `private` => `public`.

# Considerations

Rather than implementing this new logic, this PR only changes several `private` functions and variables to be `public`, thus empowering external contributors to experiment with the "Pay before Persist" workflow. The expectation is that once these experiments are complete, we either:

1. **Revert** this PR
or
3. **Implement** the "Pay before Persist" functions in the RN SDK.

### DIFF
--- a/packages/request-client.js/src/api/request-network.ts
+++ b/packages/request-client.js/src/api/request-network.ts
@@ -31,8 +31,8 @@ export default class RequestNetwork {
   public paymentNetworkFactory: PaymentNetworkFactory;
   public supportedIdentities: IdentityTypes.TYPE[] = supportedIdentities;
 
-  private requestLogic: RequestLogicTypes.IRequestLogic;
-  private transaction: TransactionTypes.ITransactionManager;
+  public requestLogic: RequestLogicTypes.IRequestLogic;
+  public transaction: TransactionTypes.ITransactionManager;
   private advancedLogic: AdvancedLogicTypes.IAdvancedLogic;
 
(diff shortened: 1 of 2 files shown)


MantisClone:
@limone-eth Here's the PR that will enable your Pay before Persist experiment.

limone-eth (approved):

sstefdev (approved):

PR #1376, opened 2024-03-29
fix: Pay before Persist experiment (Part 2)

MantisClone:
This is a fixup for #1373 

# Problem

#1373 didn't make enough variables `public` to enable the Pay before Persist experiment.

# Changes

* Make more variables public


MantisClone:
@limone-eth Here's the follow-up PR

sstefdev (approved):

PR #1403, opened 2024-05-23
chore(thegraph): fix xdai subgraph URL

alexandre-abrioux:
## Description of the changes

A bug was fixed in TheGraph Studio's interface, and we successfully transferred the Gnosis Chain subgraph. However, we're still unable to process the Ethereum one (similar issue).


leoslr (approved):

leoslr:
@alexandre-abrioux Maybe we could redeploy the same subgraph on RN studio account ? Or is there a blocker that prevent us from doing this ?

alexandre-abrioux:
@leoslr TheGraph will fix the issue, it's just a matter of time. It's better to wait, that way we keep the same URL for the decentralized subgraph. The URLs that we changed here are just the development URLs from TheGraph Studio

MantisClone (approved):

PR #1405, opened 2024-06-04
feat: meta-pn

leoslr:
### Context:

In the `Request` object, the `extensions` field is of type `Record<PN_ID, PN_STATE>`. 
As a result, the same request can't contain twice the same payment network. This is a blocker if we want a USD request to be payable in both USDC-matic and USDC-mainnet as we would need twice the `pn-any-to-erc20`.

### Meta payment network

This PR Introduces a new payment network: `META`.
Its purpose is to store information related to other payment networks.
This way we do not introduce breaking changes either at the request level, or at any of the existing payment network level.

### Scope

For now the sub-payment network supported are only:
 - `any-to-erc20-proxy`
 - `any-to-eth-proxy`
 - `any-to-declarative`

Supporting the `erc20-fee-proxy` pn would require to add a `network` field to it as it currently relies on the `request.currency` to determine the chain where the payment should happen.
I didn't want to introduce the associated changes in this PR. 
Furthermore, when there are several ways of paying a request, it's most likely denominated in FIAT. 

It can always be added later on if needed.

### State of the Meta pn:
```ts
{
  id: meta-pn,
  events: [...],
  values: {
    [key: string]: sub-pn-state // usual pn state
  } 
}

```

The key used to index each PN is the `salt` of the associated PN. This data already exists, is unique and can easily be used to identify a PN within the meta-pn extension.
Each sub-pn state contains the same data as it does when it is used as a standalone pn.

### Actions of the Meta pn

This new PN has two actions:
 - `create`
 - `apply-action-to-pn`

The create action perform basic validation regarding the meta-pn input data (e.g. uniqueness of each salt). Then it uses existing payment network to create individual sub-pn-state and index them using their associated salt. 

The `apply-action-to-pn` action applies an action to one of the sub-pn. This is required, because when we create a pn, we allow for missing information (e.g. payment address) that can be added at a later time.
If we don't have this action, it's not possible to add this information afterwards and the request will be unusable.
Same as the `create` action, it reuses all the capacities of existing payment network

### Payment detection

To get the balance for the `meta-pn` extension, we get the balance events of each sub-pn, then aggregate them. Reuse all of the existing logic.

However, reusing everything means that there's a different `paymentReference` for each sub-pn, as the salt of each sub-pn are different AND the `paymentAddress` as well.
Imo this is acceptable given the fact that the protocol already planned to have several `paymentReference` per request (it's not the same whether we proceed to a payment or a refund [see here](https://github.com/RequestNetwork/requestNetwork/blob/76ec94c6195e1ec0de1afbb8de45a62eab967630/packages/payment-detection/src/utils.ts#L147-L174))

### Payment processor

To build the transaction for the meta-pn extension, we pass the same properties as we currently do AND the key (salt) of the pn we wish to build the transaction for. It reuses all of the existing logic.


packages/advanced-logic/src/advanced-logic.ts
```diff
@@ -158,7 +162,9 @@ export default class AdvancedLogic implements AdvancedLogicTypes.IAdvancedLogic
 
   public getAnyToNativeTokenExtensionForNetwork(
     network?: CurrencyTypes.ChainName,
-  ): AnyToNative | undefined {
+  ):
+    | ExtensionTypes.IExtension<ExtensionTypes.PnAnyToAnyConversion.ICreationParameters>
+    | undefined {
```
leoslr, line 167: Unrelated but align with the method above.

packages/advanced-logic/src/extensions/abstract-extension.ts
```diff
@@ -60,7 +67,12 @@ export abstract class AbstractExtension<TCreationParameters> implements Extensio
         throw Error(`This extension has already been created`);
       }
 
-      copiedExtensionState[extensionAction.id] = this.applyCreation(extensionAction, timestamp);
+      copiedExtensionState[extensionAction.id] = this.applyCreation(extensionAction, timestamp, {
+        extensionsState,
+        extensionAction,
+        requestState,
+        actionSigner,
+      });
```
leoslr, line 75: Necessary so we can call `applyActionToExtension` from the meta-pn context to create sub-pn states ([see here](https://github.com/RequestNetwork/requestNetwork/pull/1405/files#diff-37ce446eeafd558b24a48b848941f960cfcf8703c119d6ffc3767318568d12e0R104)).

packages/advanced-logic/src/extensions/payment-network/meta.ts
```diff
@@ -0,0 +1,243 @@
+import { ICurrencyManager } from '@requestnetwork/currency';
```
leoslr, line 1: File of interest

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,190 @@
+import {
```
leoslr, line 1: File of interest

packages/payment-processor/test/payment/encoder-approval.test.ts
```diff
@@ -55,6 +55,7 @@ const approvalSettings = {
 const mnemonic = 'candy maple cake sugar pudding cream honey rich smooth crumble sweet treat';
```
leoslr, line 55: There's a bit of duplication in the payment-processor tests file, but I didn't want to refactor them in this PR

packages/types/src/advanced-logic-types.ts
```diff
@@ -36,8 +37,8 @@ export interface IAdvancedLogic {
     network: ChainName,
   ) => Extension.IExtension<Extension.PnReferenceBased.ICreationParameters> | undefined;
   getAnyToNativeTokenExtensionForNetwork: (
-    network: ChainName,
-  ) => Extension.IExtension<Extension.PnAnyToEth.ICreationParameters> | undefined;
+    network?: ChainName,
+  ) => Extension.IExtension<Extension.PnAnyToAnyConversion.ICreationParameters> | undefined;
```
leoslr, line 41: Unrelated. There was a type mismatch that wasn't raised by typescript before

packages/types/src/extension-types.ts
```diff
@@ -34,6 +36,7 @@ export interface IExtension<TCreationParameters = any> {
     actionSigner: Identity.IIdentity,
     timestamp: number,
   ) => RequestLogic.IExtensionStates;
+  createCreationAction: (parameters: TCreationParameters) => IAction<TCreationParameters>;
```
leoslr, line 39: This method is public on all PN, but it wasn't exposed in this interface.
Exposing it allows to call it [here](https://github.com/RequestNetwork/requestNetwork/pull/1405/files#diff-37ce446eeafd558b24a48b848941f960cfcf8703c119d6ffc3767318568d12e0R45)

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,190 @@
...
+          feeBalances.push(feeBalance);
+        }
+      }
+    }
+    const declaredEvents = this.getDeclarativeEvents(request);
+    const allPaymentEvents = [...declaredEvents, ...events.map((event) => event.events).flat()];
+
+    // FIXME: should be at the same level as balance
+    const values: any = this.getPaymentExtension(request).values;
+    values.feeBalance = {
```
yomarion, line 179: Have you wondered about the consequences of overriding the PN values instead of amending them? I think it makes but I don't have my head enough into it to be sure.

yomarion (commented):
(partial review, to be continued) Solid work and huge improvement to the lib 🚀

packages/payment-detection/src/meta-payment-detector.ts
leoslr: We do the exact same thing for the `any-to-erc20` payment detection. The `FIXME` comes from there.

packages/advanced-logic/src/extensions/payment-network/meta.ts
```diff
@@ -0,0 +1,244 @@
...
+
+      const subPnState: ExtensionTypes.IState =
+        request.extensions[ExtensionTypes.PAYMENT_NETWORK_ID.META]?.values?.[pnIdentifier];
+      if (!subPnState) {
+        throw new Error(`No payment network with identifier ${pnIdentifier}`);
+      }
+
+      const pn = this.getExtension(subPnState.id);
+      if (!pn) {
+        throw new Error(`Payment network ${subPnState.id} not supported`);
```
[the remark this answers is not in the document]
leoslr: Fixed.
It's now the same error everywhere, and directly thrown by the `getExtension` method.

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
```diff
@@ -0,0 +1,372 @@
...
+    it('can create a create action with all parameters', () => {
+      expect(
+        metaPn.createCreationAction({
+          [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ERC20_PROXY]: [baseParams, baseParams],
+        }),
+      ).toEqual({
+        action: 'create',
+        id: ExtensionTypes.PAYMENT_NETWORK_ID.META,
+        parameters: {
+          [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ERC20_PROXY]: [baseParams, baseParams],
```
alexandre-abrioux, line 34: Is it OK to have two with the same salt? Shouldn't it throw?

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,187 @@
...
+import { AnyToERC20PaymentDetector, AnyToEthFeeProxyPaymentDetector } from './any';
+import {
+  IPaymentNetworkModuleByType,
+  PaymentNetworkOptions,
+  ReferenceBasedDetectorOptions,
+} from './types';
+import { DeclarativePaymentDetector, DeclarativePaymentDetectorBase } from './declarative';
+import { BigNumber } from 'ethers';
+
+const supportedPns = [
```
alexandre-abrioux, line 18: We could enforce type on this to get a linting error when we allow new extensions.
```suggestion
const supportedPns: (keyof ExtensionTypes.PnMeta.ICreationParameters)[] = [
```

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,187 @@
...
+  ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ETH_PROXY,
+];
+
+const detectorMap: IPaymentNetworkModuleByType = {
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_DECLARATIVE]: DeclarativePaymentDetector,
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ERC20_PROXY]: AnyToERC20PaymentDetector,
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ETH_PROXY]: AnyToEthFeeProxyPaymentDetector,
+};
+
+const advancedLogicMap = {
```
alexandre-abrioux, line 30:
```suggestion
const advancedLogicMap: Record<keyof ExtensionTypes.PnMeta.ICreationParameters, keyof AdvancedLogicTypes.IAdvancedLogicExtensions> = {
```

packages/payment-processor/src/payment/utils.ts
```diff
@@ -420,3 +422,31 @@ export async function revokeErc20Approval(
   const tx = await signer.sendTransaction(preparedTx);
   return tx;
 }
+
+/**
+ * Format a request we wish to build a payment for.
+ * If the request does not use the meta-pn, returns it as it.
```
alexandre-abrioux, line 428:
```suggestion
 * If the request does not use the meta-pn, it returns it as is.
```

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
leoslr: Nice catch.
This validation was only in the `applyAction` logic. It's now in the `createCreationAction` logic as well. I added a new test to reflect that.

MantisClone (commented):
Thanks for the PR @leoslr! 

Here's a few comments and questions.

I've got to set it aside for today.I will continue my review tomorrow. I've only reviewed the `payment-processor`, `types`, and part of the `advanced-logic`. I haven't looked at `payment-detection` or any of the tests yet. And i'm still trying to wrap my head around the omission of the `erc20-fee-proxy` from the scope - especially whether the current design would prevent us from implementing it in the future.

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,192 @@
...
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ERC20_PROXY]: 'anyToErc20Proxy',
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ETH_PROXY]: 'anyToEthProxy',
+};
+
+/**
+ * Detect payment for the meta payment network.
+ * Recursively detects payments on each sub payment network
+ */
+export class MetaDetector extends DeclarativePaymentDetectorBase<
+  ExtensionTypes.PnMeta.IMeta<any>,
```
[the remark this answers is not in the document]
leoslr: Yes otherwise TS error

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.options = options || {};
+    this.currencyManager = currencyManager;
+    this.advancedLogic = advancedLogic;
+  }
+
+  /**
+   * Creates the extensions data for the creation of this extension.
+   *
+   * @param paymentNetworkCreationParameters Parameters to create the extension
+   * @returns The extensionData object
```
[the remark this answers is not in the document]
leoslr: At the request level there's a field called `extensionData` which contains each extenstion state (returned by this method).

packages/advanced-logic/src/extensions/payment-network/meta.ts
```diff
@@ -0,0 +1,248 @@
...
+  protected applyCreation(
+    extensionAction: ExtensionTypes.IAction<TCreationParameters>,
+    timestamp: number,
+    context: ICreationContext,
+  ): ExtensionTypes.IState;
+  protected applyCreation(
+    extensionAction: ExtensionTypes.IAction,
+    timestamp: number,
+    context?: ICreationContext,
+  ): ExtensionTypes.IState {
```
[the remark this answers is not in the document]
leoslr: I added the overload because it was throwing a TS error without it.
It does not anymore though, I removed them

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+    return this.extension.createCreationAction({
+      ...paymentNetworkCreationParameters,
+    });
+  }
+
+  /**
+   * Creates the extensions data to apply an action on a sub pn
+   *
+   * @param Parameters to add refund information
```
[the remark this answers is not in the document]
leoslr: Fixed

packages/payment-detection/src/meta-payment-detector.ts
```diff
@@ -0,0 +1,192 @@
...
+  private readonly currencyManager: CurrencyTypes.ICurrencyManager;
+  private readonly options: Partial<PaymentNetworkOptions>;
+  /**
+   * @param paymentNetworkId
+   * @param extension
+   */
+  public constructor({
+    advancedLogic,
+    currencyManager,
+    options,
```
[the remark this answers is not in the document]
leoslr: It is optional already, it's define in the types

packages/types/src/extensions/pn-meta.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/** Parameters of creation action */
+export interface ICreationParameters extends PnAnyDeclarative.ICreationParameters {
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_DECLARATIVE]?: PnAnyDeclarative.ICreationParameters[];
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ERC20_PROXY]?: PnAnyToErc20.ICreationParameters[];
+  [ExtensionTypes.PAYMENT_NETWORK_ID.ANY_TO_ETH_PROXY]?: PnAnyToEth.ICreationParameters[];
+}
+
+/** Parameters of declareSentPayment and declareSentRefund action */
```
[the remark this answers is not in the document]
leoslr: Fixed

packages/payment-processor/src/payment/utils.ts
```diff
@@ -420,3 +422,31 @@ export async function revokeErc20Approval(
...
+ * Format a request we wish to build a payment for.
+ * If the request does not use the meta-pn, it returns it as is.
+ * Otherwise, returns the request formatted with the pn of interest
+ */
+export function getFormattedRequest({
+  request,
+  options,
+}: {
+  request: ClientTypes.IRequestData;
+  options?: IRequestPaymentOptions;
```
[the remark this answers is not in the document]
leoslr: Done

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
```diff
@@ -0,0 +1,385 @@
...
+
+      it('cannot applyActionToExtensions of creation on a non supported currency', () => {
+        const requestCreatedNoExtension: RequestLogicTypes.IRequest = deepCopy(
+          TestData.requestCreatedNoExtension,
+        );
+        requestCreatedNoExtension.currency = {
+          type: RequestLogicTypes.CURRENCY.BTC,
+          value: 'BTC',
+        };
+        // 'must throw'
```
yomarion, line 159:
```suggestion
```

packages/advanced-logic/test/utils/payment-network/meta-pn-data-generator.ts
```diff
@@ -0,0 +1,448 @@
...
+  payerDelegate: undefined,
+  paymentInfo: undefined,
+  receivedPaymentAmount: '0',
+  receivedRefundAmount: '0',
+  refundInfo: undefined,
+  sentPaymentAmount: '0',
+  sentRefundAmount: '0',
+});
+// actions
+export const actionCreationFull = {
```
yomarion, line 46: Can you try to rename the fixtures in clear way so we can understand why adding a bitcoin payment is expected to throw, for instance? Example in this case `createAnyToErc20Mainnet`.

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
```diff
@@ -0,0 +1,385 @@
...
+            { ...MetaCreate.actionCreationFull, version: '' },
+            MetaCreate.requestStateNoExtensions,
+            TestData.otherIdRaw.identity,
+            TestData.arbitraryTimestamp,
+          );
+        }).toThrowError('version is required at creation');
+      });
+    });
+
+    describe('applyActionToExtension/applyApplyActionToExtension', () => {
```
yomarion, line 268:
```suggestion

    describe('applyApplyActionToExtension', () => {
```
Just a detail but if you nest it will show the test case path

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
```diff
@@ -0,0 +1,385 @@
...
+            MetaCreate.actionApplyActionToPn,
+            MetaCreate.requestStateCreatedMissingAddress,
+            TestData.payeeRaw.identity,
+            TestData.arbitraryTimestamp,
+          ),
+        ).toEqual(MetaCreate.extensionStateWithApplyAddPaymentAddressAfterCreation);
+      });
+
+      it('cannot applyActionToExtensions of addPaymentAddress without a previous state', () => {
+        // 'must throw'
```
yomarion, line 282:
```suggestion
      it('cannot applyActionToExtensions of addPaymentAddress without a previous state', () => {
```
You're missing a mirror test named "it can applyActionToExtensions of addPaymentAddress"

packages/payment-processor/src/payment/utils.ts
```diff
@@ -420,3 +421,31 @@ export async function revokeErc20Approval(
   const tx = await signer.sendTransaction(preparedTx);
   return tx;
 }
+
+/**
+ * Format a request we wish to build a payment for.
+ * If the request does not use the meta-pn, it returns it as is.
+ * Otherwise, returns the request formatted with the pn of interest
+ */
+export function getFormattedRequest({
```
yomarion, line 430:
```suggestion
export function flattenRequestByPnId({
```
Or something more explicit than "format" which looks like cosmetic convenience
```suggestion
export function unMetaRequestByPnId({
```

packages/payment-processor/src/payment/utils.ts
```diff
@@ -420,3 +421,31 @@ export async function revokeErc20Approval(
...
+}: {
+  request: ClientTypes.IRequestData;
+  pnIdentifier?: string;
+}): ClientTypes.IRequestData {
+  const pn = getPaymentNetworkExtension(request);
+  if (!pn?.id || pn.id !== ExtensionTypes.PAYMENT_NETWORK_ID.META) return request;
+  if (!pnIdentifier) throw new Error('Missing pn identifier');
+
+  const extensionOfInterest: ExtensionTypes.IState | undefined = pn.values[pnIdentifier];
+  if (!extensionOfInterest) throw new Error('Invalid pn identifier');
```
yomarion, line 442: `Invalid` is wrong, the ID might be valid but missing from the advanced logic state.

packages/payment-processor/test/payment/encoder-approval.test.ts
```diff
@@ -228,6 +229,61 @@ const validRequestEthConversionProxy: ClientTypes.IRequestData = {
   },
 };
 
+export const validMetaRequest: ClientTypes.IRequestData = {
```
yomarion, line 232: Can't you re-use the same fixture for `encoder-payment.test.ts`?

packages/payment-processor/test/payment/encoder-payment.test.ts
```diff
@@ -446,3 +502,90 @@ describe('Payment encoder handles ERC777 Stream', () => {
...
   });
 });
+
+describe('Payment encoder handles Meta PN', () => {
+  describe('Error cases', () => {
+    it('Should not be possible to encode a transaction when passing an invalid pn identifier', () => {
+      expect(() =>
+        encodeRequestPayment(validMetaRequest, provider, {
+          conversion: alphaConversionSettings,
+          pnIdentifier: 'unknown',
```
yomarion, line 512: This test makes it look like it throws because "unknown" is an invalid pnId, but it would also through with "eth-input" for instance, which is valid but missing.

packages/payment-processor/src/payment/utils.ts
yomarion:
```suggestion
export function decodeRequestMetaPn({
```

yomarion (approved):

packages/advanced-logic/test/extensions/payment-network/meta.test.ts
leoslr: I know, I followed the format of the other tests from this package

packages/payment-processor/src/payment/utils.ts
leoslr: I disagree, if the ID is valid then it is in the advanced logic state.

Given a request, in state X. we should consider that the request is correct and does not miss data. 
And the validity on `pnIdentifier` is based on X. 

Otherwise, I can pass any `pnIdentifier` with any request, and it will never be considered invalid.

packages/payment-processor/test/payment/encoder-approval.test.ts
leoslr: The data from both test are not the same. 
All the mocks should be shared in a common file, but this is already a big PR and I didn't want to include this mock refactoring as well. 
Sharing the data between both file would require to update the associated tests as the expected recipient / currencies as well as

packages/payment-processor/test/payment/encoder-payment.test.ts
leoslr: `eth-input` is not valid the `pnIdentifier` within the `meta-pn` is a unique identifier to identify each sub-pn, it's not the payment network identifiers like `any-to-erc20` `eth-input` ...

packages/payment-detection/src/index.ts
MantisClone, line 1: The `payment-detection` package provides a `getPaymentReference()` utility function.
We need a new utility function to return all the sub-paymentReferences.

MantisClone:
In the `Payment Detection` section of the description, can you clarify which paymentReference is written on chain when a "meta" request is paid?

leoslr:
@MantisClone The `paymentReference` used during a payment is the one associated to the sub-pn that is used to make the payment - I've updated the comment.
The payment subgraphs don't need to be updated.

Issue #1390, opened 2024-04-16
As a Builder I want to create a request in React Native


MantisClone:

MantisClone:
@aimensahnoun was able to successfully create a request from an expo app: https://github.com/RequestNetwork/rn-expo-support


![Image](https://github.com/RequestNetwork/requestNetwork/assets/2530913/157d0350-d2d8-41da-9607-e3c65ac2a05d)

MantisClone:
Related Follow-up Issues:
- https://github.com/RequestNetwork/expo-starter-request-network/issues/1
- https://github.com/RequestNetwork/docs.request.network/issues/58

PR #1411, opened 2024-06-14
chore: update to near-api-js 4.0.2

MantisClone:
Towards #1390 

## Description of the changes

* chore: update to near-api-js 4.0.2

## Motivation

For https://github.com/near/near-api-js/pull/1194 - to avoid the need to polyfill `http` and `https` in React Native environments. I'll be able to test it in https://github.com/RequestNetwork/expo-starter-request-network as soon as the `-next` version is released to NPM.

## Test

Using `npm link` I verified that this does indeed eliminate the need to polyfill `http` and `https` :+1:

### DIFF
--- a/packages/payment-processor/src/payment/utils-near.ts
+++ b/packages/payment-processor/src/payment/utils-near.ts
@@ -44,6 +44,7 @@ export const isNearAccountSolvent = (
     const fungibleContract = new Contract(nearWalletConnection.account(), token.value, {
       changeMethods: [],
       viewMethods: ['ft_balance_of'],
+      useLocalViewExecution: true,
     }) as any;
     return fungibleContract
       .ft_balance_of({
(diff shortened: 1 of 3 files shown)


packages/payment-processor/src/payment/utils-near.ts
```diff
@@ -44,6 +44,7 @@ export const isNearAccountSolvent = (
     const fungibleContract = new Contract(nearWalletConnection.account(), token.value, {
       changeMethods: [],
       viewMethods: ['ft_balance_of'],
+      useLocalViewExecution: true,
```
MantisClone, line 47: This parameter was added in https://github.com/near/near-api-js/pull/1172. I don't fully understand it, but I don't think it will break anything. It might make view functions to the same blockId more efficient.

alexandre-abrioux (approved):

sstefdev (approved):

aimensahnoun (approved):

PR #1428, opened 2024-07-15
feat(processor): any-declarative payment in NEAR

yomarion:
## Description of the changes
`payment-processor` util to make a payment in NEAR (native) token on a declarative request, through low-level parameters: `paymentAddress`, `paymentReference` and `network`.

## Summary by CodeRabbit

- **New Features**
  - Introduced new payment functions for NEAR tokens: `payNearAnyDeclarativeRequest`, `payNearAmountWithReference`.
  - Added `validatePaymentReference` function for enhanced payment reference validation.

- **Improvements**
  - Updated existing NEAR payment functions to use `validatePaymentReference` for better validation.
  - Enhanced `getAmountToPay` function to prioritize user-specified amounts, improving payment flexibility.

- **Bug Fixes**
  - Updated error message in `payNearConversionRequest` tests for better clarity.


packages/payment-processor/src/payment/near-any-declarative.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export async function payNearAnyDeclarativeRequest(
+  request: ClientTypes.IRequestData,
+  walletConnection: WalletConnection,
+  amount: BigNumberish,
+  callback?: INearTransactionCallback,
+): Promise<void> {
+  validateRequest(request, ExtensionTypes.PAYMENT_NETWORK_ID.ANY_DECLARATIVE);
+
+  const { paymentReference, paymentAddress, network } = getRequestPaymentValues(request);
```
leoslr, line 25: I don't think this method will return this data.

th declarative-pn has a `paymentInfo` field in the `values` extension field.
And in case of off-chain conversion, it would contain:
```ts
  "paymentInfo": {
      "paymentAddress": "xxx.near",
      "paymentCurrency": "NEAR-near"
    },
```

The network property defined [here](https://github.com/RequestNetwork/requestNetwork/blob/ac361e8fd4e44f3c2638e315b84e7a80073f8e41/packages/payment-processor/src/payment/utils.ts#L92) won't return what you expect.
Similarly, you won't have a `paymentAddress`.

I think it would be best to export a method with low level parameter (`amount`, `address`, `paymentRef`, `version`) instead of the request object, otherwise we kind of add RF-specific logic within the protocol as the shape of the `paymentInfos` field is defined in RF but not in RN

yomarion: @leoslr , I did as you advised, thanks.

leoslr (approved):

packages/payment-processor/src/payment/near-conversion.ts
```diff
@@ -12,12 +12,13 @@ import {
...
 import { CurrencyManager, NearChains, UnsupportedCurrencyError } from '@requestnetwork/currency';
+import { validatePaymentReference } from '../utils/validation';
 
 /**
  * Processes the transaction to pay a request in NEAR with on-chain conversion.
  * @param request the request to pay
- * @param walletConnection the Web3 provider, or signer. Defaults to window.ethereum.
- * @param amount optionally, the amount to pay. Defaults to remaining amount of the request.
+ * @param walletConnection the Near provider.
+ * @param amount optionally, the amount to pay in request currency. Defaults to remaining amount of the request.
```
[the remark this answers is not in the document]
MantisClone: This seems to be a correct observation - the typing in the test could use `NearProvider` instead of `any` :+1:

packages/payment-processor/src/payment/near-conversion.ts
```diff
@@ -37,13 +38,7 @@ export async function payNearConversionRequest(
     throw new UnsupportedCurrencyError(request.currencyInfo);
   }
 
-  if (!paymentReference) {
-    throw new Error('Cannot pay without a paymentReference');
-  }
-
-  if (!network || !NearChains.isChainSupported(network)) {
-    throw new Error('Should be a Near network');
-  }
```
MantisClone, line 46: Is the removal of this logic intentional?

MantisClone: Nevermind. Removed because redundant with `NearChains.assertChainSupported(network)`

MantisClone (approved):

packages/payment-processor/src/payment/near-conversion.ts
yomarion: Yes, should have made this more explicit in the PR indeed.

packages/payment-processor/src/payment/near-conversion.ts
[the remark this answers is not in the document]
yomarion: I don't understand how a more precise type for a mocked value in the tests positively impacts the library. Can you explain? I'll merge this one because of emergencies and the feedback is not related to the core of my change, but happy to later improve.

Issue #1381, opened 2024-04-02
Fix bugs in RN SDK release process


MantisClone:
# Problems

- [x] ~~While releasing 0.46.0, the release branch wasn't merged to master. This release branch is supposed to update the dependencies between packages and update the change log.~~  :call_me_hand: [**Not a Problem**
](https://github.com/RequestNetwork/requestNetwork/issues/1381#issuecomment-2239725330)
- [x] ~~the CHANGELOG.md files for each package did not update correctly. Weird duplicates of the 0.41.0 CHANGELOG~~ - :no_good: Won't Fix (at least not now. Maybe later)
- [x] The github release description was missing the CHANGELOG section, even thought it was correctly generated by the release script.
- [x] Release gets the wrong title 

![Image](https://github.com/user-attachments/assets/10b0d6b8-ea65-49a3-89bf-6d44615590e0)

# Proposed Solution

Fix errors in the private https://github.com/RequestNetwork/release-tools repo

MantisClone:
> While releasing 0.46.0, the release branch wasn't merged to master. This release branch is supposed to update the dependencies between packages and update the change log.

The `release` branch created as part of the `release.sh` script is supposed to be merged *after* publishing to NPM. Therefore, the first problem listed in the issue description is **not a problem**

MantisClone:
> the CHANGELOG did not update correctly. Weird duplicates of the 0.41.0 CHANGELOG

Likely caused by: [Publish does not detect manually created tags in fixed mode with Github/Github Enterprise](https://lerna.js.org/docs/troubleshooting#publish-does-not-detect-manually-created-tags-in-fixed-mode-with-githubgithub-enterprise)

MantisClone:
I wrote a script in https://github.com/RequestNetwork/release-tools/pull/5 that replaces the un-annotated tags with annotated ones.
I replaced the un-annotated tags with annotated ones.

```
mantisclone@mantisclone-Inspiron-7620:~/projects/requestNetwork$ git push --tags --force
Enumerating objects: 22, done.
Counting objects: 100% (22/22), done.
Delta compression using up to 20 threads
Compressing objects: 100% (22/22), done.
Writing objects: 100% (22/22), 1.31 KiB | 1.31 MiB/s, done.
Total 22 (delta 21), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (21/21), done.
To github.com:RequestNetwork/requestNetwork.git
[...]

PR #1440, opened 2024-07-25
fix: include merged tags when determining the previous release

MantisClone:
# Problem

The `next-release` workflow in CircleCI releases a `-next` package in NPM. It determines the version number based on the previous annotated, tagged commit *originating from the `master` branch*. It doesn't consider tags that were merged from other branches.

In https://github.com/RequestNetwork/requestNetwork/issues/1381#issuecomment-2241347444, I force-pushed several annotated tags in an attempt to fix the auto-generated CHANGELOG.md files. Before this, *there were zero annotated, tagged commits originating from the `master` branch* so the `next-release` workflow would fallback to counting commits from the beginning fo time. This is why the `-next` releases end with a number instead of a commit hash.

The reason there was zero annotated, tagged commits originating from the `master` branch is that we've always used squash merge to merge PRs into `master`. squash merge changes the commit hash, thus orphaning the tags produced by `lerna publish`.

# Proposed Solution

1. We could use `--include-merged-commits` option in the `next-release` workflow. We could adopt the release process in https://github.com/RequestNetwork/release-tools/pull/8 which uses `lerna publish` on a `release` branch and then `merges` (not `squash merge`) the tagged commits to master. The `-next` releases going forward would have `shortSHA` postfixes instead of `refCount` postfixes. This requires us to enable `merge` on the `master` branch which I don't love - because it runs the risk that non-conventional commits are merged to the `master` branch.

# Alternative Solution

2. We could delete the annotated tags on the `master` branch. We could still adopt the new release process in https://github.com/RequestNetwork/release-tools/pull/8 except continue using squash merge for the `release` branch. The `-next` releases going forward would still have the `refCount` postfixes, and we wouldn't have to enable `merge` on the `master` branch. I don't love this option either because it eliminates the possibility of bumping the version on only changed packages instead of force-publishing all packages every time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -510,6 +510,7 @@ jobs:
               --preid next \
               --no-verify-access \
               --dist-tag next \
+              --include-merged-tags \
               --yes \
               --force-publish=${FORCE_PUBLISH}
 


MantisClone:
We must go with option 2, cause I already messed it up and merged instead of squashing a normal PR. :facepalm:
image

MantisClone:
I'm closing this because this would change the `-next` releases to have a `shortSHA` postfix instead of the `refCount`

MantisClone:
The automatic next-release workflow [just ran](https://app.circleci.com/pipelines/github/RequestNetwork/requestNetwork/9765/workflows/7e993ff6-a126-41b7-b5df-c6b56826838a/jobs/151506/parallel-runs/0/steps/0-105) and released a correctly versioned -next release :white_check_mark:
The fix was to delete the old tags I tampered with. :face_with_peeking_eye:
This fixed it because these were the only annotated git tags that originated from the master branch rather than a merged branch.

PR #1540, opened 2025-01-06
chore(request-client): expose NoPersistHttpDataAccess

kevindavee:
## Description of the changes
There is no way to use `NoPersistHttpDataAccess` without using the `RequestNetwork` class. If we want to inject `NoPersistHttpDataAccess`, we can't use it with `RequestNetworkBase`.

Example: I want to use my own custom data access or `NoPersistHttpDataAccess` depending on my use case.
- If I use `RequestNetwork`, there's no way to inject my own custom data access.
- If I use `RequestNetworkBase`, there's no option to use `skipPersistence` which uses `NoPersistHttpDataAccess`.

By exporting `NoPersistHttpDataAccess`, I can manage which Data Access I want to use (Custom/NoPersist) and inject it by myself to `RequestNetworkBase`.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new data access method for handling HTTP requests without persistent storage.

### DIFF
--- a/packages/request-client.js/src/index.ts
+++ b/packages/request-client.js/src/index.ts
@@ -7,6 +7,7 @@ import { default as RequestNetworkBase } from './api/request-network';
 import { default as HttpMetaMaskDataAccess } from './http-metamask-data-access';
 import { default as HttpDataAccess } from './http-data-access';
 import { NodeConnectionConfig } from './http-data-access';
+import { NoPersistHttpDataAccess } from './no-persist-http-data-access';
 import * as Types from './types';
 
 export {
@@ -19,4 +20,5 @@ export {
   NodeConnectionConfig,
   Types,
   Utils,
+  NoPersistHttpDataAccess,
 };


alexandre-abrioux (approved):

aimensahnoun (approved):

leoslr (approved):

MantisClone:
I'll take a look at this. Please don't merge until I've wrapped my head around it.

kevindavee:
Well noted @MantisClone . But it looks like I'm going to submit another PR. Apparently, using `NoPersistHttpDataAccess` isn't possible for our use case

MantisClone (approved):
This change could be useful to other app builders who use `RequestNetworkBase` directly, and I'm in favor of merging it. Moreover, I think it matches the existing convention - The other `DataAccess` implementations are exported so `NoPersistHttpDataAccess` should be exported too.

MantisClone:
> Well noted @MantisClone . But it looks like I'm going to submit another PR. Apparently, using `NoPersistHttpDataAccess` isn't possible for our use case

@kevindavee can you share why `NoPersistHttpDataAccess` doesn't fit the Request Finance use case? I'm eager to understand your needs.

kevindavee:
At Request Finance, we're using our own custom `DataAccess`. Whereas `NoPersistHttpDataAccess` inherits from `HttpDataAccess` and uses Request Node. What we need is to have the same logic as `NoPersistHttpDataAccess` but use our custom `DataAccess` @MantisClone

MantisClone:
Sounds good. Can we still merge this PR?

yomarion (approved):

kevindavee:
Replaced by #1541

PR #1544, opened 2025-01-14
fix: epk encryption variable name typo

rodrigopavezi:
## Problem

EPK encryption through the EthereumPrivateKeyCipherProvider is broken. This issue was reported by @benjaminshafii where he was trying to create encrypted requests with the EthereumPrivateKeyCipherProvider and it was falling with the following error message:

```
❌ Failed to create encrypted request: Encryption key is required
Error details: 62 |             if (!data) {
63 |                 throw new Error('Data to encrypt cannot be empty');
64 |             }
65 |             console.log('options', options);
66 |             if (!((_a = options === null || options === void 0 ? void 0 : options.encryptionParams) === null || _a === void 0 ? void 0 : _a.key)) {
67 |                 throw new Error('Encryption key is required');
```

## Description of the changes
 - Fix the issue by passing a single encryptionParams item for the cipherProvider.encrypt method as expected.

## Summary by CodeRabbit

- **New Features**
  - Enhanced encryption process for creating transactions with a single encryption parameter.
  
- **Bug Fixes**
  - Improved test coverage for the transaction creation method using specific cipher providers.
  
- **Chores**
  - Updated the return type of the encryption method to provide more structured information about encryption results.

Note: These changes are internal and should not directly impact end-user functionality.


MantisClone (commented):
@rodrigopavezi Can you add more detail why the current code is broken and the impact of this bug?

rodrigopavezi:
> Are we missing a test for this?

@alexandre-abrioux good call. Test added.

packages/transaction-manager/test/unit/transactions-factory.test.ts
```diff
@@ -70,6 +70,57 @@ describe('transaction-factory', () => {
...
+        // eslint-disable-next-line no-magic-numbers
+        // 'encryptedData not right'
+        expect(encryptedTx.encryptedData.length).toBe(126);
+        // 'encryptedData not right'
+        expect(encryptedTx.encryptedData.slice(0, 2)).toEqual(
+          MultiFormatTypes.prefix.AES256_GCM_ENCRYPTED,
+        );
+      } else {
+        fail('encryptedData should not be undefined');
+      }
```
MantisClone, line 95: You can use expect(encryptedTx.encryptedData).toBeDefined() or something like that instead of an if/else.

packages/transaction-manager/test/unit/utils/test-data.ts
```diff
@@ -140,11 +140,15 @@ export class FakeEpkCipherProvider implements CipherProviderTypes.ICipherProvide
...
+  ): Promise<EncryptionTypes.IEncryptedData> {
     const encryptionParams = options.encryptionParams;
 
     if (encryptionParams.method === EncryptionTypes.METHOD.ECIES) {
-      return ecEncrypt(encryptionParams.key, data);
+      const encryptedValue = await ecEncrypt(encryptionParams.key, data);
+      return {
+        type: EncryptionTypes.METHOD.ECIES,
+        value: encryptedValue,
+      };
```
MantisClone, line 151: This change synchronizes the mock code with the actual encrypt function. :+1: 
https://github.com/RequestNetwork/requestNetwork/blob/959aa97a4965fb8e4b6663f20f07960cbd4afb78/packages/utils/src/encryption.ts#L46-L56

MantisClone (approved):
Approved :+1:  pending comment resolution :construction:

packages/transaction-manager/test/unit/utils/test-data.ts
rodrigopavezi: Sorry, is there anything I should change here?

MantisClone: No change is necessary! Sorry, I should have made that clear in my original comment. This part confused me so I wanted to comment as a form of documentation.

PR #1569, opened 2025-02-04
feat: add domain and statement support to LitProtocolCipherProvider

rodrigopavezi:
## Problem
When lit session expired Lit SDK will try to refresh it by calling a callback we implemented in Request Network SDK. This callback was using the params passed to the getSessionSignatures which makes them only available in the function scope. That caused the undefined issues when the callback was called afterwards. 

## Changes

Domain and Statement params are now an attribute of the LitProtocolCipherProvider so making it available on the class global scope. That should fix the undefined issue when the callback function is called for refreshing the session.

### DIFF
--- a/packages/lit-protocol-cipher/src/lit-protocol-cipher-provider.ts
+++ b/packages/lit-protocol-cipher/src/lit-protocol-cipher-provider.ts
@@ -51,6 +51,16 @@ export default class LitProtocolCipherProvider implements CipherProviderTypes.IC
   private decryptionEnabled = false;
 
   /**
+   * @property {string|undefined} domain - The domain to use for generating the auth sig.
+   */
+  private domain?: string;
+
+  /**
+   * @property {string|undefined} statement - The statement to use for generating the auth sig.
+   */
+  private statement?: string;
+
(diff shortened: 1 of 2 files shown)


rodrigopavezi:
@coderabbitai review

packages/lit-protocol-cipher/src/lit-protocol-cipher-provider.ts
```diff
@@ -147,8 +154,9 @@ export default class LitProtocolCipherProvider implements CipherProviderTypes.IC
 
       // Create the SIWE message
       const toSign = await createSiweMessage({
-        domain,
-        statement,
+        // Use class properties instead of function parameters
+        domain: this.domain,
+        statement: this.statement,
```
MantisClone, line 159: Since we use `this` inside the callback, should we `bind()` the callback to ensure it always uses the current `this` and not some other `this`?

```ts
const authNeededCallback = async (params: AuthCallbackParams) => {
...
  domain: this.domain,
  statement: this.statement,
...
}.bind(this); // <-- should we add this line?
```

PR #1573, opened 2025-02-18
chore(payment-processor): load hinkal lib asyncronously

alexandre-abrioux:
## Context

The Request Network library is relatively heavy to load from a front-end point of view, more especially the `payment-processor` package. Here are some stats for a front-end that includes packages needed for payment, and built with Vite.js:
```sh
> ls -lh build/assets/@requestnetwork/
total 7,3M
308K currency-BI28kHzu.js
128K data-format-CF2Ej1ak.js
1,2M payment-detection-B7AW9tt1.js
5,7M payment-processor-COb8hmRF.js
1    smart-contracts-l0sNRNKZ.js
1    types-l0sNRNKZ.js
```

## Changes
In this PR, I introduce a quick optimization that would reduce the `payment-processor` load time and resolve a CSP warning for non-Hinkal users.

- `@hinkal/common` represents a large amount of the final code for the `payment-processor` package after transpilation. It could be loaded asynchronously to lighten frontend bundles. The picture below is a visualization of the space taken by each sub-dependency of the `payment-processor` package, according to https://github.com/btd/rollup-plugin-visualizer.

- `@hinkal/common` depends on [`libsodium`](https://www.npmjs.com/package/libsodium) (also quite heavy, see below) which runs WASM code in the browser. Loading this lib is incompatible with a `Content-Security-Policy` that does not allow `script-src 'wasm-unsafe-eval'`. For builders that do not use the Hinkal integration, it does not make sense to relax the CSP header to silence such a warning. Loading Hinkal only on-demand would not trigger such warnings for non-Hinkal users. Related: https://github.com/jedisct1/libsodium.js/issues/196

image

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the processing of token payments by introducing improved asynchronous handling. This update ensures that all transaction details are fully resolved before proceeding, which leads to more reliable and consistent payment experiences for users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,5 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  },
-  "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -69,5 +69,6 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  }
+  },
+  "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
 }
--- a/packages/payment-processor/src/payment/erc-20-private-payment-hinkal.ts
+++ b/packages/payment-processor/src/payment/erc-20-private-payment-hinkal.ts
@@ -19,7 +19,7 @@ import {
 } from './utils';
 import { IPreparedPrivateTransaction } from './prepared-transaction';
 
-import { emporiumOp, IHinkal, RelayerTransaction } from '@hinkal/common';
+import type { IHinkal, RelayerTransaction } from '@hinkal/common';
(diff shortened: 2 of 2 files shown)


leoslr (approved):

MantisClone (approved):
Looks good to me :ok_person: 

As a result of these changes, the hinkal functions for preparing transactions are now asynchronous. I consider this to be a _non-breaking_ change because I assume most users of this code would use the `pay...` functions rather than the `prepare...` functions.

Thank you also for cleaning up the import ordering to make it alphabetical.

MantisClone:
I'm sad to see that one of Hinkal's sub-dependencies is incompatible with a `Content-Security-Policy` that does not allow script-src `'wasm-unsafe-eval'`. I forwarded this analysis to the Hinkal team.

alexandre-abrioux:
I haven't tested the Hinkal integration, but it might be fine even when without allowing `wasm-unsafe-eval` in your CSP. My guess is that you would get a warning in the browser console, but the underlying crypto library (`libsodium`) would switch to a Javascript fallback. Again, that is just a guess by reading through this: https://github.com/jedisct1/libsodium.js/issues/196 ; some are still having issues, but as far as I understand, the topic is closed on the `libsodium` side. It might be worth a test!

nkoreli:
Hi @alexandre-abrioux and @yomarion,

Another dependency in @hinkal/common is **snarkjs**, which relies on **wasm-unsafe-eval**. Snarkjs is widely used in zero-knowledge proof applications like Railgun and Penumbra, primarily because it enables in-browser proving.

The WASM files used by snarkjs are generated by CIRCOM (https://docs.circom.io/), a toolkit for zk-circuits. Using WASM for client-side proving offers significant performance improvements over JavaScript.

Since we are in the process of integrating with Request Finance, do you think it might be possible to adjust the content policy to permit wasm-unsafe-eval?

alexandre-abrioux:
@nkoreli Thank you for all the details 🙏 At the moment, Request Finance does not require Wasm scripts, so we do not allow it. We can of course reassess during implementation (but let's discuss this internally as this is a security topic).

PR #1205, opened 2023-10-20
feat!: output ES modules instead of CommonJS

MantisClone:
Fixes #1148 

## Description of the changes

* Make all packages output ES modules instead of CommonJS modules with exceptions:
  * `smart-contracts` must remain a CommonJS module because Hardhat uses `require()` statements. `smart-contracts` now uses `import()` to dynamically import ES modules. Reference: [Interoperability with CommonJS](https://nodejs.org/docs/latest-v18.x/api/esm.html#interoperability-with-commonjs)
* Add `.js` and `/index.js` file extensions for all path specifiers. Reference: [Mandatory File Extensions](https://nodejs.org/docs/latest-v18.x/api/esm.html#mandatory-file-extensions)
* Disable UMD bundling using Webpack in `request-client.js`, `web3-signature`, `epk-signature`, and `epk-decryption`.
* Update dependencies, for type information:
  * dotenv
  * ipfs-unixfs
  * axios
* Migrate from `typed-emitter` to `tiny-typed-emitter`
  * Fix error: `src/ethereum-storage-ethers.ts(30,35): error TS2709: Cannot use namespace 'TypedEmitter' as a type.`
* Use default imports for `Keyv`, `Bluebird`, `AJV`, and `FormData`

Issue #1148, opened 2023-09-12
As a Builder, I want package tree-shaking, so I can minimize my build size, compile time, test time, and page load time.


MantisClone:

MantisClone:
@skiv71 discovered that when we configure `"module": "es2015"` we can still use it in a node js application with `require()`. So maybe we don't need to use `"module": "node16"`.

MantisClone:
We've discussed this task a lot this past week, despite it not being part of the sprint. I've decided to move it into this sprint so that we can implement it and move on.

MantisClone:
While working on this, I disabled Webpack. I think that we'll remove Webpack as part of this task so I've added #1203 as a subtask in the description.

PR #1605, opened 2025-04-14
chore(typescript): change moduleResolution

alexandre-abrioux:
In this PR, I'm proposing to change TypeScript's `module` and `moduleResolution` to the latest recommended versions (see the section below for TypeScript recommendations). This would allow us to use ESM features in CommonJS, and forbid us to use non-ESM compatible imports, which should ease the transition to ESM later on. We would also gain:
- Support for `import ... from "esm"`: allows to import ESM modules synchronously in CommonJS, without the need to use `await import()`
- Support for the new package.json's `exports` property, which allows importing part of a module with `import {} from module/part`. e.g. `dotenv/config`. This change is required for https://github.com/RequestNetwork/requestNetwork/pull/1236
- Support for code splitting in consumer code: dynamic imports `await import('module')` will be kept as-is during transpilation to CommonJS, instead of being converted to `yield Promise.resolve().then(() => require('module'));`, which was preventing code splitting with consuming bundlers.

#### "module": "commonjs" is deprecated
https://www.typescriptlang.org/docs/handbook/modules/reference.html#commonjs
> commonjs
>
> You probably shouldn’t use this. Use node16, node18, or nodenext to emit CommonJS modules

#### "moduleResolution": "node" is deprecated
https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node
> node10 (formerly known as node)
>
> --moduleResolution node ... should no longer be used.

## Issue to address before merge

- [x] ~~`@hinkal/common` does not export types, the `types` field is missing in the `package.json`:
https://cdn.jsdelivr.net/npm/@hinkal/common@0.2.9/package.json~~
  Fixed in `0.2.12`, https://cdn.jsdelivr.net/npm/@hinkal/common@0.2.12/package.json

## Summary by CodeRabbit

- **New Features**
  - Improved validation format support by integrating ajv-formats for enhanced data validation.
  - Jest test execution now adapts to available system resources for better performance during testing.

- **Bug Fixes**
  - Updated test assertions to match new validation error message formats.

- **Chores**
  - Upgraded multiple dependencies across packages, including TypeScript, tslib, dotenv, ajv, and ts-jest, for improved compatibility and security.
  - Updated TypeScript configuration for modern module resolution and interoperability.
  - Refined type imports and type aliasing to reduce reliance on internal module paths.
  - Clarified type-only exports across various modules to improve type safety and code clarity.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ jobs:
   test-unit:
     docker:
       - *node_image
-    resource_class: xlarge
+    resource_class: large
     working_directory: *working_directory
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,6 +119,9 @@ jobs:
             --ignore @requestnetwork/payment-processor \
             --ignore @requestnetwork/integration-test \
             --concurrency=2
+          environment:
+            # Lerna starts 2 Jest commands at the same time (--concurrency=2), so we use 50% core for each
+            JEST_MAX_WORKERS: '50%'
       - store_test_results:
           path: packages/advance-logic/reports/
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,8 @@ jobs:
             --ignore @requestnetwork/integration-test \
             --concurrency=2
           environment:
-            # Lerna starts 2 Jest commands at the same time (--concurrency=2), so we use 50% core for each
+            # Lerna starts 2 Jest commands at the same time (see above --concurrency=2),
+            # so we use 50% of our CPU cores on each
             JEST_MAX_WORKERS: '50%'
       - store_test_results:
           path: packages/advance-logic/reports/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,7 @@ jobs:
   test-unit:
     docker:
       - *node_image
-    resource_class: large
+    resource_class: xlarge
     working_directory: *working_directory
     steps:
       - attach_workspace:
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "npm-package-json-lint": "5.1.0",
     "prettier": "2.8.8",
     "prettier-plugin-solidity": "1.0.0-beta.19",
-    "typescript": "5.1.3"
+    "typescript": "5.8.3"
   },
   "resolutions": {
     "underscore": "^1.12.1",
@@ -69,5 +69,6 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -69,5 +69,6 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -71,5 +71,6 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -71,6 +71,5 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -69,6 +69,5 @@
     "semver": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/197",
     "json-schema": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/51",
     "json5": "https://github.com/RequestNetwork/requestNetwork/security/dependabot/165"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/packages/data-access/src/index.ts
+++ b/packages/data-access/src/index.ts
@@ -3,6 +3,6 @@ export { CombinedDataAccess } from './combined-data-access';
 export { DataAccessWrite } from './data-write';
 export { DataAccessRead } from './data-read';
 export { PendingStore } from './pending-store';
-export { DataAccessBaseOptions } from './types';
+export type { DataAccessBaseOptions } from './types';
 export { MockDataAccess } from './mock-data-access';
 export { NoPersistDataWrite } from './no-persist-data-write';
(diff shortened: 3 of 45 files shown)


packages/payment-detection/package.json
```diff
@@ -57,7 +57,7 @@
     "@graphql-codegen/cli": "4.0.1",
     "@graphql-codegen/typescript": "4.0.1",
     "@graphql-codegen/typescript-document-nodes": "4.0.1",
-    "@graphql-codegen/typescript-graphql-request": "6.0.1",
+    "@graphql-codegen/typescript-graphql-request": "6.2.0",
```
alexandre-abrioux, line 60: Fixes the generated GraphQL code that previously contained `imports` not compatible with latest module resolution.

packages/payment-detection/src/thegraph/client.ts
```diff
@@ -2,9 +2,8 @@
 import { CurrencyTypes } from '@requestnetwork/types';
 import { NearChains } from '@requestnetwork/currency';
 import { GraphQLClient } from 'graphql-request';
-import { Block_Height, Maybe, getSdk } from './generated/graphql';
+import { Block_Height, getSdk, Maybe } from './generated/graphql';
 import { getSdk as getNearSdk } from './generated/graphql-near';
-import { RequestConfig } from 'graphql-request/src/types';
```
alexandre-abrioux, line 7: We can't do such types of imports anymore, as `graphql-request/src/types` is not an exported module.

packages/request-node/package.json
```diff
@@ -56,7 +56,7 @@
     "@requestnetwork/utils": "0.54.0",
     "chalk": "4.1.0",
     "cors": "2.8.5",
-    "dotenv": "8.2.0",
+    "dotenv": "16.5.0",
```
alexandre-abrioux, line 59: Old versions of `dotenv` did not export their types properly

packages/smart-contracts/package.json
```diff
@@ -78,7 +78,7 @@
     "@types/mocha": "8.2.3",
     "@types/node": "18.11.9",
     "chai": "4.3.4",
-    "dotenv": "10.0.0",
+    "dotenv": "16.5.0",
```
alexandre-abrioux, line 81: Even if not broken, let's use the same version of `dotenv` as in the other packages

packages/request-node/package.json
alexandre-abrioux, line 59: There was an issue with how `dotenv` exported their types in older versions

alexandre-abrioux:
Thanks a lot @nkoreli 🙏

packages/data-format/src/index.ts
```diff
@@ -1,4 +1,5 @@
-import * as AJV from 'ajv';
+import { Ajv } from 'ajv';
```
alexandre-abrioux, line 1: There was an issue with importing Ajv with the old version. I fixed it by updating the package. There was some breaking changes.

Later on, I turned `esModuleInterop=true` in the `tsconfig.json`, which may also have resolved this issue without updating the `Ajv` package. I haven't rolled back the Ajv update though...

tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "target": "es2015",
-    "module": "commonjs",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "esModuleInterop": true,
```
alexandre-abrioux, line 6: `esModuleInterop=true` is now the default: https://www.typescriptlang.org/tsconfig/#esModuleInterop

We still set it in our config file so that other tools (eslint, ts-jest) are aligned with it.

.circleci/config.yml
```diff
@@ -119,6 +119,10 @@ jobs:
             --ignore @requestnetwork/payment-processor \
             --ignore @requestnetwork/integration-test \
             --concurrency=2
+          environment:
+            # Lerna starts 2 Jest commands at the same time (see above --concurrency=2),
+            # so we use 50% of our CPU cores on each
+            JEST_MAX_WORKERS: '50%'
```
alexandre-abrioux, line 125: stabilize this CI job which was constantly using 100% RAM and failing with timeouts

tsconfig.json
```diff
@@ -1,7 +1,10 @@
...
+    "isolatedModules": true,
```
alexandre-abrioux, line 7: `ts-jest` requires `isolatedModules=true` with `module=nodenext`, see:
https://kulshekhar.github.io/ts-jest/docs/guides/esm-support/#using-hybrid-module-values

packages/types/src/data-access-types.ts
```diff
@@ -165,4 +166,4 @@ export interface IPendingStore {
   remove(channelId: string): void;
 }
 
-export { AuthSig };
+export type { AuthSig };
```
alexandre-abrioux, line 169: needed when `isolatedModules=true`

packages/data-access/src/index.ts
```diff
@@ -3,6 +3,6 @@ export { CombinedDataAccess } from './combined-data-access';
 export { DataAccessWrite } from './data-write';
 export { DataAccessRead } from './data-read';
 export { PendingStore } from './pending-store';
-export { DataAccessBaseOptions } from './types';
+export type { DataAccessBaseOptions } from './types';
```
alexandre-abrioux, line 6: needed when `isolatedModules=true`

MantisClone (commented):
Directionally, I like this change, but I"m confused on some details:

Questions:

1. When you say "code splitting" is this the same as tree-shaking?
2. Does this cause our build to output ES Modules or CommonJS modules?
3. Why doesn't this change require you to append `.js` onto every import (like `node16` did in https://github.com/RequestNetwork/requestNetwork/pull/1205)?
4. What is `import ... from "esm"` and why do we want it?
5. How did you handle the `smart-contracts`? As far as I know,
    > `smart-contracts` must remain a CommonJS module because Hardhat uses `require()` statements. `smart-contracts` now uses `import()` to dynamically import ES modules. Reference: [Interoperability with CommonJS](https://nodejs.org/docs/latest-v18.x/api/esm.html#interoperability-with-commonjs)
7. Are the RN packages still usable in CommonJS projects?
8. Why don't you set `type: 'module'` for each module?

Context:
* I previously attempted a similar refactor in PR https://github.com/RequestNetwork/requestNetwork/pull/1205 with the primary goal of enabling tree-shaking (Issue https://github.com/RequestNetwork/requestNetwork/issues/1148) but never completed the task because it was discovered that tree-shaking would still be blocked by https://github.com/RequestNetwork/requestNetwork/issues/1202 (which we've subsequently broken down to separate issues), as described by [this comment](https://github.com/RequestNetwork/requestNetwork/issues/1148#issuecomment-1779569982)
* When I stopped the refactor, I left this note to myself: https://github.com/RequestNetwork/requestNetwork/pull/1205#issuecomment-1775215569
    > TODO: Consider making currency and utils packages output ES modules AND CommonJS modules.
    > This would decrease the code churn in the smart-contracts package. smart-contracts could consume the CommonJS variants of the currency and utils packages, but everything else could consume the ES Module variants.
    > TODO: Consider making all packages output ES modules AND CommonJS modules.

package.json
```diff
@@ -52,7 +52,7 @@
     "npm-package-json-lint": "5.1.0",
     "prettier": "2.8.8",
     "prettier-plugin-solidity": "1.0.0-beta.19",
-    "typescript": "5.1.3"
+    "typescript": "5.8.3"
```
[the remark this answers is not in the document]
MantisClone: Per the link:

> `require()` only supports loading ECMAScript modules that meet the following requirements:
> 
> * The module is fully synchronous (contains no top-level `await`); and
> * One of these conditions are met:
>     * The file has a `.mjs` extension.
>     * The file has a `.js` extension, and the closest `package.json` contains `"type": "module"`
>     * The file has a `.js` extension, the closest `package.json` does not contain `"type": "commonjs"`, and the module contains ES module syntax.
> 
> If the ES Module being loaded meets the requirements, `require()` can load it and return the module namespace object. In this case it is similar to dynamic `import()` but is run synchronously and returns the name space object directly.

I'm guessing we satisfy the final bullet point because we use `.js` extension and not `"type": "commonjs"`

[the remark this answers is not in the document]
MantisClone: I'm guessing we satisfy the final bullet point because we use `.js` extension and not `"type": "commonjs"`

Per the link:

> `require()` only supports loading ECMAScript modules that meet the following requirements:
> 
> * The module is fully synchronous (contains no top-level `await`); and
> * One of these conditions are met:
>     * The file has a `.mjs` extension.
>     * The file has a `.js` extension, and the closest `package.json` contains `"type": "module"`
>     * The file has a `.js` extension, the closest `package.json` does not contain `"type": "commonjs"`, and the module contains ES module syntax.
> 
> If the ES Module being loaded meets the requirements, `require()` can load it and return the module namespace object. In this case it is similar to dynamic `import()` but is run synchronously and returns the name space object directly.

alexandre-abrioux:
Thanks for the review @MantisClone 🙏

## Context

Sorry about the confusion, I should have added more context.

ℹ️ This PR is **_not_** about migrating to ESM, our emitted code remains CommonJS.

We are migrating to a more "modern" version of CommonJS supported only by Node.js >= 12. This makes us adopt stricter `import` statements that will facilitate the migration to ESM later-on. For instance, we can't import random files in a library anymore (this won't work: `import x from "mod/dist/not-exported.js"`). We can only import objects explicitly exported by the library. This is how it behaves in ESM, but we stay in CommonJS.

## How the code is transpiled

| TS                               | Transpiled                      | Comment                                             |
|----------------------------------|---------------------------------|-----------------------------------------------------|
| `import x from "mod";`           | `const x = require("mod")`      | `mod` can be CommonJS or ESM (this is new)                  |
| `const x = await import("mod");` | `const x = yield import("mod")` | This is new, before it would transpile to:<br>`Promise.resolve().then(() => require('mod'));`<br>Thanks to this, build tools can split<br>code more efficiently for frontend apps |

## Answers
1. ✅
2. CommonJS
3. This will be required to migrate to ESM, yes
4. More and more packages export only ESM code (e.g. `@ecies/ciphers`, needed for https://github.com/RequestNetwork/requestNetwork/pull/1236). Today, our only option to load an ESM module is to load it asynchronously with `const x = await("esm")`. With this change, we can import an ESM module the same way as any other module, as long as the module in question does not contain any top-level `await`. We do not use that feature in this PR, but it will be made possible.
5. Nothing specific as we stayed in CommonJS
6. Yes
7. Same as 3, the `type: 'module'` property is how we indicate that our code is built for ESM. We should do this to migrate to ESM later-on.
8. I don't think it will change much regarding tree shaking. Migrating to ESM probably will 👍

I hope this answered your questions 🙂

MantisClone (approved):
Thanks for explaining! Approved 👍

PR #1721, opened 2026-04-21
feat: boba network

LeoSlrRf:
## Description

Support for the **boba** network.

The following contracts were deployed: 
 - RequestDeployer
 - ERC20Proxy
 - ERC20FeeProxy
 - EthProxy
 - EthFeeProxy
 - BatchConversionPayments
 
All verified on https://bobascan.com/

#### Comments to reviewers:

Contracts' addresses are different from the previous deployment due to a different compiler configuration.
Also mocked the `getLogs` calls in integration tests. They were hitting pubic RPC causing flakiness.

### DIFF
--- a/packages/payment-detection/src/thegraph/queries/superfluid/graphql.config.yml
+++ b/packages/payment-detection/src/thegraph/queries/superfluid/graphql.config.yml
@@ -1,1 +1,1 @@
-schema: https://subgraph-endpoints.superfluid.dev/eth-sepolia/protocol-v1
+schema: https://subgraph-endpoints.superfluid.dev/eth-mainnet/protocol-v1
--- a/packages/request-client.js/test/index.test.ts
+++ b/packages/request-client.js/test/index.test.ts
@@ -164,6 +164,16 @@ const waitForConfirmation = async (
   });
 };
 
+/**
+ * Avoid live `eth_getLogs` RPC in CI. With subgraph disabled, ETH/ERC20 proxy
+ * fallbacks use `getDefaultProvider().getLogs` over large block ranges, which
+ * is slow and can time out. Tests that use mocked Etherscan (or only assert
+ * currency) do not need those logs. Do not use for local `private` chain
+ * transfer integration tests.
+ */
+const mockEvmGetLogsEmpty = (): jest.SpyInstance =>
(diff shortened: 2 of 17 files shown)


packages/payment-detection/src/thegraph/queries/superfluid/graphql.config.yml
```diff
@@ -1 +1 @@
-schema: https://subgraph-endpoints.superfluid.dev/eth-sepolia/protocol-v1
```
LeoSlrRf, line 1: The subgraph at this location couldn't be found. Switched to mainnet instead.

alexandre-abrioux-rf (approved):

alexandre-abrioux-rf (approved):

packages/request-client.js/test/index.test.ts
```diff
@@ -164,6 +164,16 @@ const waitForConfirmation = async (
   });
 };
 
+/**
+ * Avoid live `eth_getLogs` RPC in CI. With subgraph disabled, ETH/ERC20 proxy
```
yomarion-rf, line 168: Why do we avoid it only for some tests? Shouldn't we record and replay the returned values instead?

yomarion-rf (approved):

packages/request-client.js/test/index.test.ts
LeoSlrRf: The `EthInputDataPaymentDetector` calls two sources of payment.

The first one is `getHistory` which was mocked and upon which the flaky tests were based.
The second one is based on the Proxy contract in case a deployment exist. For this one, we rely on Subgraph events, but fallback to `getLogs` in case the Subgraph client is `undefined` - which is the case here. I mocked the results to be an empty array, which is what we were getting when calling the public RPC - and why the tests are still passing with the mock.

yomarion-rf: But shouldn't this behaviour mock EVERY access to `getLogs` in this request client test? This is my question, why just a few?

yomarion-rf-admin (approved):

dave-rf (approved):

packages/request-client.js/test/index.test.ts
LeoSlrRf: These are the only tests where the `getLogs` logic is triggered AND the chain isn't `private`

PR #1741, opened 2026-06-19
fix(smart-contracts): address stranded balance issues for ERC20

yomarion:
## Description of the changes

Fixes https://github.com/RequestNetwork/private-issues/issues/23

The `BatchConversionContract` Refunds ERC20 tokens not based on his balance but based on what was topped up by the payer at the beginning of the transaction.

New method for the contract owner: `rescueERC20Funds`.

- [x] New method for the contract owner to drain a balance of ERC20.

### DIFF
--- a/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
+++ b/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
@@ -543,6 +543,23 @@ contract BatchNoConversionPayments is Ownable {
   }
 
   /*
+   * Admin functions for balance handling
+   */
+
+  /**
+   * @notice 
+   * @param erc20TokenAddress Address of an ERC20 to drain
+   * @param recipientAddress Address of the receiver
+   */
+  function rescueERC20Funds(address erc20TokenAddress, address recipientAddress)
+    external
+    onlyOwner
+  {
+    IERC20 token = IERC20(erc20TokenAddress);
+    token.transfer(recipientAddress, token.balanceOf(address(this)));
+  }
+
+  /*
    * Admin functions to edit the proxies address and fees
    */
 
--- a/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
+++ b/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
@@ -547,16 +547,20 @@ contract BatchNoConversionPayments is Ownable {
    */
 
   /**
-   * @notice 
-   * @param erc20TokenAddress Address of an ERC20 to drain
+   * @notice Transfers the entire ERC20 balance of this contract to the specified recipient. Only callable by the owner.
+   * @param erc20TokenAddress Address of an ERC20 token sent to the contract (eg. user mistake)
    * @param recipientAddress Address of the receiver
    */
   function rescueERC20Funds(address erc20TokenAddress, address recipientAddress)
     external
     onlyOwner
   {
+    require(erc20TokenAddress != address(0), 'Invalid token address');
     IERC20 token = IERC20(erc20TokenAddress);
-    token.transfer(recipientAddress, token.balanceOf(address(this)));
+    uint256 balance = token.balanceOf(address(this));
+    require(balance > 0, 'No funds to rescue');
+    bool success = SafeERC20.safeTransfer(token, recipientAddress, balance);
+    require(success, 'ERC20 rescue failed');
   }
 
   /*
--- a/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
+++ b/packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
@@ -44,11 +44,14 @@ contract BatchNoConversionPayments is Ownable {
   address public NativeAddress;
   address[][] public pathsNativeToUSD;
 
-  /** Contains the address of a token, the sum of the amount and fees paid with it, and the batch fee amount */
   struct Token {
     address tokenAddress;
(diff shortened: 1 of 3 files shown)


packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
```diff
@@ -536,6 +542,23 @@ contract BatchNoConversionPayments is Ownable {
     return result;
   }
 
+  /*
+   * Admin functions for balance handling
+   */
+
+  /**
+   * @notice
```
rodrigopavezi: the `@notice` is empty, worth filling in before merging — something like `Transfers the entire ERC20 balance of this contract to the specified recipient. Only callable by the owner.`

packages/smart-contracts/src/contracts/BatchNoConversionPayments.sol
```diff
@@ -536,6 +542,23 @@ contract BatchNoConversionPayments is Ownable {
...
+   * @notice 
+   * @param erc20TokenAddress Address of an ERC20 to drain
+   * @param recipientAddress Address of the receiver
+   */
+  function rescueERC20Funds(address erc20TokenAddress, address recipientAddress)
+    external
+    onlyOwner
+  {
+    IERC20 token = IERC20(erc20TokenAddress);
+    token.transfer(recipientAddress, token.balanceOf(address(this)));
```
rodrigopavezi: should use `safeTransfer` here instead of raw `transfer` — some ERC20s don't return a boolean on `transfer` and it'll silently fail. the rest of the contract already uses `safeTransfer` consistently.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -448,6 +450,110 @@ describe('contract: BatchConversionPayments', async () => {
...
+          );
+
+          // Sanity checks
+          expect(toBalance.sub(initialToTOKBalance)).to.equals(
+            expectedToTOKBalanceDiff,
+            '(sanity) Receiver should receive the request amount',
+          );
+          expect(feeBalance.sub(initialFeeTOKBalance).gt(0),
+            '(sanity) Fee address should receive fees',
+          );
```
rodrigopavezi: this `expect()` is missing a `.to.be.true` at the end — without it the call is a no-op and this sanity check never actually asserts anything

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -448,6 +450,110 @@ describe('contract: BatchConversionPayments', async () => {
...
+          expect(fromBalance.sub(initialFromTOKBalance).gt(0),
+            '(sanity) Spender should spend',
+          );
```
rodrigopavezi: same here — missing `.to.be.true`. also `fromBalance.sub(initialFromTOKBalance)` will throw a BigNumber underflow since the spender balance decreases after payment, probably needs `initialFromTOKBalance.sub(fromBalance).gt(0).to.be.true`

rodrigopavezi (commented):
hey, nice fix! the `tareBalance` approach is clean and the test coverage looks solid.

quick note — I checked `BatchNoConversionPayments` and it doesn't have excess refund logic, so the same stranded-balance bug doesn't apply there. the `tareBalance` field is currently only consumed by `BatchConversionPayments`, which is fine.

a few things I noticed inline:

yomarion:
> hey, nice fix! the `tareBalance` approach is clean and the test coverage looks solid.
> 
> quick note — I checked `BatchNoConversionPayments` and it doesn't have excess refund logic, so the same stranded-balance bug doesn't apply there. the `tareBalance` field is currently only consumed by `BatchConversionPayments`, which is fine.
> 
> a few things I noticed inline:

Thanks @rodrigopavezi for your in-depth review. I addressed all your feedback.

About the choice for implementing `rescueERC20Funds` in the `NoConversion` contract, my chain of thought was: exactly the same implementation costs and could (very hypothetically) prevent a mistake in the future.

packages/smart-contracts/test/contracts/BatchConversionPayments.test.ts
```diff
@@ -448,6 +450,117 @@ describe('contract: BatchConversionPayments', async () => {
...
+          expect(
+            initialFromTOKBalance.sub(fromBalance).gt(0),
+            `(sanity) Spender should spend >0, spent: ${initialFromTOKBalance
+              .sub(fromBalance)
+              .toString()}`,
+          ).to.be.true;
+          // The real spend must be strictly below maxToSpend, otherwise there is no
+          // excess slack and the stranded balance could not be touched at all.
+          expect(
+            fromBalance.sub(initialFromTOKBalance).lt(BigNumber.from(maxToSpend)),
```
MantisClone: This check can pass without proving the thing it is meant to prove. The test needs to confirm the payer spent less than `maxToSpend`, so there was excess TOK to refund.

Right now the subtraction is backwards. Can we flip it?

```suggestion
            initialFromTOKBalance.sub(fromBalance).lt(BigNumber.from(maxToSpend)),
```

MantisClone (approved):
Optimistically Approved :+1: pending comment resolution :construction:

Nice work. The stranded-balance fix makes sense to me, and the rescue path looks scoped correctly.

I left one small test comment: the sanity check should flip the payer-balance subtraction so it actually proves the spend was below `maxToSpend`. Once that is fixed, this looks good to me.

rodrigopavezi (approved):

PR #1743, opened 2026-06-22
feat(smart-contracts): BatchConversionContract 0.2.0 mainnet artifact

yomarion:
## Description of the changes

Main addition:
1. BatchConversionContract artifact for 0.2.0 (merged in #1741 and #1739 )

The code changes only affect local commands:

- ` yarn hardhat deploy-contracts-through-deployer`:

1. Improved error handling for lack of funds, and absence of `ADMIN_PRIVATE_KEY`
2. Don't try to configure contracts that could not be deployed (still do for previously deployed contracts)

- `yarn hardhat verify-contract-from-deployer`
1. Improved error handling with useful verbosity

### DIFF
--- a/packages/smart-contracts/README.md
+++ b/packages/smart-contracts/README.md
@@ -67,7 +67,7 @@ The package stores the following smart contracts:
 - `RequestOpenHashSubmitter` entry point to add hashes in `RequestHashStorage`. It gives the rules to get the right to submit hashes and collect the fees. This contract must be whitelisted in `RequestHashStorage`. The only condition for adding hash is to pay the fees.
 - `StorageFeeCollector` parent contract (not deployed) of `RequestOpenHashSubmitter`, computes the fees and send them to the burner.
 
-**Smart contracts for advanced-logic package**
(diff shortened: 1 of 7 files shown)


rodrigopavezi (approved):
looks good!

MantisClone (approved):
Approved 👍 pending the small comment below 👇 🚧 

I traced the SDK path: v0.2.0 is registered, and payment + approval both use the same `version` option.

Main note: this is opt-in. Default is still v0.1.0, so RF needs to pass `version: '0.2.0'` to use the new mainnet contract.

Small ops note: manual setup reruns also need the v0.2.0 address passed explicitly, otherwise the fallback points at v0.1.0.

I agree with Greptile’s warning here:

> `compute-one-address.ts` — the catch block emits a misleading "Unknown error" warning for intentional validation failures, which could slow down debugging of deployment issues.

I checked the code too: the catch also wraps known config/validation errors like missing salt, missing deployer address, or invalid initcode as “Unknown error.”

PR #1763, opened 2026-08-24
feat: ERC20RecurringPaymentProxy batch schedule v0.2.0

LeoSlrRf:
### What changed

**Batch permit model replaces the single-recipient permit**

`SchedulePermit` is replaced by `SchedulePermitBatch`, which carries two typed arrays of `Leg` structs (`initialLegs` and `recurringLegs`). Each `Leg` holds a recipient, amount, and an 8-byte payment reference. Index 1 settles `initialLegs` when present; all subsequent indices settle `recurringLegs`. Up to 8 legs per array (`MAX_LEGS = 8`) are allowed. Payment references must be unique within each array.

**Explicit due-time array instead of `firstPayment + periodSeconds`**

The permit now carries a `uint32[] dueTimes` array (one entry per cycle) that must be strictly increasing. This replaces the previous `firstPayment`/`periodSeconds` arithmetic and allows irregular schedules.

**`scheduleId` anchors replay state independently of nonce/deadline**

A `bytes32 scheduleId` field is committed into both the schedule key and the EIP-712 digest. `nonce` and `deadline` are in the digest but excluded from the key, so re-signing with a new nonce or deadline does not change the schedule key and bitmap/cancellation state persist across re-signs. A zero `scheduleId` is rejected. Changing any other signed term produces a new key with a virgin bitmap and cancelled flag.

**`Ownable` removed; all admin actions gated by `DEFAULT_ADMIN_ROLE`**

`pause`, `unpause`, `setFeeProxy`, `grantRelayer`, and `revokeRelayer` now require `DEFAULT_ADMIN_ROLE` instead of `Ownable`. `grantRelayer`/`revokeRelayer` replace the single-slot `setRelayer`. Multiple relayers may hold `RELAYER_ROLE` simultaneously and compete for the `relayerFee`.

**Subscriber self-trigger via `admitCycles` / `revokeCycles`**

Relayers can grant (`admitCycles`) or clear (`revokeCycles`) a bitmask of cycle indices on a schedule key. A subscriber may call `triggerRecurringPaymentBatch` directly only for admitted cycles. Relayers bypass the admission check entirely.

**Hardened token-pull path**

`_pullExact` verifies the contract's balance increased by exactly the requested amount, rejecting fee-on-transfer tokens and silent-fail `transferFrom` implementations. `_approveFeeProxy` and `_payRelayer` also check return values. A post-settlement balance assertion (`UnexpectedBalance`) ensures nothing leaks.

**`cancelScheduleBatch` lets subscribers halt a schedule**

Subscribers can mark a schedule key as cancelled; subsequent trigger attempts revert with `Cancelled`.

**`rescueTokens` for stuck balances**

Admins can recover any ERC-20 token accidentally left in the contract.

**EIP-1271 smart-account signatures**

`ECDSA.recover` is replaced by `SignatureChecker.isValidSignatureNow`, enabling contract-wallet subscribers.

**`FeeProxyUpdated` event on `setFeeProxy`**

The old and new proxy addresses are emitted when the fee proxy is updated.

### New test contracts

- `ERC20SilentFail` — returns `false` on a failed `transferFrom` instead of reverting.
- `ERC20FeeOnTransfer` — under-delivers by 1 token on every `transferFrom`.
- `ERC20FailTransfer` — `transfer()` always returns `false`.
- `ERC20BlockRecipient` — reverts transfers to a configurable blocked address.
- `MockERC1271` — minimal ERC-1271 wallet for smart-account signature tests.

### ABI artifact

`src/lib/artifacts/ERC20RecurringPaymentProxy/0.2.0.json` is added with the full v0.2.0 ABI.


LeoSlrRf:
* **#1767** <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1766** <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1766?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1765** <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1765?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1764** <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1764?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1763** <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1763?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1763?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rodrigopavezi (commented):
great job!!   
  
Let's only deploy to mainnet chains we support after we test everything same chain direct recurring payment on testnet. we never know how it will behave with the rest of the stack and the UX.  
  
Cheers

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
```diff
@@ -70,89 +107,376 @@ contract ERC20RecurringPaymentProxy is EIP712, AccessControl, Pausable, Reentran
...
+    return
+      keccak256(
+        abi.encode(
+          p.subscriber,
+          p.scheduleId,
+          p.token,
+          p.relayerFee,
+          p.totalPayments,
+          p.strictOrder,
+          _hashUint32Array(p.dueTimes),
```
rodrigopavezi: **Gas: the same three hashes are computed twice on every trigger.**

`_hashUint32Array(p.dueTimes)`, `_hashLegs(p.initialLegs)` and `_hashLegs(p.recurringLegs)` are computed here for the schedule key and again at L145-147 for the digest. Both run on every `triggerRecurringPaymentBatch` call.

Each helper allocates a `bytes32[] memory` and hashes it, and `dueTimes` can hold up to 255 entries with 8 legs per array — so this is real, repeated cost on the hot path, paid every cycle for the life of every schedule.

Compute the three once in `triggerRecurringPaymentBatch` and pass them into both `_scheduleKeyFromBatch` and `_hashScheduleBatch`.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
```diff
@@ -70,89 +107,376 @@ contract ERC20RecurringPaymentProxy is EIP712, AccessControl, Pausable, Reentran
...
+    }
+  }
 
-    uint256 total = p.amount + p.feeAmount + p.relayerFee;
+  function _assertOrder(
+    ScheduleState storage state,
+    uint8 index,
+    bool strictOrder
+  ) private view {
+    if (strictOrder && index != state.lastIndex + 1) {
```
rodrigopavezi: **Panics instead of reverting at the last index.**

`state.lastIndex` is `uint8`, so `state.lastIndex + 1` is `uint8` arithmetic. With `strictOrder` and `totalPayments = 255`, once index 255 is paid `lastIndex` is 255 and this expression overflows — a `Panic(0x11)` rather than `PaymentOutOfOrder`.

The bounds check above cannot prevent it: `index` is `uint8`, so `index > p.totalPayments` is never true when `totalPayments` is 255, and execution reaches this line.

Suggest widening before the add:

```solidity
if (strictOrder && uint256(index) != uint256(state.lastIndex) + 1) {
```

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
```diff
@@ -70,89 +107,376 @@ contract ERC20RecurringPaymentProxy is EIP712, AccessControl, Pausable, Reentran
...
+  function _payRelayer(IERC20 token, uint256 amount) private {
+    if (amount == 0) {
+      return;
+    }
+    if (!token.safeTransfer(msg.sender, amount)) {
+      revert ERC20RecurringPaymentProxy__TransferFailed();
+    }
+  }
+
+  function _assertNonZeroRecipient(address account, uint256 amount) private pure {
```
rodrigopavezi: **Dead code — `_assertNonZeroRecipient` is defined but never called.**

No call site anywhere in the contract. In a payment contract this reads as a guard someone intended and then dropped, which is exactly the kind of thing an auditor flags and a reader mis-trusts.

Worth checking whether it was meant to guard something that is now unguarded — otherwise delete it.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
```diff
@@ -70,89 +107,376 @@ contract ERC20RecurringPaymentProxy is EIP712, AccessControl, Pausable, Reentran
...
+    }
+  }
+
+  function _assertLegs(Leg[] calldata legs) private pure {
+    if (legs.length == 0) revert ERC20RecurringPaymentProxy__EmptyLegs();
+    for (uint256 i = 0; i < legs.length; ++i) {
+      if (legs[i].recipient == address(0)) {
+        revert ERC20RecurringPaymentProxy__ZeroAddress();
+      }
+      if (legs[i].amount == 0) {
```
rodrigopavezi, line 361: **No uniqueness check on `paymentReference` across legs.**

`_assertLegs` validates recipient and amount per leg, but nothing stops two legs in the same array carrying the same `paymentReference`. `_settleLegs` makes one `transferFromWithReferenceAndFee` call per leg, so that emits two `TransferWithReferenceAndFee` events with an identical reference **in a single transaction**.

The payments subgraph builds entity ids without a log index, so those two events collide and one is lost. Today nothing emits two same-reference events in one tx, which is why that has been latent — this change makes it reachable.

Either reject duplicate references here, or make the subgraph log-index fix a prerequisite for deploying this.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
```diff
@@ -70,89 +107,376 @@ contract ERC20RecurringPaymentProxy is EIP712, AccessControl, Pausable, Reentran
...
+      abi.encode(
+        _BATCH_TYPEHASH,
+        p.subscriber,
+        p.token,
+        p.relayerFee,
+        p.totalPayments,
+        p.nonce,
+        p.deadline,
+        p.strictOrder,
+        p.scheduleId,
```
rodrigopavezi, line 167: Small correction for the PR description rather than the code: it states that `scheduleId` is "committed into the schedule key but **excluded from the EIP-712 digest**".

It is in the digest — encoded here, and declared in `_BATCH_TYPEHASH` at L53. The field that is genuinely excluded from the *key* (and present in the digest) is `nonce`/`deadline`, which is what gives the intended re-sign behaviour.

The code looks right; just the description that would mislead a reviewer or auditor.

rodrigopavezi (commented):
Strong PR. `_pullExact` is the right shape — a `balanceOf` delta rather than just a `require`, so it catches silent-fail *and* fee-on-transfer under-delivery, which a return-value check alone would miss. The post-settlement `UnexpectedBalance` assertion is a good addition, and the test **"cannot settle an unfunded subscriber from a residual proxy balance"** closes a vector that is reachable once the subscriber can self-trigger. Digest parity against ethers `_TypedDataEncoder` covers the nested-array hashing, which was the most likely thing to go quietly wrong.

Four concrete defects inline. Two other things worth raising here rather than on a line:

**1. The single-fee entry point is gone — is that intended?**

`triggerRecurringPayment` and the `SchedulePermit` struct are removed, so `0.2.0` is a new deployment alongside `0.1.0`'s seven live ones. The shaped decision was explicitly *additive*: keep the single entry point beside the batch one, because existing schedules are bound to the old address by their signed permits and **cannot be migrated**. Dropping it means `0.1.0` monthly schedules keep draining on the old address for up to 21 years while detection, the reconciler and the dashboard handle two addresses and two ABIs.

That may be the right call if few enough `0.1.0` schedules are live — but it is a decision worth making out loud rather than absorbing.

**2. `paymentReference` uniqueness makes a subgraph fix blocking.**

Nothing enforces distinct references across legs, so two legs sharing one emit two identical-reference `TransferWithReferenceAndFee` events in a single transaction. The subgraph builds entity ids without a log index, so those collide. Either enforce uniqueness in `_assertLegs`, or land the subgraph log-index fix before this deploys — it moves from latent to blocking with this change.

**Nit on the description:** it says `scheduleId` is "excluded from the EIP-712 digest", but it is in the digest (see inline). Worth correcting, since auditors read the description.

**Worth writing down somewhere:** the schedule key changes when any signed term changes, which the tests confirm is intentional. The consequence is that an amended schedule gets a virgin bitmap **and** a virgin `cancelled` flag — so cancelling schedule A does not cancel an amended variant B, and re-signing with a changed amount resets paid cycles. Fine if the engine always treats amend as cancel-plus-new-`scheduleId`, but that assumption should be explicit rather than inferred.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
LeoSlrRf: Fixed. `_hashPermitParts` now runs once per trigger; `_keyAndDigest` passes those hashes into both the schedule key and the EIP-712 digest.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
LeoSlrRf: Fixed. The order check now widens before adding: `uint256(index) != uint256(state.lastIndex) + 1`.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
LeoSlrRf: Fixed. Removed `_assertNonZeroRecipient`. Zero recipients are still rejected in `_assertLegs`.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
LeoSlrRf: Fixed. `_assertLegs` now reverts `DuplicatePaymentReference` when two legs in the same array share a `paymentReference`.

packages/smart-contracts/src/contracts/ERC20RecurringPaymentProxy.sol
LeoSlrRf: Fixed. PR description updated: `scheduleId` is in both the key and the digest. `nonce` / `deadline` are the fields excluded from the key.

LeoSlrRf:
Thanks for the review — replies are on the inline threads as well.

R1 (testnet before mainnet): Agreed. We will only deploy to supported mainnet chains after same-chain direct recurring payment is tested on testnet.

R2a (single-fee entry point removed): Not a code change. Discussed and agreed — `0.2.0` is a new deployment. Existing `0.1.0` schedules stay on the old address and remain callable via the old ABI.

R2b (unique paymentReference): Fixed — see the inline reply on `_assertLegs`.

R2c (scheduleId in the digest): Fixed — PR description corrected; see the inline reply.

R2d (amend = new key / virgin cancelled): Fixed. Documented on the `schedules` mapping: changing any signed term except nonce/deadline yields a new key with a virgin bitmap and cancelled flag.

MantisClone (approved):
Optimistically approved 👍 pending comment resolution 🏗

**Requested change:** Remove the duplicate `paymentReference` validation before merging.

**Why:** We do not currently offer partial-payment schedules, but the contract does not need to codify that product limitation. The pairwise check runs on every trigger and adds gas. It also does not establish a global uniqueness invariant: it compares only within `initialLegs` or `recurringLegs`, while initial and recurring legs, and later recurring cycles, can still reuse a reference.

**If unaddressed:** This adds a production constraint and recurring gas cost without establishing a complete uniqueness invariant. It narrows a future product option that we do not need to decide in the contract today.

Notes, no change requested:

- The native multi-leg batch design retains the core v0.1 execution protections. Every charge still requires a valid signed schedule, checks expiry, cycle bounds and due time, prevents replay, respects signed `strictOrder`, and is protected by pausing and reentrancy guards. EIP-1271 extends signature validation to contract wallets without weakening those checks.
- Execution remains bounded by the signed schedule. A relayer can trigger an eligible cycle. A subscriber can self-trigger only a cycle that a relayer has admitted; no other caller can trigger it. `admitCycles`, `revokeCycles`, and subscriber cancellation are new controls in v0.2, rather than inherited v0.1 behavior.
- A cycle settles atomically. Each leg emits its normal `TransferWithReferenceAndFee` event for payment detection, while the recurring proxy emits one `PaymentTriggered` event only after every leg and the relayer fee succeed.
- Moving `paymentReference` from `bytes` to `bytes8` is a good change. Request’s canonical payment-reference calculator already produces exactly eight bytes, so detection remains compatible while malformed or arbitrary-length references can no longer enter this flow. The reference is now also signed with the recipient and amount.
- `strictOrder` remains an explicit field in the signed schedule, as it was before. The relayer and admin cannot change that choice after the subscriber has signed.
- The deadline does not need to be after the final due time: it can deliberately cap a schedule early. That is reasonable, but callers should understand that any later due cycles will then be unexecutable.
- Continuing to use `SafeERC20` is correct. `_pullExact` complements it by checking that the proxy received the exact scheduled amount, so fee-on-transfer or short-transfer tokens fail cleanly rather than leaving a payment underfunded.
- I checked the compiler advisories issued since Solidity 0.8.9. None applies to this contract’s configuration, so I am comfortable retaining 0.8.9 for this deployment; it also matches the version pinned by the package’s Hardhat compiler.

rodrigopavezi (approved):

LeoSlrRf:
### Merge activity

* **Aug 31, 8:49 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1763).
* **Aug 31, 8:50 AM UTC**: @LeoSlrRf merged this pull request with [Graphite](https://app.graphite.com/github/pr/RequestNetwork/requestNetwork/1763).

LeoSlrRf:
@MantisClone Indeed we only enforce that the same reference is unique within the `initialLegs`, and within the `recurringLegs`, but we don't prevent the same reference to be in both;   
  
The idea behind is that within a transaction, we won't have several payments with the same payment reference which would required a subgraph update;

